PR #7, opened 2022-08-05
Rename dms* -> catalyst*

thomshutt:
See this thread for context on the rename: https://discord.com/channels/423160867534929930/991388659699949669/1004599546933948486

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   test:
-    name: Test the dms-api project
+    name: Test the catalyst-api project
     runs-on: ubuntu-20.04
     steps:
       - name: Check out code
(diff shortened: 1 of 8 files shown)


hjpotter92:
There might be some similar changes needed in gh-workflow?

thomshutt:
> There might be some similar changes needed in gh-workflow?

Good call, for some reason my recursive `grep` didn't go into the `.github`

.github/workflows/test.yaml
```diff
@@ -48,5 +48,5 @@ jobs:
         uses: codecov/codecov-action@v3
         with:
           files: ./coverage.out
-          name: dms-api
+          name: catalyst-api
```
hjpotter92, line 51: this would apply only when the project gets renamed :|

try setting a `CODEOWNERS` file and see if that grants you access to the project :smile:

hjpotter92 (approved):
lgtm

.github/workflows/test.yaml
thomshutt: Yeah, I was going to wait for the rename before merging

victorges (approved):

PR #8, opened 2022-08-05
Add Mist API client and scaffold for triggering upload VOD

leszko:
This is the first part of work related to https://github.com/livepeer/go-livepeer/issues/2537

It includes the following parts:
- Mist API client + unit tests
- Scaffold for the upload VOD function

This is how I tested it (and you can too!).

### Testing Mist client

1. Start catalyst node (Dockerized `livepeer/catalyst` is good enough)
2. Check the mist password hash
```
$ git clone https://github.com/DDVTECH/mistserver.git && cd mistserver
$ source api/mist_api.sh
$ mistCall test test '{"variable_list":true}'
$ cat /tmp/MistBashAuth
"authorize":{"username":"test","password":"f0c5a517021bae2634bad2b2c390bbe9"},
```

The string `f0c5a517021bae2634bad2b2c390bbe9` is your password hash.

3. Cherry pick commit with local test params https://github.com/livepeer/catalyst-api/commit/a849c04819b8cbd9c430dd157e71d97e2e7b4e93
4. Update your password hash in https://github.com/livepeer/catalyst-api/commit/a849c04819b8cbd9c430dd157e71d97e2e7b4e93#diff-3b598d3223d4e9fa913b3706f7dbf60e1674f052264ada27e092fd5bf9cc20baR340
5. Start DMS API app
6. Copy sample video into the Catalyst container
```
docker cp Big_Buck_Bunny_1080_10s_1MB.mp4 catalyst-container:/home/
```
7. Start `TestWorkflow()` https://github.com/livepeer/catalyst-api/commit/a849c04819b8cbd9c430dd157e71d97e2e7b4e93#diff-442464ef69390526d2eb9120b73417beffd1586d1c691456bb3ba2d8b420d054R151

8. You should see a notification of a trigger execution in the DMS API logs.
9. You should see that the output video was created in the catalyst container
```
$ docker exec -it catalyst-container sh
# cd /media/recording
# ls
result.ts
```


handlers/client_test.go
```diff
@@ -0,0 +1,40 @@
...
+		"command=%7B%22deletestream%22%3A%7B%22somestream%22%3A%7B%7D%7D%7D",
+		payloadFor(commandDeleteStream("somestream")),
+	)
+}
+
+// TODO: Remove after initial testing
+func TestWorkflow(t *testing.T) {
+	// first copy file into /home/Big_Buck_Bunny_1080_10s_1MB.mp4
+	mc := MistClient{apiUrl: "http://localhost:4242/api2"}
+	streamName, err := mc.AddStream("/home/Big_Buck_Bunny_1080_10s_1MB.mp4")
```
thomshutt, line 31: This is actually a good point - we could potentially speed things up by doing the S3 download in Golang to local disk and handing over to Mist from there. 

We'd still want to eventually have Mist pull natively to avoid running out of disk space at scale, but this would get things working E2E with a bit less effort.

leszko: yeah, I don't wee much difference if we do it from DMS API or Mist (as already implemented by @AlexKordic ). So, if Alex already implemented this part in Mist, then we can just pass the S3 URL to Mist.

thomshutt (approved):
Looks like a great first step! I hadn't realised the Mist config doesn't support partial updates

handlers/client_test.go
stronk-dev: Isn't the upload part covered by https://github.com/DDVTECH/mistserver/pull/66 ? Afaik with that merged or cherry picked on your branch you can use S3 as a stream source directly

leszko: Ok, sounds good!

stronk-dev: LMK if you run into any issues, it might require some config on the input side to recognize that s3:// URL's are supported

handlers/client.go
```diff
@@ -0,0 +1,336 @@
...
+	return cc.Config.Triggers, nil
+}
+
+func (mc *MistClient) sendCommand(command interface{}) (string, error) {
+	c, err := commandToString(command)
+	if err != nil {
+		return "", err
+	}
+	payload := payloadFor(c)
+	resp, err := http.Post(mc.apiUrl, "application/json", bytes.NewBuffer([]byte(payload)))
```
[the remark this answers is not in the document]
leszko: Yes, actually, it should be `application/x-www-form-urlencoded`. Updated.

handlers/client.go
```diff
@@ -0,0 +1,336 @@
...
+	triggers = deleteAllTriggersFor(triggers, streamName)
+	if len(replaceTrigger.Streams) != 0 {
+		triggers = append(triggers, replaceTrigger)
+	}
+
+	triggersMap[triggerName] = triggers
+	return MistConfig{Config{Triggers: triggersMap}}
+}
+
+func deleteAllTriggersFor(triggers []ConfigTrigger, streamName string) []ConfigTrigger {
```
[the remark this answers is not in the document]
leszko: It actually collects triggers that has no `streamName`. Anyway, renamed to `filterTriggersWithoutStream`.

handlers/client.go
```diff
@@ -0,0 +1,336 @@
...
+	if err != nil {
+		return err
+	}
+	c := commandDeleteTrigger(streamName, triggerName, triggers)
+	resp, err := mc.sendCommand(c)
+	return validateDeleteTrigger(streamName, triggerName, resp, err)
+}
+
+func (mc *MistClient) getCurrentTriggers() (Triggers, error) {
+	c := commandGetTriggers()
```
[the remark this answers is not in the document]
leszko: Added a comment `// send an empty config struct returns the current Mist configuration`. PTAL.

handlers/client.go
```diff
@@ -0,0 +1,336 @@
...
+		Sync:    false,
+	}
+	return commandUpdateTrigger(streamName, triggerName, currentTriggers, newTrigger)
+}
+
+func commandDeleteTrigger(streamName, triggerName string, currentTriggers Triggers) MistConfig {
+	return commandUpdateTrigger(streamName, triggerName, currentTriggers, ConfigTrigger{})
+}
+
+func commandUpdateTrigger(streamName, triggerName string, currentTriggers Triggers, replaceTrigger ConfigTrigger) MistConfig {
```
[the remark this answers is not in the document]
leszko: That's also an option, but the way I implemented:
- makes a clear separation `command*` is just to return the command, not for making any API requests
- makes is possible to unit test

red-0ne (approved):
LGTM

PR #14, opened 2022-08-09
Add HTTP callback client for communicating status to Studio

thomshutt:
Currently implementing Transcode Status success / error based on spec in https://www.notion.so/livepeer/Technical-Design-Studio-Catalyst-API-for-VOD-Recording-bba2b160efff4ad4963dae2d4e9ff1a2

Completion Outputs block (i.e telling Studio where we've put the files) will come in a separate PR

### DIFF
--- a/clients/callback_client.go
+++ b/clients/callback_client.go
@@ -1,0 +1,111 @@
+package clients
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"time"
+)
+
+const max_retries = 3
+const backoff_millis = 200
+
+type CallbackClient struct {
+	httpClient *http.Client
+}
+
+func NewCallbackClient() CallbackClient {
+	return CallbackClient{
+		httpClient: &http.Client{},
+	}
+}
+
+func (c CallbackClient) DoWithRetries(r *http.Request) error {
+	var resp *http.Response
+	var err error
+	for x := 0; x < max_retries; x++ {
+		resp, err = c.httpClient.Do(r)
+		if err == nil && resp.StatusCode == http.StatusOK {
+			return nil
+		}
+
+		// Back off to give us more chance of succeeding during a network blip
+		time.Sleep(backoff_millis * time.Millisecond)
+	}
+
+	if err != nil {
+		return fmt.Errorf("failed to send callback to %q. Error: %q", r.URL.String(), err)
+	}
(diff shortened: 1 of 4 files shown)


clients/callback_client.go
```diff
@@ -0,0 +1,111 @@
...
+	var resp *http.Response
+	var err error
+	for x := 0; x < max_retries; x++ {
+		resp, err = c.httpClient.Do(r)
+		if err == nil && resp.StatusCode == http.StatusOK {
+			return nil
+		}
+
+		// Back off to give us more chance of succeeding during a network blip
+		time.Sleep(backoff_millis * time.Millisecond)
```
leszko, line 34: So, in total we wait: 200 + 400 + 600 = 1.2s. Is it shorter than the frequency of the updates? I think if it's not possible that callback receives the status not in order.

leszko (approved):
Looks good. Added one comment. Other than that, LGTM 👍

clients/callback_client.go
thomshutt: That's a great point, I think we should handle out-of-order updates gracefully anyway (since they could happen even without this backoff). I'll add a timestamp field to the messages we send.

thomshutt: cc @victorges - something to bear in mind for you too

thomshutt: Also @leszko, it should just be ~600ms total - I kept it simple and didn't try to do an exponential backoff

thomshutt: Added timestamps now. I think Studio will probably still need to have some level of "state machine" knowledge though, at least in terms of terminal states (i.e if it gets an in-progress notifiction after an error/completion one, it can ignore it)

AlexKordic: @leszko also consider special cases: 
- Target machine is down 
- DNS lookup takes some time if not in cache
- If target server is overloaded and kernel accepts connection there is no automatic timeout to produce response

Actual connect may take minutes before error is returned. On my box sometimes 3 seconds sometimes 130 seconds:
```
$ ./test_connect 10.15.208.3 443
Connect error DNS=0 connect=3.04924
$ ./test_connect 192.168.1.2 443
Connect error DNS=0 connect=130.417
```

Maybe we should always use timeout on our side.

thomshutt: Great point @AlexKordic, I've added a timeout of 5s to the HTTP Client now

leszko (approved):

PR #15, opened 2022-08-09
Extract schema initialization out of handlers logic

### LINKED ISSUE #11: Separate JSON schema configuration and initialization from handlers logic
Currently JsonSchema configuration and instantiation is managed at the handler level. It would be better to isolate them into an init function for more readability [1].

Also, JsonSchemas will be quite long multiline strings bloating `handlers/handlers.go` file. We should use `embed` to extract them into local files [2].

Reference:
* [1] https://github.com/livepeer/catalyst-api/pull/4#discussion_r936077924
* [2] https://github.com/livepeer/catalyst-api/pull/4#discussion_r936079139

red-0ne:
* Separate JsonSchemas from handlers logic using `go:embed`
* Use `init` function to initialize JsonSchemas (located at `schemas/`) for all api handlers

#Close #11


handlers/handlers.go
```diff
@@ -118,5 +83,33 @@ func HasContentType(r *http.Request, mimetype string) bool {
...
+		}
+
+		schemaLoader := gojsonschema.NewStringLoader(string(schemaContent))
+		schema, err := gojsonschema.NewSchema(schemaLoader)
+
+		if err != nil {
+			glog.Fatal(err.Error())
+		}
+
+		schemaName := strings.TrimSuffix(file.Name(), ".json")
```
victorges, line 111: To be more generic (e.g. yaml)
```suggestion
		schemaName := strings.TrimSuffix(file.Name(), filepath.Ext(file.Name()))
```

handlers/schemas/UploadVOD.json
```diff
@@ -0,0 +1,66 @@
+{
```
victorges, line 1: Does `gojsonschema` work with YAML? This'd be so much cleaner if so
https://gist.github.com/victorges/723fb5e751a57b5e309412faa7cd8eed

handlers/handlers.go
```diff
@@ -9,11 +10,17 @@ import (
...
 	"strings"
 
+	"github.com/golang/glog"
 	"github.com/julienschmidt/httprouter"
 	"github.com/livepeer/catalyst-api/errors"
 	"github.com/xeipuuv/gojsonschema"
 )
 
+//go:embed schemas/*
+var schemasDir embed.FS
```
victorges, line 20: Nice! Didn't know embed supported dirs as well.

victorges (approved):
LGTM!

handlers/schemas/UploadVOD.json
```diff
@@ -0,0 +1,66 @@
...
+                            "url": {
+                                "type": "string",
+                                "format": "uri"
+                            }
+                        },
+                        "required": [
+                            "type",
+                            "url"
+                        ],
+                        "additional_properties": false
```
victorges, line 35: Just found out that this schema has a bug: It is allowing additional properties on every object.

The problem is that the JSON schema field is actually `additionalProperties` in camel case, and not `additional_properties`.

It's also weird that the JSON schema library does not validate the schema of the schemas it loads? lol
Tried finding a "strict" option in the JSON schema but didn't find it anywhere. Might be worth another look.

handlers/schemas/UploadVOD.json
red-0ne: There seems to be no straight forward way to make YAML schemas work. Any guidance would be helpful

victorges: Ah, don't worry too much about this, just a nice to have! If you wanna give it a shot tho, you could try using a lib [like this](https://github.com/kubernetes-sigs/yaml#installation-and-usage) which provides a `yaml.YAMLToJSON` helper.

red-0ne: Done

red-0ne:
@AlexKordic , you may want to take a look at this too, since it's rebased and touched your jsonSchema work done on `TranscodeSegment`.

AlexKordic:
My impression is that schema accompanies handler code. This change separates schema into external files. If we want tight coupling between schema and code using it, it's better to keep schema definition string inside binary, as code constant.
What are the reasons for having additional file that we need to find at runtime?

thomshutt (approved):
LGTM, I agree with moving it away from the handler and think the YAML is more readable but also don't love having to do the embedding. That is to say

handlers/schemas/UploadVOD.json
victorges: Nice!

victorges:
@AlexKordic Even though the `embed` code looks like it's reading files, the content of those separate files actually **are** embedded in the binary since we're using this: https://pkg.go.dev/embed

I agree we should keep the schemas close to the handlers code tho, but I actually like that they are not just constants in the code. Someone using the APIs for example could check what are the interfaces without having to read any `go` (even if just to find some constant with the schema). This feels like an intermediate step towards sth like an OpenAPI spec, from which we could also auto-generate clients in any language etc.

AlexKordic (approved):

PR #17, opened 2022-08-10
Define new API endpoint + test

AlexKordic:
- New handler `/api/transcode/file` added to `func StartCatalystAPIRouter()`
- Use `gojsonschema` to check input json.
- Just return `OK` for now and invoke callback.
- Testing request json format and required fields.
- Testing is callback invoked by handler.

### DIFF
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -4,6 +4,9 @@ import (
 	"encoding/json"
 	"log"
 	"net/http"
+	"strings"
+
+	"github.com/xeipuuv/gojsonschema"
 )
 
 type apiError struct {
@@ -36,4 +39,16 @@ func WriteHTTPUnsupportedMediaType(w http.ResponseWriter, msg string, err error)
 
 func WriteHTTPInternalServerError(w http.ResponseWriter, msg string, err error) apiError {
 	return writeHttpError(w, msg, http.StatusInternalServerError, err)
+}
+
+func WriteHTTPBadBodySchema(where string, w http.ResponseWriter, errors []gojsonschema.ResultError) apiError {
+	sb := strings.Builder{}
+	sb.WriteString("Body validation error in ")
+	sb.WriteString(where)
+	sb.WriteString(" ")
+	for i := 0; i < len(errors); i++ {
+		sb.WriteString(errors[i].String())
+		sb.WriteString(" ")
+	}
+	return writeHttpError(w, sb.String(), http.StatusBadRequest, nil)
 }
--- a/handlers/handlers.go
+++ b/handlers/handlers.go
@@ -24,6 +24,44 @@ func (d *CatalystAPIHandlersCollection) Ok() httprouter.Handle {
 	}
 }
 
+func (d *CatalystAPIHandlersCollection) TranscodeSegment() httprouter.Handle {
+	schemaLoader := gojsonschema.NewStringLoader(TranscodeSegmentRequestSchemaDefinition)
+
+	schema, err := gojsonschema.NewSchema(schemaLoader)
+	if err != nil {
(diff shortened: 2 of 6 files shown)


AlexKordic:
@thomshutt Wondering about repeated requests. How would entire system detect duplicate transcode jobs? Duplicate transcode request can have different callback URL, then we should also call this URL on completion.

handlers/handlers.go
```diff
@@ -24,6 +24,44 @@ func (d *CatalystAPIHandlersCollection) Ok() httprouter.Handle {
 	}
 }
 
+func (d *CatalystAPIHandlersCollection) TranscodeSegment() httprouter.Handle {
+	schemaLoader := gojsonschema.NewStringLoader(TranscodeSegmentRequestSchemaDefinition)
+
+	schema, err := gojsonschema.NewSchema(schemaLoader)
+	if err != nil {
+		panic(err)
```
thomshutt, line 32: Let's handle it a bit more gracefully

handlers/handlers.go
```diff
@@ -24,6 +24,44 @@ func (d *CatalystAPIHandlersCollection) Ok() httprouter.Handle {
...
+		if !result.Valid() {
+			errors.WriteHTTPBadBodySchema("TranscodeSegment", w, result.Errors())
+			return
+		}
+		if err := json.Unmarshal(payload, &transcodeRequest); err != nil {
+			errors.WriteHTTPBadRequest(w, "Invalid request payload", err)
+			return
+		}
+		if len(transcodeRequest.Profiles) == 0 {
+			// TODO: Check if this is correct. What if we only do detection?
```
thomshutt, line 55: I don't think it's required. We'll default to a standard set if not present

handlers/webhook.go
```diff
@@ -0,0 +1,72 @@
...
+
+type WebhookReceiver struct {
+	handler  func(w http.ResponseWriter, req *http.Request, _ httprouter.Params)
+	port     int
+	requests chan []byte
+	router   *httprouter.Router
+	server   *http.Server
+}
+
+func (s *WebhookReceiver) Init() {
```
thomshutt, line 23: The webhook exists in Studio and we call out to it. I don't think we need this (but might be missing something)

AlexKordic: Used just for callback test. Last point in #2552 spec. Do you want it removed?

handlers/handlers.go
AlexKordic: I could move the error on program start and panic early.. let me see how

AlexKordic: Fixed please check this [commit](https://github.com/livepeer/catalyst-api/pull/17/commits/c7d75babcfc286b86ce85ac000c1f18e0fead78a)

handlers/webhook.go
AlexKordic: Fixed

errors/errors.go
```diff
@@ -37,3 +40,15 @@ func WriteHTTPUnsupportedMediaType(w http.ResponseWriter, msg string, err error)
 func WriteHTTPInternalServerError(w http.ResponseWriter, msg string, err error) apiError {
 	return writeHttpError(w, msg, http.StatusInternalServerError, err)
 }
+
+func WriteHTTPBadBodySchema(where string, w http.ResponseWriter, errors []gojsonschema.ResultError) apiError {
+	sb := strings.Builder{}
+	sb.WriteString("Body validation error in ")
```
thomshutt, line 46: I don't see people use the String Builder very often in Go, normally stuff like this is done with a `fmt.Sprintf`

thomshutt (approved):

PR #20, opened 2022-08-12
Process Upload VOD

leszko:
WIP

- [x] Upload VOD process
- [x] Interacting with Callback API
- [ ] Check if it works with S3 stream source
- [ ] Refactor clients
- [ ] Add unit tests (with mocked clients)
- [ ] Add timeout with stream cleanup
- [x] Remove temporary code for local testing (auth, TestWorkflow)

PR #21, opened 2022-08-12
Run on 0.0.0.0 instead of localhost (#19)

victorges:
Re-merging this one cause we ended up reverting it on a rebase in #20


red-0ne (approved):
All good here

PR #22, opened 2022-08-15
Replace localhost with 0.0.0.0

thomshutt:


cmd/http-server/http-server.go
```diff
@@ -28,8 +28,8 @@ func main() {
 	}
 
 	mc := &handlers.MistClient{
-		ApiUrl:          fmt.Sprintf("http://localhost:%d/api2", *mistPort),
-		TriggerCallback: fmt.Sprintf("http://localhost:%d/api/mist/trigger", *port),
+		ApiUrl:          fmt.Sprintf("http://0.0.0.0:%d/api2", *mistPort),
+		TriggerCallback: fmt.Sprintf("http://0.0.0.0:%d/api/mist/trigger", *port),
```
victorges, line 32: Is this right? I thought the communication between catalyst-api and mist was made via localhost. Also, if we try to call 0.0.0.0 it's probably gonna fail, it's only a special value to mean _listening_ on any IP, isn't it?

handlers/handlers_test.go
```diff
@@ -32,12 +32,12 @@ func TestSuccessfulVODUploadHandler(t *testing.T) {
...
 	catalystApiHandlers := CatalystAPIHandlersCollection{MistClient: StubMistClient{}, StreamCache: make(map[string]StreamInfo)}
 	var jsonData = `{
-		"url": "http://localhost/input",
+		"url": "http://0.0.0.0/input",
 		"callback_url": "CALLBACK_URL",
 		"output_locations": [
 			{
 				"type": "object_store",
-				"url": "memory://localhost/output",
+				"url": "memory://0.0.0.0/output",
```
victorges, line 40: I think this host is ignored, but feels very weird to have memory://0.0.0.0. Might as well keep it localhost (and enforcing on parsing perhaps?)

victorges (commented):
I'm not sure I understood the need for this. The new replacements look either weird (memory://0.0.0.0) or seem like they won't work (calling mist on 0.0.0.0). Could you add some context in the PR description?

thomshutt:
Nevermind, this was me doing a blanket find/replace after seeing your comment this morning, but then realised you'd already reapplied the place where it's needed

PR #26, opened 2022-08-15
Switch to HTTP retry library

thomshutt:
We were getting errors because of not resetting the POST body correctly before retrying, so decided to switch to a library and not have to think about that stuff.

### DIFF
--- a/clients/callback_client.go
+++ b/clients/callback_client.go
@@ -7,41 +7,39 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/hashicorp/go-retryablehttp"
 	"github.com/livepeer/catalyst-api/config"
 )
 
-const max_retries = 3
-const backoff_millis = 200
-
 type CallbackClient struct {
-	httpClient *http.Client
+	httpClient *retryablehttp.Client
 }
 
 func NewCallbackClient() CallbackClient {
+	client := retryablehttp.NewClient()
+	client.RetryMax = 2                          // Retry a maximum of this+1 times
+	client.RetryWaitMin = 200 * time.Millisecond // Wait at least this long between retries
+	client.RetryWaitMax = 1 * time.Second        // Wait at most this long between retries (exponential backoff)
+	client.HTTPClient = &http.Client{
+		Timeout: 5 * time.Second, // Give up on requests that take more than this long
+	}
+
 	return CallbackClient{
-		httpClient: &http.Client{
-			Timeout: 5 * time.Second,
-		},
+		httpClient: client,
 	}
 }
 
-func (c CallbackClient) DoWithRetries(r *http.Request) error {
-	var resp *http.Response
-	var err error
-	for x := 0; x < max_retries; x++ {
-		resp, err = c.httpClient.Do(r)
-		if err == nil && resp.StatusCode == http.StatusOK {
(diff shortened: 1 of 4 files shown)


clients/callback_client.go
```diff
@@ -7,41 +7,39 @@ import (
...
+	"github.com/hashicorp/go-retryablehttp"
 	"github.com/livepeer/catalyst-api/config"
 )
 
-const max_retries = 3
-const backoff_millis = 200
-
 type CallbackClient struct {
-	httpClient *http.Client
+	httpClient *retryablehttp.Client
```
victorges, line 15: WDYT of keeping this as a regular `http.Client`, just so we can switch from the retry lib more seamlessly?

clients/callback_client.go
```diff
@@ -7,41 +7,39 @@ import (
...
 
-	if err != nil {
-		return fmt.Errorf("failed to send callback to %q. Error: %q", r.URL.String(), err)
+	if resp.StatusCode >= 400 {
+		return fmt.Errorf("failed to send callback to %q. HTTP Code: %d", r.URL.String(), resp.StatusCode)
 	}
-	return fmt.Errorf("failed to send callback to %q. Response Status Code: %d", r.URL.String(), resp.StatusCode)
+
+	return nil
 }
```
victorges, line 43: Is this still necessary? I think it could be replaced by a simpler helper like `responseToError(res *http.Response) error`

victorges (approved):
Nice! Let's make `catalyst-api` the testground for this lib and consider applying it in most of our services (especially `go-livepeer` which handles large payloads all around)

clients/callback_client.go
```diff
@@ -7,41 +7,39 @@ import (
...
-			return nil
-		}
-
-		// Back off to give us more chance of succeeding during a network blip
-		time.Sleep(backoff_millis * time.Millisecond)
+func (c CallbackClient) DoWithRetries(r *retryablehttp.Request) error {
+	resp, err := c.httpClient.Do(r)
+	if err != nil {
+		return fmt.Errorf("failed to send callback to %q. Error: %s", r.URL.String(), err)
 	}
```
victorges, line 36: Ah we likely also need to close the response body to avoid leaks.
```suggestion
	}
	defer resp.Body.Close()
```

thomshutt: Good catch, I always forget that

clients/callback_client.go
thomshutt: Not sure I 100% understand where the helper would sit, but we do still need the StatusCode check since the retry library only errors on `>= 500`

clients/callback_client.go
thomshutt: `http.Client` isn't an interface though, how would we switch between the two?

victorges: There's apparently a function in the retry client to return a corresponding http.Client! I saw it mentioned in the end of their readme I think

thomshutt: Huh nice, I had seen that but had missed that it includes the retry functionality

> StandardClient returns a stdlib *http.Client with a custom Transport, which shims in a *retryablehttp.Client for added retries. 

Will make a new PR for that

PR #29, opened 2022-08-18
Document local integration testing / development

thomshutt:


leszko (approved):
In the future, I guess we need to bundle it inside the `livepeer/catalyst` and remove these instructions. Is that correct?

thomshutt:
> In the future, I guess we need to bundle it inside the `livepeer/catalyst` and remove these instructions. Is that correct?

I don't think so, unless we want to move to a monorepo type setup

PR #30, opened 2022-08-24
transcoding endpoint implementation

AlexKordic:
## workflow

- Studio calls HTTP `/api/transcode/file` endpoint
- Add new SOURCE stream on Mist server having S3 input URL
- RENDITIONS stream is already configured on Mist server along with triggers `LIVE_TRACK_LIST`,`PUSH_END`
- Start MistProcLivepeer subprocess from go code
- SOURCE stream consumes S3 input and creates video & audio tracks in Mist server
- MistProcLivepeer sends INPUT to broadcaster to transcode
- MistProcLivepeer ingests transcoded tracks into RENDITIONS stream
- API responds to `LIVE_TRACK_LIST` trigger. 
    - For each video track API starts push of video track + audio track to S3 destination URL
        - When rendition is stored to S3, API responds to `PUSH_END` trigger and invokes callback URL
- When all rendition streams are uploaded callback API is invoked
- If error happens at any stage callback API is invoked. Multiple errors are possible even in case of completed transcode (success)

## How to test server

Starting API server and using curl:
- `go run cmd/http-server/http-server.go`
- `curl -v --header "Content-Type: application/json" --header "Authorization: Bearer IAmAuthorized" --request POST --data @../req.json "http://localhost:4949/api/transcode/file"`

## How to run TestSegmentTranscode

Setup repos:
- `mistserver` - `catalyst` branch
- `catalyst-api` - [ak/transcoding](https://github.com/livepeer/catalyst-api/tree/ak/transcoding) branch
- `go-livepeer` - `mock_transcoder` branch

Start broadcaster node:
- `cd go-livepeer`
- `go run cmd/offchain_transcoder/transcoder.go`

Apply `catalyst-api/mist_transcoding.json` to Mist config (for ex. `mistserver/startup.json`)

Start Mist:
- `cd mistserver/build`
- `./MistController --debug=8 -c ../startup.json`

Set S3 credentials, replace with proper value:
- `export GCP_SECRETS="some-s3-id:some-s3-secret"`

Temporary fix:
- `sudo ln -s /dev/null /opt/null.ts`

Start integration test:
- `cd catalyst-api/handlers`
- `go test . -v -count 1 -tags externaldeps -run TestSegmentTranscode`


handlers/handlers.go
```diff
@@ -55,12 +58,78 @@ func (d *CatalystAPIHandlersCollection) TranscodeSegment() httprouter.Handle {
...
+			errorOutInternal("error start transcodeCommand", err)
+			return
+		}
+
+		// TODO: remove when Mist code is updated <
+		// Starting SOURCE_PREFIX stream. Why we need to start it manually? MistProcLivepeer should start it??
+		if err := d.MistClient.PushStart(inputStream, "/opt/null.ts"); err != nil {
+			errorOutInternal("error PushStart(inputStream)", err)
+			return
+		}
```
AlexKordic, line 116: Waiting on @Thulinma changes to remove this push step.

leszko:
@AlexKordic I think 2 reviewers should be enough for this PR. Could you please remove someone? Thanks!

handlers/client.go
```diff
@@ -25,11 +26,38 @@ type MistClient struct {
...
 }
 
+func (mc *MistClient) RemoveAllStreams() error {
+	c := RemoveAllStreamsCommand{Streams: map[string]Stream{}}
+	response, err := mc.sendCommand(c)
+	fmt.Printf("RemoveAllStreams %s\n", response)
+	return err
+}
+
+func (mc *MistClient) DeleteAllTriggers() error {
```
leszko, line 36: What is the use of this function when we already have `DeleteAllTriggersFor()` deleting all triggers for the given stream?

handlers/handlers.go
```diff
@@ -17,25 +18,28 @@ import (
...
 
 func (d *CatalystAPIHandlersCollection) Ok() httprouter.Handle {
-	return func(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	return func(w http.ResponseWriter, _ *http.Request, _ httprouter.Params) {
 		io.WriteString(w, "OK")
 	}
 }
 
-func (d *CatalystAPIHandlersCollection) TranscodeSegment() httprouter.Handle {
+// Takes mpegts segment as input, ingests in Mist under SOURCE_PREFIX streamName, transcodes into RENDITION_PREFIX stream then pushes renditions to specified destination.
```
leszko, line 32: Nit: Start function comments with the function name. That's the convention in Go. So here, it should start like `TranscodeSegment takes mpegts...`.

handlers/handlers.go
```diff
@@ -55,12 +59,79 @@ func (d *CatalystAPIHandlersCollection) TranscodeSegment() httprouter.Handle {
...
 		}
-
 		callbackClient := clients.NewCallbackClient()
-		if err := callbackClient.SendTranscodeStatusError(transcodeRequest.CallbackUrl, "NYI - not yet implemented"); err != nil {
-			errors.WriteHTTPInternalServerError(w, "error send transcode error", err)
+		errorOutInternal := func(where string, err error) {
+			// Mist does not respond or handle returned error codes. We do this for correctness.
+			errors.WriteHTTPInternalServerError(w, where, err)
+			if err := callbackClient.SendSegmentTranscodeError(transcodeRequest.CallbackUrl, where, err.Error(), transcodeRequest.SourceFile); err != nil {
+				fmt.Printf("error send transcode error %v\n", err)
```
leszko, line 67: leftover?

Or if you want to log, then use the logger.

handlers/handlers.go
```diff
@@ -55,12 +59,79 @@ func (d *CatalystAPIHandlersCollection) TranscodeSegment() httprouter.Handle {
...
+				return
+			}
+		}
+
+		inputStream, renditionsStream := generateStreamNames()
+		if err = d.MistClient.AddStream(inputStream, transcodeRequest.SourceFile); err != nil {
+			errorOutInternal("error AddStream(inputStream)", err)
```
leszko, line 74: Nit: Maybe we should include the stream name to help with the analysis?

handlers/handlers.go
```diff
@@ -55,12 +59,79 @@ func (d *CatalystAPIHandlersCollection) TranscodeSegment() httprouter.Handle {
...
+			return
+		}
+
+		// Start MistProcLivepeer
+		configPayload, err := json.Marshal(configForSubprocess(&transcodeRequest, broadcasterPort, inputStream, renditionsStream))
+		if err != nil {
+			errorOutInternal("error ProcLivepeerConfig json encode", err)
+			return
+		}
+		fmt.Printf("configPayload %s\n", configPayload)
```
leszko, line 84: leftover?

handlers/handlers.go
```diff
@@ -17,25 +18,28 @@ import (
...
+// Takes mpegts segment as input, ingests in Mist under SOURCE_PREFIX streamName, transcodes into RENDITION_PREFIX stream then pushes renditions to specified destination.
+// Manually starts MistLivepeerProc binary. For this we require port of B-node that listens on 127.0.0.1 .
+// When `MistProcLivepeer` starts, SOURCE_PREFIX stream is taken as input and (will automatically - WIP) start SOURCE_PREFIX stream.
+// No trigger is registered on SOURCE_PREFIX so configured sourceUrl is used.
+// Also `MistProcLivepeer` outputs to RENDITION_PREFIX stream where we listen for triggers to start new push to rendition destination.
+// Because RENDITION_PREFIX stream contains multiple video tracks we start multiple push-es each selecting one video and one audio track to push to S3.
+func (d *CatalystAPIHandlersCollection) TranscodeSegment(broadcasterPort int, mistProcPath string) httprouter.Handle {
 	schema := inputSchemasCompiled["TranscodeSegment"]
 
 	return func(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
```
leszko, line 41: For the sake of readability, could you try to split this function into multiple functions? Or at least extract a few functions from it? E.g. the logic related to starting the `MistProcLivepeer` could be moved to a separate function.

handlers/handlers.go
leszko: Ok, then should this PR be converted to draft and merged after the changes to @Thulinma ? I think we should try to avoid such `TODOs` in the `main` branch.

handlers/handlers.go
```diff
@@ -17,25 +18,28 @@ import (
...
 		io.WriteString(w, "OK")
 	}
 }
 
-func (d *CatalystAPIHandlersCollection) TranscodeSegment() httprouter.Handle {
+// Takes mpegts segment as input, ingests in Mist under SOURCE_PREFIX streamName, transcodes into RENDITION_PREFIX stream then pushes renditions to specified destination.
+// Manually starts MistLivepeerProc binary. For this we require port of B-node that listens on 127.0.0.1 .
+// When `MistProcLivepeer` starts, SOURCE_PREFIX stream is taken as input and (will automatically - WIP) start SOURCE_PREFIX stream.
+// No trigger is registered on SOURCE_PREFIX so configured sourceUrl is used.
+// Also `MistProcLivepeer` outputs to RENDITION_PREFIX stream where we listen for triggers to start new push to rendition destination.
```
leszko, line 36: I'm missing the context, but why is `MistLivepeeerProc` a separate binary? What is its purpose? I think you could add this explanation here. Or even better, move the code related to starting this binary to a separate function and add this explanation there.

handlers/handlers.go
```diff
@@ -55,12 +59,79 @@ func (d *CatalystAPIHandlersCollection) TranscodeSegment() httprouter.Handle {
...
+
+		// TODO: remove when Mist code is updated <
+		// Starting SOURCE_PREFIX stream. Why we need to start it manually? MistProcLivepeer should start it??
+		if err := d.MistClient.PushStart(inputStream, "/opt/null.ts"); err != nil {
+			errorOutInternal("error PushStart(inputStream)", err)
+			return
+		}
+
+		// Create entry in our pushes books, storing callback url and source url.
+		// Later triggers will use this info and add destination urls.
```
leszko, line 120:
```suggestion
		// Cache the stream data, later used in the trigger handlers called by Mist
```
Nit

handlers/handlers.go
```diff
@@ -55,12 +59,79 @@ func (d *CatalystAPIHandlersCollection) TranscodeSegment() httprouter.Handle {
...
+		d.StreamCache.Transcoding.Store(renditionsStream, SegmentInfo{
+			CallbackUrl: transcodeRequest.CallbackUrl,
+			Source:      transcodeRequest.SourceFile,
+			UploadDir:   path.Dir(transcodeRequest.SourceFile),
+		})
+
+		err = transcodeCommand.Wait()
```
leszko, line 127: So, this HTTP call will be sync, right? How long to we expect to wait here? Won't the caller (studio?) time out on this call?

handlers/handlers.go
```diff
@@ -55,12 +59,79 @@ func (d *CatalystAPIHandlersCollection) TranscodeSegment() httprouter.Handle {
...
+		if exit, ok := err.(*exec.ExitError); ok {
+			fmt.Printf("MistProcLivepeer returned %d\n", exit.ExitCode())
```
leszko, line 129: For logging please use `log` instead of `fmt`.

handlers/handlers.go
```diff
@@ -150,62 +202,203 @@ func (d *CatalystAPIHandlersCollection) processUploadVOD(streamName, sourceURL,
...
-		errS := d.MistClient.DeleteStream(s)
-		if errT != nil {
-			errors.WriteHTTPInternalServerError(w, fmt.Sprintf("Cannot remove PUSH_END trigger for stream '%s'", s), errT)
+	payload, err := io.ReadAll(req.Body)
+	if err != nil {
+		errorOutInternal("Cannot read payload", err)
+		return
+	}
+	lines := strings.Split(strings.TrimSuffix(string(payload), "\n"), "\n")
+	if len(lines) < 2 {
```
leszko, line 310: I think this condition should be different, because you read more lines than `2`.

handlers/handlers.go
```diff
@@ -150,62 +202,203 @@ func (d *CatalystAPIHandlersCollection) processUploadVOD(streamName, sourceURL,
...
+}
+
+// Only single trigger callback is allowed on Mist.
+// All created streams and our handlers (segmenting, transcoding, et.) must share this endpoint.
+// If handler logic grows more complicated we may consider adding dispatch mechanism here.
+func (d *MistCallbackHandlersCollection) Trigger() httprouter.Handle {
+	return func(w http.ResponseWriter, req *http.Request, params httprouter.Params) {
+		whichOne := req.Header.Get("X-Trigger")
+		switch whichOne {
+		case "PUSH_END":
```
leszko, line 391: Nice, that you've split it here in many functions 🚀

handlers/transcode.go
```diff
@@ -1,17 +1,44 @@
...
+	inputStream := SOURCE_PREFIX + suffix
+	renditionsStream := RENDITION_PREFIX + suffix
+	return inputStream, renditionsStream
+}
+
+func isTranscodeStream(streamName string) (bool, string) {
+	if strings.HasPrefix(streamName, RENDITION_PREFIX) {
+		return true, streamName[len(RENDITION_PREFIX):]
+	}
+	// if strings.HasPrefix(streamName, SOURCE_PREFIX) {
```
leszko, line 24: What is this commented-out code?

leszko (changes requested):
@AlexKordic Thanks for the PR. I did partly reviewed it. Added some first comments.

I suggest to split this PR into multiple PRs, because it's very hard to do a good review with so many changes.

Also, IFAIU, there are some WIP comments in the PR, so we don't plan to merge it right now. Is that the case? If yes, then, maybe better to covert this PR into draft?

clients/callback_client.go
```diff
@@ -13,6 +13,7 @@ import (
 
 type CallbackClient struct {
 	httpClient *http.Client
+	Errors     chan error
```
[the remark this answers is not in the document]
AlexKordic: Since network operation is moved to separate goroutine i wanted to keep the original functionality. This channel serves to store result of async operation. do you think we should remove it until it is used?

handlers/stream_cache.go
```diff
@@ -0,0 +1,117 @@
...
+	return SegmentInfo{}, fmt.Errorf("cache mismatch for %s", streamName)
+}
+
+func (c *TranscodingCache) AddDestination(streamName, destination string) {
+	c.mutex.Lock()
+	info, ok := c.pushes[streamName]
+	if ok {
+		info.Destionations = append(info.Destionations, destination)
+	}
+	c.mutex.Unlock()
```
thomshutt, line 74: Should this lock be deferred?

AlexKordic: hm, is there any difference? i am not familiar with panic yet

clients/callback_client.go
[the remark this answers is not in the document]
thomshutt: I think we can remove it and just log if the callback fails - there shouldn't be any other reason we need to know about callback errors

handlers/transcode_test.go
```diff
@@ -0,0 +1,162 @@
+//go:build externaldeps
+// +build externaldeps
+
+/* Depends:
```
thomshutt, line 4: I think there's too much environmental dependency here for it to be useful as a unit test. Since you've tested and are happy the new functionality works, can we get rid of this / break into down into unit tests with e.g a fake Mist process

cmd/http-server/http-server.go
```diff
@@ -17,9 +17,13 @@ import (
 )
 
 func main() {
-	port := flag.Int("port", 4949, "Port to listen on")
 	mistPort := flag.Int("mist-port", 4242, "Port to listen on")
+	port := flag.Int("port", 4949, "Port to listen on")
 	mistJson := flag.Bool("j", false, "Print application info as JSON. Used by Mist to present flags in its UI.")
+	// Transcode endpoint starts MistProcLivepeer as subprocess so we need path to the binary(mistlp) and port of local Broadcaster(b-port)
+	bPort := flag.Int("b-port", 8935, "Port of local B node")
```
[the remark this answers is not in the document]
AlexKordic: I am unfamiliar with deploy procedure. It has proper default value, why should we remove it from parameters?

cmd/http-server/http-server.go
```diff
@@ -17,9 +17,13 @@ import (
...
+	mistProcPath := flag.String("mistlp", "../mistserver/build/MistProcLivepeer", "path to MistProcLivepeer")
```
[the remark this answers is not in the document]
AlexKordic: Agree. Does anyone know what is proper default value at this point?

handlers/client.go
AlexKordic: Removing..
Actually there is a bug when removing last trigger, as json field is defined as `omitempty`.
Was used when we were adding triggers from API server. 
Mist good practice is to define triggers via config json.
`AddTrigger()` and `DeleteTrigger()` are also planned for removal.

handlers/handlers.go
```diff
@@ -17,25 +18,28 @@ import (
...
-	return func(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	return func(w http.ResponseWriter, _ *http.Request, _ httprouter.Params) {
 		io.WriteString(w, "OK")
 	}
 }
 
-func (d *CatalystAPIHandlersCollection) TranscodeSegment() httprouter.Handle {
+// Takes mpegts segment as input, ingests in Mist under SOURCE_PREFIX streamName, transcodes into RENDITION_PREFIX stream then pushes renditions to specified destination.
+// Manually starts MistLivepeerProc binary. For this we require port of B-node that listens on 127.0.0.1 .
+// When `MistProcLivepeer` starts, SOURCE_PREFIX stream is taken as input and (will automatically - WIP) start SOURCE_PREFIX stream.
```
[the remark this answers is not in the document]
AlexKordic: actually, workaround. WIP is on Mist side.

handlers/handlers.go
AlexKordic: Great comment. I really prefer splitting into multiple functions.

handlers/handlers.go
AlexKordic: will use separate goroutine for transcoding operation

handlers/transcode.go
AlexKordic: old code, removing

cmd/http-server/http-server.go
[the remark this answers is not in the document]
leszko: My point is just that if you plan to never change this value, then it should not be a flag, but rather a hardcoded const. But if there is any use-case for changing this port, then we can keep it as a flag.

cmd/http-server/http-server.go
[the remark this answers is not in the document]
leszko: I don't know. But I think that if we have no clue, then I'd suggest to keep the current directory, so just `./MistProcLivepeer`.

cmd/http-server/http-server.go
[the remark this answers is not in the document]
AlexKordic: I see your point. Still don't have an answer. Let me ask Thom

handlers/handlers.go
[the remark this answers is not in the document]
leszko: Ok, I see. So, from this code's perspective it's not WIP, it's more like a workaround. To avoid confusion, I suggest to change it to something like:

```
// TODO: Remove `MistProcLivepeer` after this issue is fixed <link-to-related-mist-gh-isssue>
// `MistProcLivepeer` is a workaround until the automatic stream start is implemented in Mist 
```

handlers/handlers.go
```diff
@@ -17,50 +17,38 @@ import (
...
-	return func(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	return func(w http.ResponseWriter, _ *http.Request, _ httprouter.Params) {
 		io.WriteString(w, "OK")
 	}
 }
 
-func (d *CatalystAPIHandlersCollection) TranscodeSegment() httprouter.Handle {
-	schema := inputSchemasCompiled["TranscodeSegment"]
-
+// TranscodeSegment takes mpegts segment as input, ingests in Mist under SOURCE_PREFIX streamName, transcodes into RENDITION_PREFIX stream then pushes renditions to specified destination.
```
leszko, line 31: I think this comment is very important, because it explains the whole logic of the `Mist <> Catalyst API` interactions for the transcoding part. Therefore, I suggest to improve it slightly to look something like this:

```
// TranscodeSegment takes mpegts segment as input and asynchronously transcode it into the specified destination.
//
// The process looks as follows:
// 1. Add stream to Mist
// 2. Execute the `MistProceLivepeer` process to perform the following operations:
//   - Start SOURCE_PREFIX stream transcoding in Mist
//     - ...
```

handlers/handlers.go
```diff
@@ -150,62 +119,197 @@ func (d *CatalystAPIHandlersCollection) processUploadVOD(streamName, sourceURL,
...
-	StreamCache map[string]StreamInfo
+	StreamCache *StreamCache
 }
 
-func (d *MistCallbackHandlersCollection) Trigger() httprouter.Handle {
-	return func(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
-		if t := req.Header.Get("X-Trigger"); t != "PUSH_END" {
-			errors.WriteHTTPBadRequest(w, "Unsupported X-Trigger", fmt.Errorf("unknown trigger '%s'", t))
-			return
+// This trigger is stream-specific and must be blocking.
```
leszko, line 141: Please change all comments to follow Golang standards. In this case, it should start from the function name.

handlers/transcode.go
```diff
@@ -36,3 +68,208 @@ type TranscodeSegmentRequest struct {
...
+		t.errorOut("error stdinPipe.Close()", err)
+		return
+	}
+	err = transcodeCommand.Start()
+	if err != nil {
+		t.errorOut("error start transcodeCommand", err)
+		return
+	}
+
+	// TODO: remove when Mist code is updated <
```
leszko, line 205: Please add the related Mist GH Issue link.

handlers/transcode.go
```diff
@@ -36,3 +68,208 @@ type TranscodeSegmentRequest struct {
...
+	// Starting SOURCE_PREFIX stream. Why we need to start it manually? MistProcLivepeer should start it??
```
leszko, line 206: I don't get this questions, could you please clarify this comment or remove it?

handlers/handlers.go
```diff
@@ -150,62 +119,197 @@ func (d *CatalystAPIHandlersCollection) processUploadVOD(streamName, sourceURL,
...
+// The payload for this trigger is multiple lines, each separated by a single newline character
+// (without an ending newline), containing data as such: `stream name`\n`push target URI`
+func (d *MistCallbackHandlersCollection) Trigger_LIVE_TRACK_LIST(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
```
leszko, line 144: Nit: Move this function under `Trigger()` to follow the [Stepdown Rule](https://dzone.com/articles/the-stepdown-rule)

handlers/handlers.go
```diff
@@ -150,62 +119,197 @@ func (d *CatalystAPIHandlersCollection) processUploadVOD(streamName, sourceURL,
...
+	// Following code is responding to transcoding handler:
```
leszko, line 145: Nit: I think you can remove this code or move to the comment above the function.

handlers/handlers.go
```diff
@@ -150,62 +119,197 @@ func (d *CatalystAPIHandlersCollection) processUploadVOD(streamName, sourceURL,
...
 }
 
-func (d *MistCallbackHandlersCollection) Trigger() httprouter.Handle {
-	return func(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
-		if t := req.Header.Get("X-Trigger"); t != "PUSH_END" {
-			errors.WriteHTTPBadRequest(w, "Unsupported X-Trigger", fmt.Errorf("unknown trigger '%s'", t))
-			return
+// This trigger is stream-specific and must be blocking.
+// The payload for this trigger is multiple lines, each separated by a single newline character
+// (without an ending newline), containing data as such: `stream name`\n`push target URI`
```
leszko, line 143: I think it's hard to grasp where in the whole transcoding process the `LIVE_TRACK_LIST` trigger is called. Could you add this information to the `TranscodeSegment` function comment. I think we need a place which described the whole of process of interactions between `Catalyst API <> MistLivepeeerProc <> Mist`.

handlers/handlers.go
```diff
@@ -17,50 +17,38 @@ import (
...
 )
 
-type StreamInfo struct {
-	callbackUrl string
-}
-
 type CatalystAPIHandlersCollection struct {
 	MistClient  MistAPIClient
-	StreamCache map[string]StreamInfo
+	StreamCache *StreamCache
```
leszko, line 22: I think we need to add somewhere the cleanup for this `StreamCache`. Currently we just add values here, but if there is any error in this whole process, then we may end up with some streams handing in Mist. 

Saying that, I think it's not part of this PR, but I suggest to create a separate GH Issue for it.

handlers/handlers.go
```diff
@@ -150,62 +119,197 @@ func (d *CatalystAPIHandlersCollection) processUploadVOD(streamName, sourceURL,
...
+		errors.WriteHTTPBadRequest(w, "Bad request payload", fmt.Errorf("unknown payload '%s'", string(payload)))
+		return
+	}
+	// stream name is the second line in the Mist Trigger payload
+	streamName = lines[1]
+	destination := lines[2]
+	actualDestination := lines[3]
+	pushStatus := lines[5]
+	if yes, _ := isTranscodeStream(streamName); yes {
+		// Following code is responding to transcoding handler: TODO: encapsulate
```
leszko, line 235: 1. Could you address this `TODO`?
2. Also, the same comment as before, we need a place in the comments to explain the whole flow, because it's very hard to grasp the whole logic.

leszko (changes requested):
Added some more comments. I think the most important is to describe in some comment the whole flow, because it's very hard to understand how the flow between `Catalyst API <> MistLivepeeerProc <> Mist` looks like.

handlers/stream_cache.go
```diff
@@ -0,0 +1,123 @@
+package handlers
+
+import (
+	"fmt"
+	"sync"
+)
+
+// Shared server state. Each pipeline (usually endpoint) having separate books.
```
leszko, line 8: The same comment about comments 🙃 Please use the Golang convections for comments.

handlers/stream_cache.go
```diff
@@ -0,0 +1,123 @@
...
+func (c *SegmentingCache) GetCallbackUrl(streamName string) (string, error) {
+	info, ok := c.cache[streamName]
+	if ok {
+		return info.callbackUrl, nil
+	}
+	return "", fmt.Errorf("cache mismatch for %s", streamName)
+}
+
+func (c *SegmentingCache) Store(streamName, callbackUrl string) {
+	c.mutex.Lock()
```
leszko, line 115: If you use this mutex for storing (so we allow concurrent cache modification), then I guess you'll also need to add it to `Remove()` and `GetCallbackUrl()`. Isn't that the case?

handlers/stream_cache.go
```diff
@@ -0,0 +1,123 @@
...
+type TranscodingCache struct {
+	pushes map[string]*SegmentInfo
+	mutex  sync.Mutex
+}
+
+type SegmentInfo struct {
+	CallbackUrl   string
+	Source        string   // S3 input we are transcoding
+	UploadDir     string   // S3 destination url for multiple renditions
+	Destionations []string // Rendition URLS go here on push start and removed on push end
```
leszko, line 32: Typo

handlers/stream_cache.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func (c *TranscodingCache) Init() {
+	c.pushes = make(map[string]*SegmentInfo)
+}
+
+type Empty = bool
```
leszko, line 39: Nit: I'm not sure you need it. But if yes, then maybe call it `IsEmpty`?

handlers/stream_cache.go
AlexKordic: Yes, good catch

handlers/stream_cache.go
AlexKordic: I keep missing this..

handlers/stream_cache.go
AlexKordic: 1. typo to fix
2. Kind of no, let me explain

Note that one input segment is transcoded into multiple rendition segments. We use single `UploadDir` as base for rendition URLs. Each rendition gets unique name and final URL is stored as record in `Destinations` when upload starts. On upload end we remove record from `Destinations` and finally on `Destinations` empty we remove `SegmentInfo` struct.

leszko: Ok, then it makes sense!

handlers/mist_triggers.go
```diff
@@ -0,0 +1,207 @@
...
+	StreamCache *StreamCache
+}
+
+// Trigger dispatches request to mapped method according to trigger name
+// Only single trigger callback is allowed on Mist.
+// All created streams and our handlers (segmenting, transcoding, et.) must share this endpoint.
+// If handler logic grows more complicated we may consider adding dispatch mechanism here.
+func (d *MistCallbackHandlersCollection) Trigger() httprouter.Handle {
+	return func(w http.ResponseWriter, req *http.Request, params httprouter.Params) {
+		t := &Trigger{w: w, req: req}
```
leszko, line 27: Do we need this type `Trigger` at all? I think you can pass `w` and `req` directly. To me this type looks weird, because it has one additional field `streamName` which is sometimes filled, sometimes now. So, I'd suggest just keeping `w`, `req`, and `streamName` as separate variables.

handlers/mist_triggers.go
```diff
@@ -0,0 +1,207 @@
...
+	t.streamName = lines[0]
+	encodedTracks := lines[1]
+
+	yes, suffix := isTranscodeStream(t.streamName)
+	if !yes {
+		t.Err("unknown streamName format", nil)
+		return
+	}
+
+	if streamEnded := encodedTracks == "null"; streamEnded {
```
leszko, line 63: Maybe just like this?
```
if encodedTracks == "null" {
...
}
```

handlers/mist_triggers.go
```diff
@@ -0,0 +1,207 @@
...
+		// SOURCE_PREFIX stream is no longer needed
+		inputStream := fmt.Sprintf("%s%s", SOURCE_PREFIX, suffix)
+		if err = d.MistClient.DeleteStream(inputStream); err != nil {
+			log.Printf("ERROR LIVE_TRACK_LIST DeleteStream(%s) %v", inputStream, err)
+		}
+		// Multiple pushes from RENDITION_PREFIX are in progress.
```
leszko, line 69: I don't get this comment. We try to remove a stream from Mist and it returns an error. Does it mean that we're currently processing some other rendition segments? 🤔

handlers/mist_triggers.go
```diff
@@ -0,0 +1,207 @@
...
+		t.Err("LiveTrackListTriggerJson json decode error", err)
+		return
+	}
+	// Start push per each video track
+	info, err := d.StreamCache.Transcoding.Get(t.streamName)
+	if err != nil {
+		t.Err("LIVE_TRACK_LIST unknown push source", err)
+		return
+	}
+	for i := range tracks { // i is generated name, not important, all info is contained in element
```
leszko, line 83: If `i` is not important, then why do we use in the line 84?

handlers/mist_triggers.go
```diff
@@ -0,0 +1,207 @@
...
+		if tracks[i].Type != "video" {
+			// Only produce an rendition per each video track, selecting best audio track
+			continue
+		}
+		destination := fmt.Sprintf("%s/%s__%dx%d.ts?video=%d&audio=maxbps", info.UploadDir, t.streamName, tracks[i].Width, tracks[i].Height, tracks[i].Index) //.Id)
+		if err := d.MistClient.PushStart(t.streamName, destination); err != nil {
+			log.Printf("> ERROR push to %s %v", destination, err)
+		} else {
+			d.StreamCache.Transcoding.AddDestination(t.streamName, destination)
```
leszko, line 92: Why if `PushStart()` returns an error, then we add a new destination? I think it's not clear 🤔

handlers/mist_triggers.go
AlexKordic: The information can be in the comment also:
```go
if encodedTracks == "null" {
  // stream ended, take action...
```

But condition variable makes it more readable

handlers/mist_triggers.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+	// Start push per each video track
+	info, err := d.StreamCache.Transcoding.Get(t.streamName)
+	if err != nil {
+		t.Err("LIVE_TRACK_LIST unknown push source", err)
+		return
+	}
+	for i := range tracks { // i is generated name, not important, all info is contained in element
+		if tracks[i].Type != "video" {
+			// Only produce an rendition per each video track, selecting best audio track
```
leszko, line 85: I don't get this comment. So the other `Type` can be `Audio`, right? And then why don't we need to process it?

handlers/mist_triggers.go
```diff
@@ -0,0 +1,207 @@
...
+// This trigger is stream-specific and non-blocking. The payload for this trigger is multiple lines,
+// each separated by a single newline character (without an ending newline), containing data:
+//   push ID (integer)
+//   stream name (string)
+//   target URI, before variables/triggers affected it (string)
+//   target URI, afterwards, as actually used (string)
+//   last 10 log messages (JSON array string)
+//   most recent push status (JSON object string)
+func (d *MistCallbackHandlersCollection) TriggerPushEnd(t *Trigger) {
+	payload, err := io.ReadAll(t.req.Body)
```
leszko, line 108: This part looks the same as in other trigger handlers. Is it maybe possible to extract and reuse?

handlers/mist_triggers.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+}
+
+func (d *MistCallbackHandlersCollection) TranscodingPushEnd(t *Trigger, destination, actualDestination, pushStatus string) {
+	info, err := d.StreamCache.Transcoding.Get(t.streamName)
+	if err != nil {
+		t.Err("unknown push source", err)
+		return
+	}
+	inOurBooks := false
```
leszko, line 136: Nit: Maybe `inProgress`? Or `isBeingProcessed`?

handlers/mist_triggers.go
```diff
@@ -0,0 +1,207 @@
...
+			t.Err("Cannot send segment transcode status", err)
+		}
+		d.StreamCache.Transcoding.Remove(t.streamName)
+	}
+	return
+}
+
+func (d *MistCallbackHandlersCollection) SegmentingPushEnd(t *Trigger) {
+	// when uploading is done, remove trigger and stream from Mist
+	errT := d.MistClient.DeleteTrigger(t.streamName, "PUSH_END")
```
leszko, line 170: I think it'd safer to first notify callbackClient and then delete triggers / streams for Mist. If for some reason the deletion fails, then I think the callback client should still get informed about the process.

handlers/mist_triggers.go
AlexKordic: Here we have few pipeline stages:
- Pull from S3 HTTP into SOURCE stream
- Transcode from SOURCE to RENDITION
- Push from RENDITION stream to S3 rendition 1
- Push from RENDITION stream to S3 rendition 2
- Push from RENDITION stream to S3 rendition 3

When previous stage ends next one is usually still in progress. Here `Transcode from SOURCE to RENDITION` stage ended, hence the note next stage is still in progress and we can cleanup stream SOURCE involved in stages 1,2

handlers/mist_triggers.go
```diff
@@ -0,0 +1,207 @@
...
+	if err := callbackClient.SendTranscodeStatus(callbackUrl, clients.TranscodeStatusTranscoding, 0.0); err != nil {
+		t.Err("Cannot send transcode status", err)
+	}
+	d.StreamCache.Segmenting.Remove(t.streamName)
+
+	// TODO: add timeout for the stream upload
+	// TODO: start transcoding
+}
+
+type Trigger struct {
```
leszko, line 195: As mentioned earlier, I'd suggest to delete this type.

handlers/mist_triggers.go
AlexKordic: `i` is generated name - not id of any sort, either given by us to Mist or used by Mist internally. JSON format is such that it could be array but it is map. So we iterate like its array.

handlers/mist_triggers.go
AlexKordic: Intent is opposite: no error (push started) then `AddDestination()`

handlers/mist_triggers.go
AlexKordic: The process looks as follows:
1. Add SOURCE stream to Mist using MistClient
2. Execute the `MistProceLivepeer` process to perform the following operations:
   - Take SOURCE stream and segment it
   - Push each segment to Broadcaster
   - Receive transcoded segment from Broadcaster, ingest into RENDITION stream
3. Start SOURCE stream by creating, unused, push to /dev/null.ts
4. `MistProceLivepeer` process is unblocked, starts to work
5. Return 200 OK response
5. Respond to LIVE_TRACK_LIST trigger:
   - Extract track information
   - **> For each video track create push to S3 destination**
   - Transcoding is complete if there is no tracks. Then delete SOURCE stream
6. Respond to PUSH_END trigger:
   - Determine is it error or successful upload
   - Invoke callback to studio
7. When all started pushes are complete invoke callback to studio; remove stream from StreamCache

We are at the step: For each video track create push to S3 destination

Note that we want to create self-contained rendition segments, with both one audio and one video track. Our RENDITION stream contains all video tracks and duplicate audio tracks.

Finally for each video track we start new push to create separate rendition. If we had `x` transcoded video tracks from `broadcaster` we want `x` new pushes to produce `x` `mpegts` segments on S3 storage.

handlers/mist_triggers.go
AlexKordic: Makes sense.

Note i wasn't changing segmenting logic in this PR, just mechanisms around it for trigger callbacks, error reporting and similar..

handlers/mist_triggers.go
```diff
@@ -0,0 +1,196 @@
...
+	actualDestination := lines[3]
+	pushStatus := lines[5]
+	if yes, _ := isTranscodeStream(streamName); yes {
+		d.TranscodingPushEnd(w, req, streamName, destination, actualDestination, pushStatus)
+	} else {
+		d.SegmentingPushEnd(w, req, streamName)
+	}
+}
+
+func (d *MistCallbackHandlersCollection) TranscodingPushEnd(w http.ResponseWriter, req *http.Request, streamName, destination, actualDestination, pushStatus string) {
```
leszko, line 124: I see `req` param is not used. How about removing it?

handlers/mist_triggers.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+	isBeingProcessed := false
+	for _, existing := range info.Destinations {
+		if existing == destination {
+			isBeingProcessed = true
+			break
+		}
+	}
+	if !isBeingProcessed {
+		Err(streamName, "missing from our books", nil, w)
```
leszko, line 138: Maybe `"stream not found in cache"`? Not sure we use `"books"` in our domain language.

handlers/mist_triggers.go
```diff
@@ -0,0 +1,196 @@
...
+			Err(streamName, "Cannot send segment transcode status", err, w)
+		}
+		d.StreamCache.Transcoding.Remove(streamName)
+	}
+	return
+}
+
+func (d *MistCallbackHandlersCollection) SegmentingPushEnd(w http.ResponseWriter, req *http.Request, streamName string) {
+	// when uploading is done, remove trigger and stream from Mist
+	defer d.StreamCache.Segmenting.Remove(streamName)
```
leszko, line 164: Maybe without `defer`? Do we need anything from the cache in this function?

handlers/mist_triggers.go
```diff
@@ -0,0 +1,196 @@
...
+	if errS != nil {
+		Err(streamName, fmt.Sprintf("Cannot remove stream '%s'", streamName), errS, w)
+		return
+	}
+	if errC != nil {
+		Err(streamName, "Cannot send transcode status", errC, w)
+	}
+
+	// TODO: add timeout for the stream upload
+	// TODO: start transcoding
```
leszko, line 188:
```suggestion
```

I believe we don't need the `start transcoding` TODO anymore

clients/callback_client.go
```diff
@@ -35,44 +36,74 @@ func (c CallbackClient) DoWithRetries(r *http.Request) error {
 
 	resp, err := c.httpClient.Do(r)
 	if err != nil {
-		return fmt.Errorf("failed to send callback to %q. Error: %s", r.URL.String(), err)
+		err = fmt.Errorf("failed to send callback to %q. Error: %s", r.URL.String(), err)
```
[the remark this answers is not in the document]
AlexKordic: Looks like mistake, fixing next condition also

emranemran:
Where is `TestSegmentTranscode`? 
```
root@ubuntu-linux-20-04-desktop:/media/psf/AllFiles/Users/emran/catalyst-api/handlers# go test . -v -count 1 -tags externaldeps -run TestSegmentTranscode
testing: warning: no tests to run
PASS
ok      github.com/livepeer/catalyst-api/handlers       0.004s [no tests to run]
```

emranemran:
Also tried testing the API server per your instructions above and got this: 
``` 
2022/09/06 21:52:38 body schema validation failed EOF
2022/09/06 21:52:38 TranscodeSegment request validation failed EOF
ts=2022-09-07T04:52:38.985781216Z remote=127.0.0.1:59920 proto=HTTP/1.1 method=POST uri=/api/transcode/file duration=123.959µs status=500

```

emranemran:
Can the commits be grouped and explained in detail so that it's easier to review all the changes?

AlexKordic:
> Where is `TestSegmentTranscode`?

@emranemran removed in process of this code review

clients/callback_client_test.go
```diff
@@ -90,4 +98,19 @@ func TestTranscodeStatusErrorNotifcation(t *testing.T) {
 	// Send the callback and confirm the number of times we retried
 	client := NewCallbackClient()
 	require.NoError(t, client.SendTranscodeStatusError(svr.URL, "something went wrong"))
+	waitForCallbacks(t, callbacks, 1, 5*time.Second)
```
thomshutt, line 101: Why do we need to wait?

PR #33, opened 2022-09-02
Send overall (rather than per-stage) completion ratio to Studio

thomshutt:


clients/callback_client.go
```diff
@@ -86,21 +90,54 @@ func (c CallbackClient) SendTranscodeStatusError(callbackURL, errorMsg string) e
...
+	if currentStageCompletionRatio < 0 {
+		currentStageCompletionRatio = 0
+	}
+	if currentStageCompletionRatio > 1 {
+		currentStageCompletionRatio = 1
+	}
+
+	// Define the "base" numbers - e.g the overall ratio we start each stage at
+	var TranscodeStatusPreparingBase float64 = 0.0
+	var TranscodeStatusPreparingTranscodingBase float64 = 0.4
```
victorges, line 107: Should it be only "Transcoding"?
```suggestion
	var TranscodeStatusTranscodingBase float64 = 0.4
```

clients/callback_client.go
```diff
@@ -86,21 +90,54 @@ func (c CallbackClient) SendTranscodeStatusError(callbackURL, errorMsg string) e
...
+	var TranscodeStatusCompletedBase float64 = 1
+
+	switch status {
+	case TranscodeStatusPreparing:
+		return TranscodeStatusPreparingBase + (currentStageCompletionRatio * (TranscodeStatusPreparingTranscodingBase - TranscodeStatusPreparingBase))
+	case TranscodeStatusPreparingCompleted:
+		return TranscodeStatusPreparingTranscodingBase
+	case TranscodeStatusTranscoding:
+		return TranscodeStatusPreparingTranscodingBase + (currentStageCompletionRatio * (TranscodeStatusCompletedBase - TranscodeStatusPreparingTranscodingBase))
```
victorges, line 116: Another way of implementing this, to have the more complex math only in 1 place, would be to have 2 vars "start" and "end" and force the `currentStageCompletionRatio` to be `1` on the "Completed" steps (but use their same start/end as the non-completed). Sth like
```go
	switch status {
	case TranscodeStatusPreparing, TranscodeStatusPreparingCompleted:
		start, end = 0, 0.4
	case TranscodeStatusTranscoding, TranscodeStatusTranscodingCompleted:
		start, end = 0.4, 1
        }
        if (status == TranscodeStatusPreparingCompleted || status == TranscodeStatusTranscodingCompleted) {
                currentStageCompletionRatio = 1
        }
```


Then there could be a single progress-scaling function like this: https://github.com/livepeer/task-runner/blob/4d1ae1768c01e979969bf7f00b81267db05e299c/task/progress.go#L66

Certainly not a required change, just tripping on the impl. Could make sense when we start adding more steps here. WDYT?

victorges (approved):
Cool!

clients/callback_client.go
thomshutt: The "Base" was trying to signify that this is the lower limit / floor / whatever, open to suggestions for better names though. Maybe "Floor"?

thomshutt: Just realised what you were actually pointing out 👓

clients/callback_client.go
thomshutt: I like that a lot!

clients/callback_client.go
```diff
@@ -86,21 +90,54 @@ func (c CallbackClient) SendTranscodeStatusError(callbackURL, errorMsg string) e
...
+	if status == TranscodeStatusPreparingCompleted || status == TranscodeStatusCompleted {
+		currentStageCompletionRatio = 1
+	}
+
+	switch status {
+	case TranscodeStatusPreparing, TranscodeStatusPreparingCompleted:
+		return scaleProgress(currentStageCompletionRatio, 0, 0.4)
+	case TranscodeStatusTranscoding, TranscodeStatusCompleted:
+		return scaleProgress(currentStageCompletionRatio, 0.4, 1)
+	}
```
victorges, line 115: Nice!

PR #35, opened 2022-09-06
Stub "Transcode Complete" callback

thomshutt:
This is so we can do any final iterations on how the output will look and slightly unblock Studio

### DIFF
--- a/clients/callback_client.go
+++ b/clients/callback_client.go
@@ -182,7 +182,7 @@ func (ts TranscodeStatus) String() string {
 type TranscodeStatusMessage struct {
 	CompletionRatio float64 `json:"completion_ratio"` // No omitempty or we lose this for 0% completion case
 	Error           string  `json:"error,omitempty"`
-	Retriable       *bool   `json:"retriable,omitempty"` // Has to be a pointer or we can't differentiate omission from 'false'
+	Unretriable     bool    `json:"unretriable,omitempty"`
 	Status          string  `json:"status,omitempty"`
 	Timestamp       int64   `json:"timestamp"`
 }
--- a/clients/callback_client.go
+++ b/clients/callback_client.go
@@ -202,9 +202,10 @@ type InputVideoTrack struct {
 }
 
 type InputVideo struct {
-	Format   string            `json:"format"`
-	Tracks   []InputVideoTrack `json:"tracks"`
-	Duration float64           `json:"duration"`
+	Format    string            `json:"format"`
+	Tracks    []InputVideoTrack `json:"tracks"`
+	Duration  float64           `json:"duration"`
+	SizeBytes float64           `json:"size"`
 }
 
 type OutputVideoFile struct {
--- a/clients/callback_client.go
+++ b/clients/callback_client.go
@@ -205,7 +205,7 @@ type InputVideo struct {
 	Format    string            `json:"format"`
 	Tracks    []InputVideoTrack `json:"tracks"`
 	Duration  float64           `json:"duration"`
-	SizeBytes float64           `json:"size"`
+	SizeBytes int               `json:"size"`
 }
 
 type OutputVideoFile struct {
--- a/clients/callback_client.go
+++ b/clients/callback_client.go
@@ -200,11 +200,12 @@ type AudioTrack struct {
 }
 
 type InputTrack struct {
-	Type        string  `json:"type"`
-	Codec       string  `json:"codec"`
-	Bitrate     int     `json:"bitrate"`
-	DurationSec float64 `json:"duration"`
-	SizeBytes   int     `json:"size"`
+	Type         string  `json:"type"`
+	Codec        string  `json:"codec"`
+	Bitrate      int     `json:"bitrate"`
+	DurationSec  float64 `json:"duration"`
+	SizeBytes    int     `json:"size"`
+	StartTimeSec float64 `json:"start_time"`
 
 	// Fields only used if this is a Video Track
 	VideoTrack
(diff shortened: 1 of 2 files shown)


clients/callback_client.go
```diff
@@ -157,3 +176,52 @@ func (ts TranscodeStatus) String() string {
...
+	Type        string  `json:"type,omitempty"`
+	Codec       string  `json:"codec,omitempty"`
+	Width       int     `json:"width,omitempty"`
+	Height      int     `json:"height,omitempty"`
+	Bitrate     int     `json:"bitrate,omitempty"`
+	Duration    float64 `json:"duration,omitempty"`
+	PixelFormat string  `json:"pixel_format,omitempty"`
+	Channels    int     `json:"channels,omitempty"`
+	SampleRate  int     `json:"sampleRate,omitempty"`
+	SizeBytes   int     `json:"size,omitempty"`
```
victorges, line 201: Might be worth reordering/categorizing the fields here just so it's easier to understand whats higher-level, what's specific for given codecs, etc. e.g.
https://github.com/livepeer/go-api-client/blob/76a0884161c19aea4c15f99897540b3880df5aad/api.go#L317-L332

Also, I like to remove the `omitempty` on fields that are not optional, like `type` (and maybe `codec`). Not necessarily following any guideline tho I guess.

clients/callback_client.go
```diff
@@ -157,3 +176,52 @@ func (ts TranscodeStatus) String() string {
...
+	Fps         int     `json:"fps,omitempty"`
+	Type        string  `json:"type,omitempty"`
+	Codec       string  `json:"codec,omitempty"`
+	Width       int     `json:"width,omitempty"`
+	Height      int     `json:"height,omitempty"`
+	Bitrate     int     `json:"bitrate,omitempty"`
+	Duration    float64 `json:"duration,omitempty"`
+	PixelFormat string  `json:"pixel_format,omitempty"`
+	Channels    int     `json:"channels,omitempty"`
+	SampleRate  int     `json:"sampleRate,omitempty"`
```
victorges, line 200: Should use consistent casing here. We use `camelCase` in the Studio API, but I think the rest of Catalyst's API is in `snake_case` right? So `sample_rate` should probably make more sense here.

clients/callback_client.go
```diff
@@ -157,3 +176,52 @@ func (ts TranscodeStatus) String() string {
...
+
+type TranscodeStatusMessage struct {
+	CompletionRatio float64 `json:"completion_ratio"` // No omitempty or we lose this for 0% completion case
+	Error           string  `json:"error,omitempty"`
+	Retriable       *bool   `json:"retriable,omitempty"` // Has to be a pointer or we can't differentiate omission from 'false'
+	Status          string  `json:"status,omitempty"`
+	Timestamp       int64   `json:"timestamp"`
+}
+
+type InputVideoTrack struct {
```
victorges, line 190: Missing `start_time` here, isn't it relevant? Maybe not if we handle it correctly I guess (which we don't do in current pipeline 😆) but raising it here for consideration.

clients/callback_client.go
```diff
@@ -157,3 +176,52 @@ func (ts TranscodeStatus) String() string {
...
+	Height      int     `json:"height,omitempty"`
+	Bitrate     int     `json:"bitrate,omitempty"`
+	Duration    float64 `json:"duration,omitempty"`
+	PixelFormat string  `json:"pixel_format,omitempty"`
+	Channels    int     `json:"channels,omitempty"`
+	SampleRate  int     `json:"sampleRate,omitempty"`
+	SizeBytes   int     `json:"size,omitempty"`
+}
+
+type InputVideo struct {
```
victorges, line 204: Could we get a total file size as well? (Does it make sense to have it in the tracks? I think only in the whole file is sufficient)

clients/callback_client.go
```diff
@@ -157,3 +176,52 @@ func (ts TranscodeStatus) String() string {
 	}
 	return "unknown"
 }
+
+// The various status messages we can send
+
+type TranscodeStatusMessage struct {
+	CompletionRatio float64 `json:"completion_ratio"` // No omitempty or we lose this for 0% completion case
+	Error           string  `json:"error,omitempty"`
+	Retriable       *bool   `json:"retriable,omitempty"` // Has to be a pointer or we can't differentiate omission from 'false'
```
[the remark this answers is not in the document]
thomshutt: 👍 done

clients/callback_client.go
thomshutt: Sure, what unit would it be?

clients/callback_client.go
thomshutt: Added! I think just on the input video as a whole makes sense

clients/callback_client.go
thomshutt: Split it out into seperate sub-structs now, let me know what you think.

Removed the omitemptys - we can add back in as needed in the future

victorges: Ah neat! I liked the separate structs.

Could add the `omitempty` at least on the audio/video track fields, since those are not present for the other track type.

clients/callback_client.go
victorges: I think it's in floating seconds like the duration

clients/callback_client.go
thomshutt: Good point, done

clients/callback_client.go
thomshutt: Added

thomshutt:
@victorges I'm going to merge this now all the comments are addressed, just to avoid conflicts with other in-flight PRs but can change anything else you want in a followup PR

PR #36, opened 2022-09-07
Split out HTTP + Handlers

thomshutt:
We're going to deprecate the giant PR in https://github.com/livepeer/catalyst-api/pull/30 and try to split it into a series of more manageable / reviewable ones.

This one creates a more manageable base in terms of package structure by only doing CLI stuff in the `cmd` package and moving HTTP layer functions into `api`. It also splits out `handlers/handlers.go` into different files based on functionality - e.g `handlers/upload.go`. It also moves the Mist client into the `clients` package.


emranemran (approved):

emranemran:
LGTM - didn't find any issues with the logic but we can address after everything is merged and tested.

PR #41, opened 2022-09-09
Call Mist to transcode input manifest and push renditions to S3

thomshutt:
```
curl --location --request POST 'http://localhost:4949/api/transcode/file' \ 
--header 'Authorization: Bearer IAmAuthorized' \ 
--data-raw '{
  "source_location": "s3+https://REDACTED:REDACTED@storage.googleapis.com/alexk-dms-upload-test/avsample.mp4",
  "callback_url": "https://webhook.site/YOUR-UNIQUE-URL-HERE",
  "manifestID": "somestream123",
  "profiles": [
        {
        "name": "360p",
          "width": 640,
          "height": 360,
          "bitrate": 200000,
          "fps": 24
        }
    ], 
  "verificationFreq": 1}'
```

This PR still needs some tidying, async transcode (i.e HTTP call shouldn't wait for that) added back in and lots of unit tests - ideally mostly at the function rather than HTTP level since we don't have a good story around mocking MistProcLivepeer yet.


config/config.go
```diff
@@ -14,14 +14,12 @@ var Clock TimestampGenerator = RealTimestampGenerator{}
 // Path to Mist's "Livepeer" process that we shell out to for the transcoding
 const PathMistProcLivepeer = "./MistProcLivepeer"
 
+// Port that the local Broadcaster runs on
+const BroadcasterPort = 8935
```
thomshutt, line 18: Hardcoding until we need to change it

handlers/handlers_test.go
```diff
@@ -29,70 +25,6 @@ func TestOKHandler(t *testing.T) {
 	require.Equal(rr.Body.String(), "OK")
 }
 
-func TestSegmentCallback(t *testing.T) {
```
thomshutt, line 32: This test needs to be rewritten / split up (along with the underlying code) to test something specific rather than just and E2E call into the HTTP endpoint

config/config.go
emranemran: I have a change to read this from the transcode http request and parse into json to pass to MistProcLivepeer - left it out in my latest commit for now to reduce areas of debug. Will push it once we've worked through the critical bugs.

handlers/misttriggers/live_tracklist.go
```diff
@@ -70,13 +73,33 @@ func (d *MistCallbackHandlersCollection) TriggerLiveTrackList(w http.ResponseWri
...
 
-	// Start push per each video track
+	// Upload each track (transcoded rendition) returned by Mist to S3
 	for i := range tracks {
+		// Only produce a rendition for each video track, selecting best audio track
 		if tracks[i].Type != "video" {
-			// Only produce an rendition per each video track, selecting best audio track
 			continue
 		}
-		destination := fmt.Sprintf("%s/%s__%dx%d.ts?video=%d&audio=maxbps", info.UploadDir, streamName, tracks[i].Width, tracks[i].Height, tracks[i].Index) //.Id)
```
yondonfu: Was the bug that the destination URL we were constructing here was using a .ts file extension, but we want a .m3u8 file extension?

emranemran: Yeah

handlers/misttriggers/push_end.go
```diff
@@ -30,22 +30,53 @@ func (d *MistCallbackHandlersCollection) TriggerPushEnd(w http.ResponseWriter, r
...
+	uploadSuccess := pushStatus == "null"
+	if uploadSuccess {
+		// TODO: Do some maths so that we don't always send 0.5
+		if err := clients.DefaultCallbackClient.SendTranscodeStatus(info.CallbackUrl, clients.TranscodeStatusTranscoding, 0.5); err != nil {
+			_ = config.Logger.Log("msg", "Error sending transcode status in TranscodingPushEnd", "err", err)
+		}
+	} else {
+		// We forward pushStatus json to callback
+		if err := clients.DefaultCallbackClient.SendTranscodeStatusError(info.CallbackUrl, fmt.Sprintf("Error while pushing to %s: %s", actualDestination, pushStatus)); err != nil {
+			_ = config.Logger.Log("msg", "Error sending transcode error status in TranscodingPushEnd", "err", err)
```
yondonfu, line 67: Do we need the `_` if we're ignoring the return value?

handlers/misttriggers/push_end.go
```diff
@@ -30,22 +30,53 @@ func (d *MistCallbackHandlersCollection) TriggerPushEnd(w http.ResponseWriter, r
...
+		}
+	}
+
+	// We do not delete triggers as source stream is wildcard stream: RENDITION_PREFIX
+	cache.DefaultStreamCache.Transcoding.RemovePushDestination(streamName, destination)
```
yondonfu, line 72: Just to clarify - the desired behavior is to clean up the transcoding destinations and remove the stream regardless of whether the push was successful or not (pushstatus == null)?

handlers/transcode.go
```diff
@@ -72,9 +68,195 @@ func (d *CatalystAPIHandlersCollection) TranscodeSegment() httprouter.Handle {
...
+		}
+		if err == io.EOF {
+			return fmt.Errorf("Improper termination: %v", line)
 		}
-		_, _ = io.WriteString(w, "OK") // TODO later
+		if err != nil {
+			return err
+		}
+
+		mw.Write(line)
```
yondonfu, line 94: Let's handle the return value here.

handlers/transcode.go
```diff
@@ -72,9 +68,195 @@ func (d *CatalystAPIHandlersCollection) TranscodeSegment() httprouter.Handle {
...
+	configPayload, err := json.Marshal(configForSubprocess(request, inputStream, renditionsStream))
+	if err != nil {
+		return fmt.Errorf("ProcLivepeerConfig json encode: %s", err)
+	}
+	args := string(configPayload)
+
+	transcodeCommand := exec.Command(config.PathMistProcLivepeer, args, "--debug", "8")
+
+	var stdout, stderr bytes.Buffer
+	stderrPipe, err := transcodeCommand.StderrPipe()
```
yondonfu, line 116: Let's handle the error.

handlers/transcode.go
```diff
@@ -72,9 +68,195 @@ func (d *CatalystAPIHandlersCollection) TranscodeSegment() httprouter.Handle {
...
+	stdoutPipe, err := transcodeCommand.StdoutPipe()
```
yondonfu, line 117: Let's handle the error.

handlers/transcode.go
```diff
@@ -72,9 +68,195 @@ func (d *CatalystAPIHandlersCollection) TranscodeSegment() httprouter.Handle {
...
+
+	// Start the Transcode Command asynchronously - we call Wait() later in this method
+	fmt.Printf("Starting transcode via: %s\n", transcodeCommand.String())
+	err = transcodeCommand.Start()
+	if err != nil {
+		fmt.Printf("start transcodeCommand: %s\n", err)
+	}
+
+	go func() {
+		streamOutput(stdoutPipe, &stdout, os.Stdout)
```
yondonfu, line 127: Let's handle the error.

handlers/transcode.go
```diff
@@ -72,9 +68,195 @@ func (d *CatalystAPIHandlersCollection) TranscodeSegment() httprouter.Handle {
...
+	}()
+	go func() {
+		streamOutput(stderrPipe, &stderr, os.Stderr)
```
yondonfu, line 130: Let's handle the error.

handlers/transcode.go
```diff
@@ -72,9 +68,195 @@ func (d *CatalystAPIHandlersCollection) TranscodeSegment() httprouter.Handle {
...
+
+	var stdout, stderr bytes.Buffer
+	stderrPipe, err := transcodeCommand.StderrPipe()
+	stdoutPipe, err := transcodeCommand.StdoutPipe()
+
+	// Start the Transcode Command asynchronously - we call Wait() later in this method
+	fmt.Printf("Starting transcode via: %s\n", transcodeCommand.String())
+	err = transcodeCommand.Start()
+	if err != nil {
+		fmt.Printf("start transcodeCommand: %s\n", err)
```
yondonfu, line 123: If `Start()` returns an error should exit early?

handlers/transcode.go
```diff
@@ -72,9 +68,195 @@ func (d *CatalystAPIHandlersCollection) TranscodeSegment() httprouter.Handle {
...
+			Height:     &Height,
+			Fps:        &Fps,
+			GOP:        profile.GOP,
+			AvcProfile: profile.Profile,
+		})
+	}
+	return conf
+}
+
+func commandOutputToLog(cmd *exec.Cmd, name string) {
```
yondonfu, line 216: Can be removed?

handlers/transcode.go
```diff
@@ -72,9 +68,195 @@ func (d *CatalystAPIHandlersCollection) TranscodeSegment() httprouter.Handle {
...
+	go pipeToLog(stdoutPipe, name)
+	stderrPipe, err := cmd.StderrPipe()
+	if err != nil {
+		log.Printf("ERROR: cmd.StderrPipe() %v", err)
+		return
+	}
+	go pipeToLog(stderrPipe, name)
+}
+
+func pipeToLog(pipe io.ReadCloser, name string) {
```
yondonfu, line 231: Can be removed?

handlers/handlers_test.go
yondonfu, line 32: Just to clarify - we're removing this test instead of updating it because now that transcode is not stubbed out, we don't have a good way to stub the MistProcLivepeer exec command so making this test pass would be tricky?

yondonfu (changes requested):

handlers/misttriggers/push_end.go
emranemran: Yes, for now. We will revisit error recovery as part of this here and address it in a follow up PR: 
https://github.com/livepeer/catalyst/issues/171

config/config.go
```diff
@@ -12,16 +12,16 @@ var Version string
 var Clock TimestampGenerator = RealTimestampGenerator{}
 
 // Path to Mist's "Livepeer" process that we shell out to for the transcoding
-const PathMistProcLivepeer = "./MistProcLivepeer"
+const PathMistProcLivepeer = "/media/psf/AllFiles/Users/emran/catalyst/bin/MistProcLivepeer"
```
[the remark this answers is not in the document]
emranemran: Yeah updated to /usr/bin for now but will follow up to try and fail smarter if it's not found in $PATH
https://github.com/livepeer/catalyst/issues/172

handlers/transcode.go
```diff
@@ -72,9 +68,179 @@ func (d *CatalystAPIHandlersCollection) TranscodeSegment() httprouter.Handle {
...
+
+	configPayload, err := json.Marshal(configForSubprocess(request, inputStream, renditionsStream))
+	if err != nil {
+		return fmt.Errorf("ProcLivepeerConfig json encode: %s", err)
+	}
+	args := string(configPayload)
+
+	transcodeCommand := exec.Command(config.PathMistProcLivepeer, args, "--debug", "8", "--kickoff")
+
+	var stdout, stderr bytes.Buffer
```
AlexKordic, line 116: If buffers are unused, can we remove them? Maybe change `streamOutput()` to reflect this.

handlers/transcode.go
```diff
@@ -72,9 +68,179 @@ func (d *CatalystAPIHandlersCollection) TranscodeSegment() httprouter.Handle {
...
+	// TODO: remove when Mist code is updated https://github.com/DDVTECH/mistserver/issues/81
+	// Starting SOURCE_PREFIX stream because MistProcLivepeer is unable to start it automatically
+	file, err := os.CreateTemp("", "tmp-*.ts")
+	if err != nil {
+		return fmt.Errorf("error creating temporary file for SOURCE_PREFIX stream: %s", err)
+	}
+
+	if err := mistClient.PushStart(inputStream, file.Name()); err != nil {
+		return fmt.Errorf("PushStart(inputStream): %s", err)
+	}
```
AlexKordic, line 153: With latest mistserver updates, is this step still required?

handlers/transcode.go
```diff
@@ -72,9 +68,179 @@ func (d *CatalystAPIHandlersCollection) TranscodeSegment() httprouter.Handle {
...
+
+	inputUrl, err := url.Parse(request.SourceFile)
+	if err != nil {
+		return fmt.Errorf("invalid request source_location: %s", err)
+	}
```
AlexKordic, line 158: This check should take place before subprocess is started

handlers/transcode.go
```diff
@@ -72,9 +68,179 @@ func (d *CatalystAPIHandlersCollection) TranscodeSegment() httprouter.Handle {
...
+		Profiles:    request.Profiles[:],
+		UploadDir:   uploadDir.String(),
+	})
+
+	if err := transcodeCommand.Wait(); err != nil {
+		if exit, ok := err.(*exec.ExitError); ok {
+			log.Printf("MistProcLivepeer returned %d", exit.ExitCode())
+		}
+		return fmt.Errorf("exec transcodeCommand: %s", err)
+	}
```
AlexKordic, line 175: Should be started in new goroutine so HTTP API call can return with success. Studio should rely on callbacks for results on pipeline completion & errors.

handlers/transcode.go
```diff
@@ -72,9 +68,179 @@ func (d *CatalystAPIHandlersCollection) TranscodeSegment() httprouter.Handle {
...
+
+	return nil
+}
+
+// configForSubprocess transforms request information to MistProcLivepeer config json
+// We use .HardcodedBroadcasters assuming we have local B-node.
```
AlexKordic, line 181: Comment should be updated about added option to use remote B-node.

AlexKordic (approved):
Well done. Some of my comments may be out of scope of this PR

handlers/misttriggers/live_tracklist.go
```diff
@@ -22,10 +33,57 @@ type MistTrack struct {
...
+	}
+	return multivariantPlaylist
+
+}
+
+func uploadPlaylist(uploadPath, manifest string) {
+
+	storageDriver, err := drivers.ParseOSURL(uploadPath, true)
+	if err != nil {
+		log.Printf("error drivers.ParseOSURL %v %s", err, uploadPath)
```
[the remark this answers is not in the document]
emranemran: Good catch, will fix.

handlers/transcode.go
emranemran: Hmm, unclear on what you mean - can you clarify? The buffers are used to store stdout/stderr from `io.MultiWriter`

handlers/transcode.go
emranemran: Good point, will test it out after removing it - technically shouldn't be required anymore.

handlers/transcode.go
emranemran: Good catch, fixed.

handlers/misttriggers/live_tracklist.go
```diff
@@ -0,0 +1,184 @@
...
+	"github.com/livepeer/catalyst-api/cache"
+	"github.com/livepeer/catalyst-api/config"
+	"github.com/livepeer/catalyst-api/errors"
+)
+
+type MistTrack struct {
+	// populated by mist when trigger is received
+	Id          int32  `json:"trackid"`
+	ByteRate    int32  `json:"bps"`
+	Kfps        int32  `json:"fpks"`
```
yondonfu, line 28: Is this a typo? Is the actual field name in the trigger "fpks" and not "kfps"?

handlers/misttriggers/live_tracklist.go
```diff
@@ -0,0 +1,109 @@
...
+	// Check if LIVE_TRACK_LIST trigger is being fired *after* the push-from-Mist-to-S3 is complete
+	var streamEnded = (encodedTracks == "null")
+	if streamEnded {
+		// SOURCE_PREFIX stream is no longer needed
+		suffix := strings.TrimPrefix(streamName, config.RENDITION_PREFIX)
+		inputStream := fmt.Sprintf("%s%s", config.SOURCE_PREFIX, suffix)
+		if err := d.MistClient.DeleteStream(inputStream); err != nil {
+			log.Printf("ERROR LIVE_TRACK_LIST DeleteStream(%s) %v", inputStream, err)
+		}
+		// Multiple pushes from RENDITION_PREFIX are in progress.
```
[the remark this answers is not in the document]
yondonfu: Following up here just for clarification.

handlers/transcode.go
emranemran: Fixed.

yondonfu (commented):
Needs a `go mod tidy`.

handlers/transcode.go
emranemran: Yes, agree - I'm tempted to leave it like this for us to do some initial tests in canary environment first and then turn this into a goroutine -- mainly because it's a lot easier to debug and root cause this way.

handlers/misttriggers/live_tracklist.go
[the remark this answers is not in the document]
emranemran: I don't think this comment applies anymore in this current flow. @AlexKordic - can you confirm? Ignoring the comment, the current implementation will return after deleting the streamName which is correct. I can remove this comment in a follow-up PR.

handlers/handlers_test.go
emranemran: Yes

handlers/transcode.go
emranemran: Removed

handlers/transcode.go
AlexKordic: Noticing we don't need to duplicate stream from `stdoutPipe`&`stderrPipe` into buffers. We just want it printed into log? Is this true? Maybe remnant of debugging?

handlers/misttriggers/live_tracklist.go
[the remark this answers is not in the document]
AlexKordic: All pushes are still ongoing when push-source `tr_rend_+...` stream ends. Here we can delete s3-source stream. The comment is to remind us this is not last trigger in the pipeline and we are still waiting for PUSH_END triggers.

PR #48, opened 2022-09-21
Define pipeline constants

AlexKordic:
The number of required functions are growing. Preparing for recording pipeline changes, lets define new constants for each pipeline. In trigger handling we can use stream prefix to determine to which function a stream belongs to.

Changes:
- Define pipeline constants
- Triggers: Use switch to branch to proper pipeline logic based on stream name
- Move string constant from UploadVOD() to config

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -23,5 +23,7 @@ func init() {
 
 // Prefixes used in Mist stream names to let us determine whether a given "stream" in Mist is being used
 // for the segmenting or transcoding phase
+const SEGMENTING_PREFIX = "catalyst_vod_"
 const SOURCE_PREFIX = "tr_src_"
 const RENDITION_PREFIX = "tr_rend_+"
+const RECORDING_PREFIX = "video"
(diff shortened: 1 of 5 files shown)


yondonfu (commented):
LGTM, but since I am not as familiar with this codebase, I'll leave the approval for @emranemran.

config/config.go
```diff
@@ -23,5 +23,7 @@ func init() {
 
 // Prefixes used in Mist stream names to let us determine whether a given "stream" in Mist is being used
 // for the segmenting or transcoding phase
+const SEGMENTING_PREFIX = "catalyst_vod_"
```
emranemran, line 26: Should we just call this "vod"? Technically everything is catalyst now.

config/config.go
```diff
@@ -23,5 +23,7 @@ func init() {
...
 const SOURCE_PREFIX = "tr_src_"
 const RENDITION_PREFIX = "tr_rend_+"
+const RECORDING_PREFIX = "video"
```
emranemran, line 29: Should we call this "live"? Would it be clearer to indicate it's a live stream?

emranemran (approved):
LGTM - two minor comments re: naming.

config/config.go
AlexKordic: Good point. Maybe we should save `vod+` prefix for playback.

Created issue https://github.com/livepeer/catalyst-api/issues/51

config/config.go
AlexKordic: Yes, makes sense. We need to change this in live workflow and then update in catalyst API.

PR #49, opened 2022-09-26
Recording start callback

AlexKordic:


clients/callback_client.go
```diff
@@ -48,6 +49,22 @@ func (c CallbackClient) DoWithRetries(r *http.Request) error {
 	return nil
 }
 
+func (c CallbackClient) SendRecordingEvent(event *RecordingEvent) {
+	eventJson, err := json.Marshal(event)
+	if err != nil {
+		log.Printf("SendRecordingStarted json marshal %v", err)
+		return
+	}
+	req, err := http.NewRequest(http.MethodPost, config.RecordingCallback, bytes.NewReader(eventJson))
```
victorges, line 58: If this is run in background and errors ignored, WDYT of at least adding a couple of retries here so we don't miss recordings out of ephemeral errors?

clients/callback_client.go
```diff
@@ -181,6 +198,15 @@ func (ts TranscodeStatus) String() string {
 
 // The various status messages we can send
 
+type RecordingEvent struct {
+	When        string `json:"when"`
```
victorges, line 202: Likely just a personal preference, but WDYT of `event` or `type` for this field instead?

clients/callback_client.go
```diff
@@ -181,6 +198,15 @@ func (ts TranscodeStatus) String() string {
...
+	StreamId    string `json:"streamId"`
```
victorges, line 203: Is this the Mist stream name (like `video+123`) or the Livepeer Studio Stream ID. I actually doubt it is the Studio ID since catalyst doesn't know Studio, so in that case WDYT of naming this `stream_name` instead?

clients/callback_client.go
```diff
@@ -181,6 +198,15 @@ func (ts TranscodeStatus) String() string {
...
+	RecordingId string `json:"recordingId"`
+	Hostname    string `json:"hostName"`
```
victorges, line 205: For consistency with the other Catalyst APIs, I believe this should use `snake_case`
```suggestion
	StreamId    string `json:"stream_id"`
	RecordingId string `json:"recording_id"`
	Hostname    string `json:"host_name"`
```

handlers/misttriggers/push_out_start.go
```diff
@@ -0,0 +1,82 @@
...
+	pushUrl, err := url.Parse(destination)
+	if err != nil {
+		log.Printf("RecordingPushOutStart url.Parse %v", err)
+		return destination
+	}
+	if err = addUuidToUrl(pushUrl, event.RecordingId); err != nil {
+		log.Printf("RecordingPushOutStart addUuidToUrl() %v", err)
+		return destination
+	}
+	go clients.DefaultCallbackClient.SendRecordingEvent(event)
```
victorges, line 59: I wonder if we should make these hooks blocking instead. As in if they fail we should not let the stream go through either. Only opening for consideration tho, not sure what i prefer.

handlers/misttriggers/push_out_start.go
```diff
@@ -0,0 +1,82 @@
...
+
+// addUuidToUrl modifies pushUrl path from:
+//
+//	s3://livepeer-recordings-bucket/$stream/index.m3u8
+//
+// to:
+//
+//	s3://livepeer-recordings-bucket/$stream/<UUID>/index.m3u8
+func addUuidToUrl(pushUrl *url.URL, recordingUUID string) error {
+	path := strings.Split(pushUrl.EscapedPath(), "/")
```
victorges, line 71: We read from `.EscapedPath()` but actually save on `.Path`. Is that correct? Don't know this helper specifically but seems like we might be effectively modifying the path here regarding escaped chars.

handlers/misttriggers/push_out_start.go
```diff
@@ -0,0 +1,82 @@
...
+	last := len(path) - 1
+	path[last] = path[last-1]
+	path[last-1] = recordingUUID
+	if path[last-2] != "$stream" {
+		// Additional check to ensure we rewrite proper streams
+		return fmt.Errorf("stream-id variable not found in proper place")
+	}
+	pushUrl.Path = strings.Join(path, "/")
+	return nil
+}
```
victorges, line 82: WDYT of only doing a `strings.Replace(path, "$stream", "$stream/"+recordingId)`?

main.go
```diff
@@ -20,6 +21,8 @@ func main() {
 		return
 	}
 
+	config.RecordingCallback = *recording
```
victorges, line 24: **nit:** Can also use `flag.StringVar(&config.RecordingCallback, ...)` to avoid this step

victorges (approved):
LGTM!

main.go
AlexKordic: Nice

handlers/misttriggers/push_out_start.go
AlexKordic: Much simpler, great idea

handlers/misttriggers/push_out_start.go
AlexKordic: That URL struct is tricky

handlers/misttriggers/push_out_start.go
AlexKordic: Good practice is to return as soon as possible so mistserver can continue without additional latency increase. 

Lets consider a watchdog process that kills bad streams separately from our happy-path.

clients/callback_client.go
AlexKordic: good name

clients/callback_client.go
AlexKordic: Good point. This is served by following `c.DoWithRetries(req)` mechanism available for all callbacks to studio.

PR #50, opened 2022-09-26
Notify studio when recording completes

AlexKordic:


handlers/misttriggers/push_end.go
```diff
@@ -32,20 +35,39 @@ func (d *MistCallbackHandlersCollection) TriggerPushEnd(w http.ResponseWriter, r
...
+	case Recording:
+		d.RecordingPushEnd(w, req, streamName, actualDestination, pushStatus)
 	default:
 		// Not related to API logic
 	}
 }
 
+func (d *MistCallbackHandlersCollection) RecordingPushEnd(w http.ResponseWriter, req *http.Request, streamName, actualDestination, pushStatus string) {
+	var err error
+	pushSuccess := pushStatus == "null"
```
victorges, line 56: What is the expected behavior from Studio if this is `false`? Should it show an error to the user somehow or can it still recover the recording somehow?

handlers/misttriggers/push_end.go
```diff
@@ -72,3 +94,15 @@ func (d *MistCallbackHandlersCollection) SegmentingPushEnd(w http.ResponseWriter
...
+func uuidFromPushUrl(uri string) (string, error) {
+	pushUrl, err := url.Parse(uri)
+	if err != nil {
+		return "", err
+	}
+	path := strings.Split(pushUrl.EscapedPath(), "/")
+	if len(path) < 4 {
+		return "", fmt.Errorf("push url path malformed: element count %d %s", len(path), pushUrl.EscapedPath())
+	}
+	return path[len(path)-2], nil
```
victorges, line 107: Should we have a regex here instead? Perhaps validating some other parts of this URL as well. (I particularly don't remember what it is, from this. is it the `s3+https://` uri?)

victorges (approved):
LGTM from studio side of things!

handlers/misttriggers/push_end.go
AlexKordic: Fail in our first version. Improve by recovering in later tasks. Same applies for transcoding stages where some of them can fail independently. To be optimal we can just re-run failed stages. For now we re-run entire process.

handlers/misttriggers/push_end.go
AlexKordic: Good question. We now have plans to expand on URL path prefix with region and bucket name later. I figured its safer to grab path element before filename which should stay in this format in near future.

PR #53, opened 2022-09-29
config: MistProcLivepeer in /usr/local/bin/

yondonfu:
See livepeer/catalyst@83d3bcd4d8aa52795cb27883605bcf7e5a0d3dcc

MistProcLivepeer needs to be invoked in /usr/local/bin after the above commit that changes where MistController is invoked from.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -12,7 +12,7 @@ var Version string
 var Clock TimestampGenerator = RealTimestampGenerator{}
 
 // Path to Mist's "Livepeer" process that we shell out to for the transcoding
-const PathMistProcLivepeer = "/usr/bin/MistProcLivepeer"
+const PathMistProcLivepeer = "/usr/local/bin/MistProcLivepeer"
 
 // Port that the local Broadcaster runs on
 const DefaultBroadcasterPort = 8935


emranemran (approved):

config/config.go
```diff
@@ -12,7 +12,7 @@ var Version string
 var Clock TimestampGenerator = RealTimestampGenerator{}
 
 // Path to Mist's "Livepeer" process that we shell out to for the transcoding
-const PathMistProcLivepeer = "/usr/bin/MistProcLivepeer"
+const PathMistProcLivepeer = "/usr/local/bin/MistProcLivepeer"
```
hjpotter92, line 15: is it necessary to hardcode the full path to binary? wouldn't go be able to exec them from `$PATH` value?

yondonfu: If we can simplify this so that Go can exec MistProcLivepeer from `$PATH` instead of hardcoding a path here (which resulted in issues previously when the path that MistProcLivepeer was at changed), then that seems like a worthwhile improvement. @hjpotter92 mind tracking this is in an issue so this can be addressed?

yondonfu: Actually looks like this is tracked here https://github.com/livepeer/catalyst/issues/172

PR #54, opened 2022-10-06
Begin transcode right after PUSH_END trigger

emranemran:
Begin transcode right after PUSH_END trigger is receieved after segmentation is completed:

* generate transcodeRequest in SegmentingPushEnd handler (since transcodeRequest will not be generated from the HTTP API endpoint)
* update Segmenting cache Store function to store additional fields which will be used in transcodeRequest above

### DIFF
--- a/handlers/misttriggers/push_end.go
+++ b/handlers/misttriggers/push_end.go
@@ -130,7 +130,14 @@ func (d *MistCallbackHandlersCollection) SegmentingPushEnd(w http.ResponseWriter
 	}
 
 	// TODO: Start Transcoding (stubbed for now with below method)
-	stubTranscodingCallbacksForStudio(callbackUrl)
+	//stubTranscodingCallbacksForStudio(callbackUrl)
+
+	// Get the source stream's detailed track info before kicking off transcode
+	infoJson, err := d.MistClient.GetStreamInfo(streamName)
+	if err != nil {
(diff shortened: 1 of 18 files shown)


handlers/misttriggers/push_end.go
```diff
@@ -130,6 +131,13 @@ func (d *MistCallbackHandlersCollection) SegmentingPushEnd(w http.ResponseWriter
 
 	// TODO: Start Transcoding (stubbed for now with below method)
 	stubTranscodingCallbacksForStudio(callbackUrl)
+	
+	si := cache.DefaultStreamCache.Segmenting.Get(streamName)
+	transcodeRequest := handlers.TranscodeSegmentRequest {
+		SourceFile:    si.SourceFile,
+		CallbackUrl:   si.CallbackUrl,
+	}
+	go handlers.RunTranscodeProcess(d.MistClient, transcodeRequest)
```
victorges, line 140: Cool! I think for the `/vod` API behavior, you might want to wrap this into a local helper which handles the results sending the appropriate callbacks to Studio etc. I guess that's somewhat what you've done on your local change that you mentioned tho?

Btw does this `StreamInfo` have the "probing" information of the file? If so, should be really easy to generate the profiles copying that logic from `task-runner`

thomshutt (approved):

victorges:
Woohoo nice

PR #57, opened 2022-10-10
Add Manifest URL to output metadata

thomshutt:


handlers/misttriggers/triggers.go
```diff
@@ -33,6 +33,11 @@ func (d *MistCallbackHandlersCollection) Trigger() httprouter.Handle {
 		}
 
 		triggerName := req.Header.Get("X-Trigger")
+		_ = config.Logger.Log(
+			"msg", "Received Mist Trigger",
```
emranemran, line 37: minor: should we log similarly when we ask mist to add PUSH destinations? Might give us a clue as to why we see 2x PUSH_END triggers firing. I'm suspecting we're inadvertently adding an extra PUSH somewhere.

emranemran (approved):

handlers/misttriggers/triggers.go
thomshutt: Good call, will add in a new PR

thomshutt: https://github.com/livepeer/catalyst-api/pull/58

PR #61, opened 2022-10-13
Upload transcodes to output target location rather than input location

thomshutt:
Puts all transcoded segments + manifests under a `transcoded/` dir, rather than alongside the source MP4.

Doesn't account for multiple output locations yet, but is enough for us to test with.

image

### DIFF
--- a/cache/segmenting.go
+++ b/cache/segmenting.go
@@ -11,8 +11,8 @@ type SegmentingCache struct {
 
 type StreamInfo struct {
 	SourceFile            string
-	CallbackUrl           string
-	UploadDir             string
+	CallbackURL           string
+	UploadURL             string
 	AccessToken           string
 	TranscodeAPIUrl       string
 	HardcodedBroadcasters string
@@ -29,7 +29,7 @@ func (c *SegmentingCache) GetCallbackUrl(streamName string) string {
 	defer c.mutex.Unlock()
 	info, ok := c.cache[streamName]
 	if ok {
-		return info.CallbackUrl
+		return info.CallbackURL
 	}
 	return ""
 }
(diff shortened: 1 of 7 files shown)


cache/segmenting.go
```diff
@@ -11,8 +11,8 @@ type SegmentingCache struct {
 
 type StreamInfo struct {
 	SourceFile            string
-	CallbackUrl           string
-	UploadDir             string
+	CallbackURL           string
+	UploadURL             string
```
thomshutt, line 15: Changed Url -> URL because that's Go convention.

Changed "Dir" to URL because it's a URL and not a Dir

handlers/misttriggers/triggers.go
```diff
@@ -36,6 +36,7 @@ func (d *MistCallbackHandlersCollection) Trigger() httprouter.Handle {
 		_ = config.Logger.Log(
 			"msg", "Received Mist Trigger",
 			"trigger_name", triggerName,
+			"payload", string(payload),
```
emranemran, line 39: good idea

emranemran (approved):

handlers/transcode.go
```diff
@@ -107,11 +100,11 @@ func RunTranscodeProcess(mistClient clients.MistAPIClient, request TranscodeSegm
 		return fmt.Errorf("failed to start MistProcLivepeer: %s", err)
 	}
 
-	dir, _ := url.Parse(".")
-	uploadDir := inputUrl.ResolveReference(dir)
+	dir, _ := url.Parse("transcoded/")
```
hjpotter92, line 103: does it need to be hard-coded or can it be configurable in request payload?

thomshutt: At least initially, hardcoded - can look at making it variable in the future if needed

PR #65, opened 2022-10-17
Add callback heartbeats for in-flight transcodes

thomshutt:
This doesn't do the checking of transcode progress with Mist, but does provide the framework where we'd slot that in.

For now, if an item is in the "Transcode" cache, we update Studio every 15 seconds. I'd guess this is also going to expose some bugs where we're not removing things from the cache when we should be.

### DIFF
--- a/cache/transcoding.go
+++ b/cache/transcoding.go
@@ -2,8 +2,10 @@ package cache
 
 import (
 	"sync"
+	"time"
 
 	"github.com/livepeer/catalyst-api/clients"
+	"github.com/livepeer/catalyst-api/config"
 )
 
 type TranscodingCache struct {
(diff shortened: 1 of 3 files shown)


cache/transcoding.go
```diff
@@ -34,6 +36,29 @@ type SegmentInfo struct {
...
+func (t *TranscodingCache) SendTranscodingHeartbeats(interval time.Duration, quit chan bool) {
+	for {
+		// Stop the infinite loop if we receive a quit message
+		select {
+		case <-quit:
+			return
+		default:
+		}
+
+		jobs := t.GetAll()
```
AlexKordic, line 49: Note: seems possible to have a race between this thread and transcode thread. We should expect on studio side to receive a heartbeat callback after transcode-complete callback. We should notify studio team about this quirk.

AlexKordic (approved):
Well done!

cache/transcoding.go
thomshutt: Good point! I think Studio already has logic where they'll throw away messages if "Completion %" is less than one they've already received, which is quite a nice way of getting some thread safety around this stuff without them implementing a full on state machine

emranemran (approved):

PR #68, opened 2022-10-18
Check status for Segmenting PUSH_END trigger

thomshutt:


test/cucumber_test.go
```diff
@@ -65,7 +65,9 @@ func InitializeScenario(ctx *godog.ScenarioContext) {
 				fmt.Println("Error while killing app process:", err.Error())
 			}
 			if err := app.Wait(); err != nil {
-				fmt.Println("Error while waiting for app to exit:", err.Error())
+				if err.Error() != "signal: killed" {
```
thomshutt, line 68: This is just to suppress an error being printed when the app has aactually already been successfully killed

AlexKordic (approved):
I was also confused when exactly is trigger indicating failure..

PR #72, opened 2022-10-19
Don't fail codecov if nothing significant has changed

thomshutt:
Trying to avoid this: 
image

(seen on most recent commit to `main` https://github.com/livepeer/catalyst-api/commit/a637409b81b9342a4246694e1e7048d18c57368f)

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,0 +1,7 @@
+coverage:
+  status:
+    project:
+      default:
+        target: auto
+        threshold: 0.1%
+        base: auto 


hjpotter92:
Added it to the global yaml config

> image

Current global config is:

```yaml
codecov:
  bot: codecov-io
  require_ci_to_pass: true
comment:
  behavior: default
  layout: reach,diff,flags,files,footer,tree,betaprofiling
  require_changes: true
  show_critical_paths: true
coverage:
  status:
    project:
      default:
        target: auto
        threshold: 0.1%
        base: auto 
  precision: 5
  range: 20..40
  round: nearest
parsers:
  gcov:
    branch_detection:
      conditional: true
      loop: true
      macro: false
      method: false
```

thomshutt:
Awesome, didn't know we had global config. Thanks @hjpotter92!

PR #74, opened 2022-10-19
Create basic RECORDING_END trigger handler

thomshutt:
Doesn't do anything yet, but all the boilerplate we'll need to handle this trigger (which should make this following PR easier to read).


AlexKordic (approved):
Very clear, well done

PR #76, opened 2022-10-20
cache: print cache for easier debugging

emranemran:
It's nice to see print the cache contents when things are stored/deleted from the cache - makes debugging certain situations easier.

### DIFF
--- a/cache/transcoding.go
+++ b/cache/transcoding.go
@@ -4,6 +4,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/kylelemons/godebug/pretty"
 	"github.com/livepeer/catalyst-api/clients"
 	"github.com/livepeer/catalyst-api/config"
 )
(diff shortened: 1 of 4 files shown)


cache/transcoding.go
```diff
@@ -134,7 +137,16 @@ func (c *TranscodingCache) GetAll() map[string]*SegmentInfo {
 
 func (c *TranscodingCache) Store(streamName string, info SegmentInfo) {
 	c.mutex.Lock()
+	defer c.mutex.Unlock()
 	info.updatedAt = time.Now()
 	c.pushes[streamName] = &info
-	c.mutex.Unlock()
+	c.Print()
```
thomshutt, line 143: We should definitely print information about what's happening (Storing to / removing from the cache) and I don't love this "Debug" flag. Let's just be bold and log it until we hit traffic levels where the logging volume is an issue

AlexKordic (approved):

emranemran:
Updates: 
* address comments
* added `pretty` module for pretty printing cache tables -- I was hesitant to add yet another package but the benefits of being able to grasp info from noise logs outweighed the negatives.

PR #78, opened 2022-10-20
Move from using PUSH_END to signal to end of the segmenting flow to using RECORDING_END

thomshutt:


handlers/misttriggers/recording_end.go
```diff
@@ -72,6 +73,19 @@ func (d *MistCallbackHandlersCollection) triggerRecordingEndSegmenting(w http.Re
...
+	var inputVideoLengthMillis int
+	for track, trackInfo := range streamInfo.Meta.Tracks {
+		if strings.Contains(track, "video") {
+			inputVideoLengthMillis = trackInfo.Lastms
+		}
+	}
+	if math.Abs(float64(inputVideoLengthMillis-p.StreamMediaDurationMillis)) > 500 {
+		_ = config.Logger.Log("msg", "Input video duration does not match segmented video duration",
+			"input video duration (ms):", inputVideoLengthMillis, "segmented video duration (ms):", p.StreamMediaDurationMillis)
+		return
```
emranemran, line 86: I'm thinking we should have a callback to Studio here to indicate the segmented file's duration doesn't match the input file's duration. Currently arbitrarily set a 500ms delta as "acceptable". wdyt @victorges @thomshutt?

emranemran:
Tested this PR locally and worked as expected - need to sanity check on canary prior to merging.

handlers/misttriggers/recording_end.go
thomshutt: Yeah, we definitely need an error callback anywhere we're returning early. Delta sounds good, could maybe even be more lenient to begin with

victorges: Agreed, if this is a terminal error there must be an error callback. Might even be an unretriable one? Or is this a transient error?

The contract is that the upload process will always end with a callback. It can't just disappear unless something unexpected happens with the service like the machine or pod shutting down (then it's the regular "task lost" which we'll retry).

emranemran: OK, will follow up in https://github.com/livepeer/catalyst-api/issues/82 to address these integration points.

emranemran (approved):

PR #89, opened 2022-10-26
Write Master + Rendition manifests

thomshutt:


transcode/manifest.go
```diff
@@ -2,46 +2,127 @@ package transcode
...
+func GenerateAndUploadManifests(sourceManifest m3u8.MediaPlaylist, targetManifestOSURL string, transcodeProfiles []cache.EncodedProfile) error {
+	// Generate the base target OS URL to which we can append filenames / subdirectories to
+	baseURL := targetManifestOSURL[:strings.LastIndex(targetManifestOSURL, "/")]
+
+	// Generate the master + rendition output manifests
+	masterPlaylist := m3u8.NewMasterPlaylist()
+
+	for i, profile := range transcodeProfiles {
+		// For each profile, add a new entry to the master manifest
+		masterPlaylist.Append(
```
emranemran, line 70: According to the Apple HLS spec, the master playlist is supposed to be in reverse descending order of bitrate then resolution. I think that's how the HLS players need the master playlist to be ordered (?). If so, I don't believe we're guaranteed to get the highest bitrate rendition first from the BOT flow. I added reverse sorting playlist logic in the previous flow that might help if the HLS spec is how it's supposed to be implemented.

emranemran:
Just one comment that can be addressed in follow-up if needed but otherwise LGTM.

emranemran (approved):

transcode/manifest.go
thomshutt: Good point, I created https://github.com/livepeer/catalyst-api/issues/94 to do it as a follow-up once transcoding is in. I'll have to look at the spec, but I seem to remember that they should be generally ordered like that but that there's some more subtle stuff around which one comes first

PR #93, opened 2022-10-27
Test vod endpoint

red-0ne:
Add `/api/vod` test scenario using godog. It consists of initializers for Mist mock, Studio mock and minio (they are restarted after each scenario) and more generic steps to prepare and submit http requests. The mocking structures are only returning valid responses without downloading or uploading anything to minio yet.


thomshutt:
DMed @red-0ne but should've put here for visibility - there are a few linting checks failing at the moment

thomshutt:
This is a great start, thanks @red-0ne! Next step should probably be to have some assertions that our fake MistServer received the calls we expect

thomshutt (approved):

PR #98, opened 2022-10-28
transcode: trigger the transcode via remote or local broadcaster

emranemran:
If the access-token is provided in the transcode request, the remote broadcaster (via livepeer.com) will be used to transcode. Otherwise, the local broadcaster will be used to transcode.


emranemran:
The transcoded segments are being returned correctly, but still needs a call to upload to s3.

AlexKordic (approved):
Looks good, LGTM

thomshutt (approved):

PR #103, opened 2022-10-28
Transcoding stage: use SourceFile instead of UploadURL

AlexKordic:
Segmenting stage SourceFile is mp4, produces m3u8 that is SourceFile of transcoding stage.
We want to place transcoding paths in same dir as mp4 file so we should set UploadURL of transcoding stage to same value as SourceFile.


victorges:
Are we still merging this? Seems to be a month old and some of the fields here don't exist anymore (I think `UploadURL` became `TargetURL`)? There was also the refactor I made for the MediaConvert pipeline which moved some things around.

LMK if we still need this and I can help with the rebase, giving pointers here about where each code piece went (spoiler: most are in `pipeline/mist.go` now)

thomshutt:
Yeah, let's close for now and revisit if needed

Issue #99, opened 2022-10-28
Transcode of BBB fails with "Invalid Duration"


thomshutt:
Request:

```
{   
    "url": "s3+https://REDACTED:REDACTED@storage.googleapis.com/alexk-dms-upload-test/test-thom/src/bbb.mp4",
    "callback_url": "https://webhook.site/cd72124f-33a2-47e0-aacb-562008b91cc1",
    "output_locations": [
        {                      

                                "type": "object_store",
                                "url": "s3+https://REDACTED:REDACTED@storage.googleapis.com/alexk-dms-upload-test/test-thom/vod/segmented-bbb/source.m3u8",         
                                "outputs": {
                                        "source_segments": true
                                }
                        }
    ]
}
```

Source manifest:

[segmented-manifest.txt](https://github.com/livepeer/catalyst-api/files/9888387/segmented-manifest.txt)

```[2022-10-28 12:15:22] livepeer (387302) INFO: E1028 12:15:22.147193  387325 mediaserver.go:745] http push error processing segment url=http://127.0.0.1:8935/live/egeedeff/0.ts manifestID=egeedeff err="invalid duration 596.427"```

**Comment by AlexKordic** [comment]
Trace:
- [go-livepeer](https://github.com/livepeer/go-livepeer/blob/master/server/mediaserver.go#L973) parsing of duration
- [catalyst-api](https://github.com/livepeer/catalyst-api/blob/main/clients/broadcaster.go#L88) setting the duration

Because of `err="invalid duration 596.427"` we see `mediaDurationMillis` was `596427` when passed to `transcodeSegment()`. It is 1000x more than normal.

**Comment by thomshutt** [comment]
Weirdly, the source manfiest we're working from starts with

```
#EXTM3U
#EXT-X-VERSION:3
#EXT-X-PLAYLIST-TYPE:VOD
#EXT-X-TARGETDURATION:5
#EXT-X-MEDIA-SEQUENCE:0
#EXTINF:10.4160000000,
0.ts
```

PR #105, opened 2022-10-30
Pass segment duration to transcoder instead of overall duration

thomshutt:
Resolves https://github.com/livepeer/catalyst-api/issues/99


AlexKordic (approved):
Well Done!

PR #107, opened 2022-10-31
transcode: update logic for uploading renditions to handle remote Bro…

emranemran:
…adcaster transcodes


thomshutt:
Duration thing is fixed in https://github.com/livepeer/catalyst-api/pull/105

Let's hold off on using remote Bs for now and see if we can fix the local one today

AlexKordic:
Looks like duplicate https://github.com/livepeer/catalyst-api/pull/112 :)

PR #117, opened 2022-10-31
transcode: use seconds instead of milliseconds for target segment dur…

emranemran:
…ation

### DIFF
--- a/transcode/manifest.go
+++ b/transcode/manifest.go
@@ -59,7 +59,7 @@ func GetSourceSegmentURLs(sourceManifestURL string, manifest m3u8.MediaPlaylist)
 			urls,
 			SourceSegment{
 				URL:            u,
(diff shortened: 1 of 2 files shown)


victorges (approved):
LGTM

thomshutt (approved):

transcode/manifest.go
```diff
@@ -59,7 +59,7 @@ func GetSourceSegmentURLs(sourceManifestURL string, manifest m3u8.MediaPlaylist)
 			urls,
 			SourceSegment{
 				URL:            u,
-				DurationMillis: int64(segment.Duration * 1000),
+				DurationMillis: int64(segment.Duration),
```
victorges, line 62: Can we get milliseconds-precise durations at some point? Wouldn't be nice to round every duration to a second, even if only on the metadat

Issue #113, opened 2022-10-31
Error: "Too many connections for Stream ID"


thomshutt:
Transcoding fails after 10 segments - we should either reuse the connection or close it properly and use a unique stream name each time

```
[2022-10-31 14:16:35] livepeer (2305617) INFO: E1031 14:16:35.675550 2305625 mediaserver.go:367] manifestID=ccbbfbhd nonce=1130572396757980313 clientIP=127.0.0.1 Too many connections for streamID url=http://127.0.0.1:8935/live/ccbbfbhd/10.ts err=%!q(<nil>)
```

PR #121, opened 2022-10-31
Re-use Manifest ID with local B to avoid creating new sessions

thomshutt:
This is the second part of a fix for https://github.com/livepeer/catalyst-api/issues/113

We were seeing very slow transcode times and lots of logs where the B was struggling to find an O because we were passing in a blank manifestID and so it 
was creating a new session for each segment.

### DIFF
--- a/clients/broadcaster.go
+++ b/clients/broadcaster.go
@@ -61,19 +61,16 @@ type EncodedProfile struct {
 	ChromaFormat int32  `json:"chromaFormat,omitempty"`
 }
 
+var client = &http.Client{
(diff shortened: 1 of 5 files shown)


clients/broadcaster.go
```diff
@@ -61,19 +61,16 @@ type EncodedProfile struct {
 	ChromaFormat int32  `json:"chromaFormat,omitempty"`
 }
 
+var client = &http.Client{
```
thomshutt, line 64: Did some testing and didn't see any negative performance impact from reusing the HTTP client and Transport.

emranemran (approved):

Issue #111, opened 2022-10-31
Parallelise pushing of segments to Broadcaster


thomshutt:
```
NOTE: This shouldn't be done until we're confident that our current method works consistently
```

Currently we push segments to the Broadcaster sequentially, but should use a pre-defined level of parallelism have X goroutines pushing segments simultaneously for a performance boost.

PR #129, opened 2022-11-01
parallel segment push

AlexKordic:
Submitting multiple segments to local B node.

`config.TranscodingParallelJobs` defines number of parallel segments being transcoded.

parallel_local_b_load

Using `TranscodingParallelJobs int = 15` and additional interval with purpose to misalign job start of each parallel worker we get nice GPU load graph without spikes on local B:
desynced_parallel_local_b_load

This resolves https://github.com/livepeer/catalyst-api/issues/111

### DIFF
--- a/transcode/transcode.go
+++ b/transcode/transcode.go
@@ -127,8 +127,22 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
 	var completed sync.WaitGroup
 	completed.Add(config.TranscodingParallelJobs)
 	for index := 0; index < config.TranscodingParallelJobs; index++ {
-		go transcodeSegment(streamName, manifestID, transcodeRequest, transcodeProfiles, queue, errors, sourceSegmentURLs, targetOSURL, &completed)
-		time.Sleep(713 * time.Millisecond) // Add some desync interval to avoid load spikes on segment-encode-end
+		go func() {
+			defer completed.Done()
+			for segment := range queue {
+				err := transcodeSegment(segment, streamName, manifestID, transcodeRequest, transcodeProfiles, targetOSURL)
+				if err != nil {
+					errors <- err
+					return
+				}
+				var completedRatio = calculateCompletedRatio(len(sourceSegmentURLs), segment.Index+1)
+				if err = clients.DefaultCallbackClient.SendTranscodeStatus(transcodeRequest.CallbackURL, clients.TranscodeStatusTranscoding, completedRatio); err != nil {
+					log.LogError(transcodeRequest.RequestID, "failed to send transcode status callback", err, "url", transcodeRequest.CallbackURL)
+				}
+			}
+		}()
+		// Add some desync interval to avoid load spikes on segment-encode-end
+		time.Sleep(713 * time.Millisecond)
 	}
 	// Wait for all segments to transcode or first error
 	select {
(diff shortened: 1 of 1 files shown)


transcode/transcode.go
```diff
@@ -111,14 +111,57 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
...
+		queue <- segmentInfo{Input: u, Index: segmentIndex}
+	}
+	close(queue)
+	// Use channel for recieving the errors
+	errors := make(chan error, 100)
+	// Start number of workers in parallel
+	var completed sync.WaitGroup
+	completed.Add(config.TranscodingParallelJobs)
+	for index := 0; index < config.TranscodingParallelJobs; index++ {
+		go transcodeSegment(streamName, manifestID, transcodeRequest, transcodeProfiles, queue, errors, sourceSegmentURLs, targetOSURL, &completed)
```
thomshutt, line 126: I don't love this pattern of passing channels around, because it makes the control of the flow harder to follow and also makes it difficult to test - I'd prefer to keep `transcodeSegment` as a simple function that knows how to transcode a segment and then have all of the management of the segment queue and allocation of work to goroutines sitting up at this level

transcode/transcode.go
```diff
@@ -111,14 +111,57 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
...
+		time.Sleep(713 * time.Millisecond) // Add some desync interval to avoid load spikes on segment-encode-end
```
thomshutt, line 127: Nice, good thinking

transcode/transcode.go
thomshutt: Also there's enough going on here that it's probably worth moving out of the main Transcode body, so the flow would be something like

<Transcode - high level iterative transcode process> -> <Parallel Segment Transcoder> -> <transcodeSegment>

transcode/transcode.go
```diff
@@ -115,59 +115,40 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
...
 			}
-		}
-
-		var completedRatio = calculateCompletedRatio(len(sourceSegmentURLs), segmentIndex+1)
-		if err = clients.DefaultCallbackClient.SendTranscodeStatus(callbackURL, clients.TranscodeStatusTranscoding, completedRatio); err != nil {
-			log.LogError(transcodeRequest.RequestID, "failed to send transcode status callback", err, "url", callbackURL)
-		}
+		}()
+		// Add some desync interval to avoid load spikes on segment-encode-end
+		time.Sleep(713 * time.Millisecond)
```
emranemran, line 145: Feels pretty arbitrary - are the load spikes on GPU or CPU? Did you see any issues with the load esp if it's GPU?

transcode/transcode.go
```diff
@@ -115,59 +115,40 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
...
+		go func() {
+			defer completed.Done()
+			for segment := range queue {
+				err := transcodeSegment(segment, streamName, manifestID, transcodeRequest, transcodeProfiles, targetOSURL)
+				if err != nil {
+					errors <- err
+					return
+				}
+				var completedRatio = calculateCompletedRatio(len(sourceSegmentURLs), segment.Index+1)
+				if err = clients.DefaultCallbackClient.SendTranscodeStatus(transcodeRequest.CallbackURL, clients.TranscodeStatusTranscoding, completedRatio); err != nil {
```
emranemran, line 139: hmm won't we get multiple completed callbacks now instead of just the one that Studio was expecting?

emranemran:
Would also recommend building the linux targets and copying over the `catalyst-api` to canary for some quick testing prior to merging to see what perf (or `top`) shows when a server is being used instead of a local machine.

transcode/transcode.go
AlexKordic: This is actually progress callback that looks like:
```json
{
  "completion_ratio": 0.6666666666666667,
  "status": "transcoding",
  "timestamp": 1667312537
}
```

And final completed callback is [here](https://github.com/livepeer/catalyst-api/blob/ak/parallel_seg_push/transcode/transcode.go#L167):
```json
{
  "completion_ratio": 1,
  "status": "success",
  "timestamp": 1667317010,
  "type": "video",
  "video_spec": {
    "format": "mp4",
    "tracks": [
      {
        "type": "video",
        "codec": "H264",
        "bitrate": 7508656,
        "duration": 184.541,
        "size": 0,
        "start_time": 0,
        "width": 2048,
        "height": 858
      },
      {
        "type": "audio",
        "codec": "AAC",
        "bitrate": 256000,
        "duration": 184.661,
        "size": 0,
        "start_time": 0,
        "channels": 2,
        "sample_rate": 48000,
        "sample_bits": 16
      }
    ],
    "duration": 184.64,
    "size": 183755148
  },
  "outputs": [
    {
      "type": "google-s3",
      "manifest": "s3+https:/GOOG1EXCXWNKVLKBEJS62OMGXMQ5C5FH67YZMQXYGFBQYQRBYXQQJT5OM6L4I:rzM9CFtiunoflGzBwU3c%2FOBw9D5hy0Ntbmmedu6x@storage.googleapis.com/alexk-dms-upload-test/bbb/long/transcoded/index.m3u8",
      "videos": null
    }
  ]
}
```

thomshutt:
Manually tested this on Canary with Alex today and it seems to work well. Progress callbacks occasionally out of order as expected, but this shouldn't be an issue (and we'll be refactoring the progress stuff to reduce the number of callbacks soon anyway)

thomshutt (approved):

PR #132, opened 2022-11-02
Order master playlist by bitrate and resolution

### LINKED ISSUE #94: Ordering of manifests should follow Apple's HLS spec
According to the Apple HLS spec, the master playlist is supposed to be in reverse descending order of bitrate then resolution. I think that's how the HLS players need the master playlist to be ordered (?). If so, I don't believe we're guaranteed to get the highest bitrate rendition first from the BOT flow.

red-0ne:
Order master playlist by descending bitrate or resolution if bitrate is the same.

Still need to confirm that the bandwidth is correctly sourced.

Closes #94

### DIFF
--- a/transcode/manifest.go
+++ b/transcode/manifest.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net/url"
 	"path/filepath"
+	"sort"
 	"strings"
 
 	"github.com/grafov/m3u8"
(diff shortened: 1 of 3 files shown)


thomshutt (changes requested):
Looks great! Could you split the sorting out into a function and add an explicit unit test for it though please

red-0ne:
Will wait for https://github.com/livepeer/catalyst-api/pull/135 to land before merging.

thomshutt (approved):

PR #133, opened 2022-11-02
Update directory tree of output segments from VoD pipeline

emranemran:
Fix https://github.com/livepeer/catalyst-api/issues/134

### DIFF
--- a/handlers/handlers_test.go
+++ b/handlers/handlers_test.go
@@ -126,7 +126,7 @@ func TestSuccessfulVODUploadHandler(t *testing.T) {
 		"output_locations": [
 			{
 				"type": "object_store",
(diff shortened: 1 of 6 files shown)


emranemran:
TODOs before merging: 
* Fix failing tests
* Update rendition folder names to use profile names instead
* Double check re-use of any unnecessary variables introduced in path/url manipulation

thomshutt (approved):
LGTM pending the changes you mentioned

emranemran:
These changes are being handled in a follow-up PR as it required refactor of larger parts for code readability: 
* Update rendition folder names to use profile names instead (example 360p/ instead of rendition-1/
* Update profile list to use 240p, 480p, 720p

Issue #127, opened 2022-11-01
EncodedProfile not passed to transcoding stage


AlexKordic:
Right now we always use `defaultTranscodeProfiles` because at the [end of segmenting stage](https://github.com/livepeer/catalyst-api/blob/main/handlers/misttriggers/recording_end.go#L92) we do not set `TranscodeSegmentRequest.Profiles`.

thomshutt:
Completed

PR #136, opened 2022-11-02
Ak/encoded profiles in segmenting

AlexKordic:
Fixes https://github.com/livepeer/catalyst-api/issues/127


thomshutt (approved):
Looks good! Would be great to add a unit test, probably for `transcode/transcode.go` to test the override behaviour (i.e that it uses this field rather than the default profiles)

thomshutt (approved):

PR #145, opened 2022-11-04
Fix failing cucumber test + unit test + remove unused code

thomshutt:


transcode/transcode.go
```diff
@@ -76,18 +76,6 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
...
-	targetTranscodedOutputPath := path.Join(targetTranscodedPath, targetTranscodedManifestFilename)
-	// Generate the manifest output URL from the manifest output path (e.g. s3+https://USER:PASS@storage.googleapis.com/user/hls/index.m3u8)
-	tpath, err := url.Parse(targetTranscodedOutputPath)
-	if err != nil {
-		return outputs, fmt.Errorf("failed to parse targetTranscodedOutputPath: %s", err)
-	}
-	targetTranscodedOutputURL := segmentedOutputManifestURL.ResolveReference(tpath)
-	fmt.Println(targetTranscodedOutputURL)
-	// Generate the rendition output URL (e.g. s3+https://USER:PASS@storage.googleapis.com/user/hls/)
 	tout, err := url.Parse(targetTranscodedPath)
```
emranemran, line 79: Don't we need the above deleted code to generate the filenames?

thomshutt: targetTranscodedOutputURL is just being printed though and not used, which is why I could delete it

emranemran: ooh yeah i left that there for a follow up PR to change the rendition folder structures. I'll add it back when I put up that PR.

emranemran (approved):

red-0ne (approved):
LGTM, had to resolve cucumber tests and `TestItCanTranscode` almost the same way on my current branch.

PR #146, opened 2022-11-04
transcode: Improve asset transcoded profiles

victorges:
This is to make some improvements to the asset profile generation logic, leveraging
from the existing logic in `task-runner` today. 

Main changes: 
 - There's a longer list of default profiles, but those are filtered out depending on the source video
 - With that, changed the logic on including the source profile in the output. It's always added now (since there will be no profile equal or better than it from the above process)


thomshutt (approved):
Amazing, thanks Victor!

PR #147, opened 2022-11-04
Implement a timer based progress reporter

### LINKED ISSUE #104: Don't spam callback URL with Progress callbacks
Currently we send one after every transcoded segment, but should figure out a way to do 1 every X seconds.

@victorges suggested:

> you could probably copy 100% of this progress reporter from the task-runner (we could move it to some shared utils lib at some point)
https://github.com/livepeer/task-runner/blob/13db95e3b5c784119d200629e443a32349d899d4/task/progress.go#L35
> actually, would only need to change what it does when it "sends a progress update" here
https://github.com/livepeer/task-runner/blob/13db95e3b5c784119d200629e443a32349d899d4/task/progress.go#L115

red-0ne:
Adding Victor since it's based on his reporter's implementation from task-runner.

Closes #104


progress/progress.go
```diff
@@ -0,0 +1,94 @@
...
+)
+
+var Clock = clock.New()
+
+var progressReportBuckets = []float64{0, 0.25, 0.5, 0.75, 1}
+
+const minProgressReportInterval = 10 * time.Second
+const progressCheckInterval = 1 * time.Second
+
+func ReportProgress(ctx context.Context, client clients.CallbackClient, url, taskID string, size uint64, getCount func() uint64, startFraction, endFraction float64) {
```
victorges, line 25: Nice! Btw this is a simpler implementation of the progress reporter. It has been refactored for a nicer interface which you can dynamically update the end fraction. I had sent the link to the right branch, but I guess you ended up basing this on `main`. If you wanna take a look on the new one, here it is: https://github.com/livepeer/task-runner/blob/c47a871f37f4df8e8bea72a0b7f6e97a1713b8e5/task/progress.go#L35

victorges (commented):
Nice! The progress reporter looks good, will leave the rest of the code for someone with catalyst context to review.

progress/progress.go
red-0ne: @victorges, just changed this to the new reporter's implementation. Would you please mind to take a look? BTW, if we pass a callback instead of a client, we would decouple the package from the notifier logic, WDYT?

progress/reader.go
```diff
@@ -0,0 +1,67 @@
...
+	"sync/atomic"
+)
+
+type ReadHasher struct {
+	r      io.Reader
+	md5    hash.Hash
+	sha256 hash.Hash
+}
+
+func NewReadHasher(r io.Reader) *ReadHasher {
```
thomshutt, line 18: I can't see this being used anywhere, is it still needed?

victorges: We are calculating the hashes on the studio side for now. Ideally Catalyst would eventually start doing that tho

progress/progress.go
victorges: Yeah I like the decoupling! Can even move it to `go-tools` if we deem appropriate 😄 (might get a little annoying for logging tho)

thomshutt:
Superseded by https://github.com/livepeer/catalyst-api/pull/218, which tracks all job statuses in one place and should mean we have a foundation for doing some of the load calculation / balancing work that's been discussed

PR #159, opened 2022-11-08
Update transcoded rendition folder names

emranemran:
Instead of rendition-* folder names, use the profile name (e.g. 720p,
1080p, etc) as the folder name to make it more user friendly.


thomshutt (approved):
Lovely

PR #163, opened 2022-11-09
Send the success callback after dtsh upload

AlexKordic:
First playback will fail via mistserver if .dtsh file is not present on S3.

Introducing `playback-ready` status callback when .dtsh file is created.

If studio is using mist playback then this additional callback signals ready state, if not then `success` (TranscodeStatusCompleted) signals ready state.

### DIFF
--- a/handlers/misttriggers/recording_end.go
+++ b/handlers/misttriggers/recording_end.go
@@ -144,6 +144,13 @@ func (d *MistCallbackHandlersCollection) triggerRecordingEndSegmenting(w http.Re
 			return
 		}
 
+		// When createDtsh() completes issue another callback signaling to studio playback is ready
+		defer func() {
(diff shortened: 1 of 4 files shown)


handlers/misttriggers/recording_end.go
```diff
@@ -144,6 +144,15 @@ func (d *MistCallbackHandlersCollection) triggerRecordingEndSegmenting(w http.Re
 			return
 		}
 
+		// When createDtsh() completes issue another callback signaling to studio playback is ready
+		defer func() {
```
thomshutt, line 148: Really nice use of `defer`! I like that it means we're not dependent on DTSH generation success to send the successful transcode callback

thomshutt (approved):

PR #164, opened 2022-11-09
Transcode completion % should reflect that there's now a DTSH generation stage after transcoding

thomshutt:


AlexKordic (approved):
LGTM!

PR #168, opened 2022-11-10
handlers: exit if a bad request or error is encountered

emranemran:


red-0ne (commented):
There is a missing one here https://github.com/livepeer/catalyst-api/blob/10dd863d55e1148e5505e079e4d22b0e809a336b/handlers/upload.go#L147

```
if err := clients.DefaultCallbackClient.SendTranscodeStatus(uploadVODRequest.CallbackUrl, clients.TranscodeStatusPreparing, 0.0); err != nil {
	errors.WriteHTTPInternalServerError(w, "Cannot send transcode status", err)
}
```

emranemran:
> There is a missing one here
> 
> https://github.com/livepeer/catalyst-api/blob/10dd863d55e1148e5505e079e4d22b0e809a336b/handlers/upload.go#L147
> 
> ```
> if err := clients.DefaultCallbackClient.SendTranscodeStatus(uploadVODRequest.CallbackUrl, clients.TranscodeStatusPreparing, 0.0); err != nil {
> 	errors.WriteHTTPInternalServerError(w, "Cannot send transcode status", err)
> }
> ```

@red-0ne I left that one out on purpose. That's the callback to studio so it didn't make sense to bail early if it happens to fail to notify studio - we should continue with everything anyways.

thomshutt (approved):

thomshutt:
LGTM, good job catching this @red-0ne!

red-0ne:
> > There is a missing one here
> > https://github.com/livepeer/catalyst-api/blob/10dd863d55e1148e5505e079e4d22b0e809a336b/handlers/upload.go#L147
> > 
> > ```
> > if err := clients.DefaultCallbackClient.SendTranscodeStatus(uploadVODRequest.CallbackUrl, clients.TranscodeStatusPreparing, 0.0); err != nil {
> > 	errors.WriteHTTPInternalServerError(w, "Cannot send transcode status", err)
> > }
> > ```
> 
> @red-0ne I left that one out on purpose. That's the callback to studio so it didn't make sense to bail early if it happens to fail to notify studio - we should continue with everything anyways.

You are right. Thanks for the clarification. Shouldn't we have this in a goroutine to save up some transcoding time then?

red-0ne:
Also, we shouldn't write the error to the response then proceed to write response bytes coming out of transcoding. IMHO we should not do anything with responses from `SendTranscodeStatus`.

thomshutt:
I think you're right on both counts @red-0ne - feel free to put in a followup PR to fix. I think the goroutine should be spawned intside the callback client and then it should just log and not return anything on errors

red-0ne:
> I think you're right on both counts @red-0ne - feel free to put in a followup PR to fix. I think the goroutine should be spawned intside the callback client and then it should just log and not return anything on errors

I've worked on this followup and it seems that it is easier to have the goroutine outside of the callback if we want to keep testing errors of the sendTranscodeStatus request without adding some mechanism to capture the error from outside. Otherwise we be removing 4 assertions from 2 tests of the callback client. WDYT?

PR #171, opened 2022-11-11
Pass millisecond duration correctly to Broadcaster

thomshutt:


transcode/manifest.go
```diff
@@ -64,7 +64,7 @@ func GetSourceSegmentURLs(sourceManifestURL string, manifest m3u8.MediaPlaylist)
 			urls,
 			SourceSegment{
 				URL:            u,
-				DurationMillis: int64(segment.Duration),
+				DurationMillis: int64(segment.Duration * 1000),
```
thomshutt, line 67: This is where the issue was - the duration in the source manifest is in Seconds

transcode/transcode.go
```diff
@@ -231,7 +231,7 @@ func transcodeSegment(segment segmentInfo, streamName, manifestID string, transc
 	}
 
 	for _, stats := range transcodedStats {
-		stats.BitsPerSecond = uint32(float64(stats.Bytes) * 8000.0 / float64(stats.DurationMs))
+		stats.BitsPerSecond = uint32(float64(stats.Bytes) * 8000.0 / float64(stats.DurationMs/1000))
```
thomshutt, line 234: Had to update our bitrate calculation, which is bits / second but uses a millisecond variable to calculate it

transcode/manifest.go
yondonfu:
```suggestion
				DurationMillis: int64(segment.Duration * time.Second),
```

WDYT?

transcode/transcode.go
yondonfu:
```suggestion
		stats.BitsPerSecond = uint32(float64(stats.Bytes) * 8000.0 / float64(stats.DurationMs/time.Second))
```

WDYT?

yondonfu (approved):
Just 2 nits - feel free to choose to address before merge.

LGMT

transcode/manifest.go
thomshutt: Much nicer!

emranemran (approved):

PR #172, opened 2022-11-11
Send a Segment Upload Multiplier of 3 to local broadcaster

thomshutt:
This is to be more lenient than the Live workflow when waiting for transcoders.

Broadcaster change in https://github.com/livepeer/go-livepeer/pull/2650

### DIFF
--- a/clients/broadcaster_local.go
+++ b/clients/broadcaster_local.go
@@ -30,7 +30,9 @@ func NewLocalBroadcasterClient(broadcasterURL string) (LocalBroadcasterClient, e
 }
 
 func (c LocalBroadcasterClient) TranscodeSegment(segment io.Reader, sequenceNumber int64, profiles []EncodedProfile, durationMillis int64, manifestID string) (TranscodeResult, error) {
-	conf := LivepeerTranscodeConfiguration{}
+	conf := LivepeerTranscodeConfiguration{
+		SegUploadTimeoutMultiplier: 3,
+	}
 	conf.Profiles = append(conf.Profiles, profiles...)
 	transcodeConfig, err := json.Marshal(&conf)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


clients/broadcaster_local.go
```diff
@@ -30,7 +30,9 @@ func NewLocalBroadcasterClient(broadcasterURL string) (LocalBroadcasterClient, e
 }
 
 func (c LocalBroadcasterClient) TranscodeSegment(segment io.Reader, sequenceNumber int64, profiles []EncodedProfile, durationMillis int64, manifestID string) (TranscodeResult, error) {
-	conf := LivepeerTranscodeConfiguration{}
+	conf := LivepeerTranscodeConfiguration{
+		SegUploadTimeoutMultiplier: 3,
```
yondonfu, line 34: Just noting that this means that the timeout for upload from B -> O will always be 3x the seg duration and the overall HTTP timeout between B and O for a single segment (including upload + transcode response) is max(4 x seg duration, 8 seconds).

So, given seg duration = N, we have the upload timeout = 3N and the HTTP timeout = 4N.

In the worst case, where the upload takes 3N, that leaves N for the transcode response (i.e. O completes transcode and returns the URLs for the renditions to B) before we hit the HTTP timeout.

I'm not sure if that is the right ratio between timeouts, but this is probably fine for now as we can tweak as is needed

References:

https://github.com/livepeer/go-livepeer/blob/master/server/segment_rpc.go#L473
https://github.com/livepeer/go-livepeer/blob/2f306f6afdb21535612bb82867c5417c4643fe94/common/util.go#L38

thomshutt: Yep, good point @yondonfu - my thinking was that if this somewhat improves things or at least leads to them timing out later in the request then we can look at multiplying the overall timeout as well

Issue #173, opened 2022-11-14
Stop transcoding on first error


AlexKordic:
Example error looks like:
```json
{
  "completion_ratio": 0,
  "error": "Transcoding Failed: failed to run TranscodeSegment: http POST(http://127.0.0.1:8935/live/manifest-cffcgbae/4.ts) returned 400 400 Bad Request",
  "status": "error",
  "timestamp": 1668189511
}
```

Also having further callbacks:
```json
{
  "completion_ratio": 0.45,
  "status": "transcoding",
  "timestamp": 1668189519
}
```

After an error further progress should not be reported to studio.

We have several segments being processed in parallel. When first transcoding error happens all further work should be halted. Segments submitted to Broadcaster are allowed to complete but any further transcoding should not happen.

PR #174, opened 2022-11-14
Stop transcoding on first error

AlexKordic:
- Stop all workers on first error
- Fix progress calculation
- Unit tests

Fixes https://github.com/livepeer/catalyst-api/issues/173


thomshutt (approved):
Good spot and great to get the parallelism factored out into its own method

PR #177, opened 2022-11-16
upload: generate dtsh before segmenting input file

emranemran:
After a push request is sent, Mist attempt to read the input file, generate the dtsh header if there is none, and then begin segmenting the file. For certain files (typically large in size), the segmenting step would fail and generate an single empty 0.ts segment before bailing.

This change attempts to workaround this raciness within Mist by attempting to create the dtsh using an out-of-band call to MistInMP4 before segmenting begins.

Note: requires fix from mist repo for MistInMP4 to work as expected: 
https://github.com/DDVTECH/mistserver/pull/131


handlers/upload.go
```diff
@@ -193,3 +200,26 @@ func (d *CatalystAPIHandlersCollection) processUploadVOD(streamName, sourceURL,
...
+	url.Fragment = ""
+	headerPrepare := exec.Command(path.Join(config.PathMistDir, "MistInMP4"), "-H", url.String(), "-g", "5")
+	if err = subprocess.LogOutputs(headerPrepare); err != nil {
+		return err
+	}
+	if err = headerPrepare.Start(); err != nil {
+		return err
+	}
+	go func() {
+		if err := headerPrepare.Wait(); err != nil {
```
thomshutt, line 220: Why do we do the `Wait()` in an async function? Don't we need to wait for it to finish to get the benefits of the DTSH file being there?

handlers/misttriggers/recording_end.go
```diff
@@ -153,16 +147,6 @@ func (d *MistCallbackHandlersCollection) triggerRecordingEndSegmenting(w http.Re
...
-		// prepare .dtsh headers for all rendition playlists
-		for _, output := range outputs {
-			// output is multivariant playlist
-			err := createDtsh(requestID, output.Manifest)
-			if err != nil {
-				// should not block the ingestion flow or make it fail on error.
-				log.LogError(requestID, "master createDtsh() failed", err, "destination", output.Manifest)
-			}
-		}
-
```
AlexKordic, line 165: We still need this for mist playback. Am i missing some info?

handlers/upload.go
AlexKordic: Good catch! My mistake from time when dtsh was being created after transcode response

handlers/upload.go
```diff
@@ -193,3 +200,26 @@ func (d *CatalystAPIHandlersCollection) processUploadVOD(streamName, sourceURL,
...
+			log.LogError(requestID, "createDtsh return code", err, "destination", destination)
+		}
+	}()
```
AlexKordic, line 223:
```suggestion
	if err := headerPrepare.Wait(); err != nil {
		log.LogError(requestID, "createDtsh return code", err, "destination", destination)
	}
```

handlers/upload.go
```diff
@@ -122,51 +123,74 @@ func (d *CatalystAPIHandlersCollection) UploadVOD() httprouter.Handle {
...
-		// This workaround copies the file from Arweave to S3 and then tells Mist to use the S3 URL
-		if clients.IsArweaveURL(uploadVODRequest.Url) {
-			newSourceOutputPath := path.Join(targetDirPath, "source", "arweave-source.mp4")
-			newSourceOutputPathURL, err := url.Parse(newSourceOutputPath)
-			if err != nil {
-				errors.WriteHTTPInternalServerError(w, "Cannot parse newSourceOutputPath", err)
-				return
+		// Once we're happy with the request, do the rest of the Segmenting stage asynchronously to allow us to
+		// from the API call and free up the HTTP connection
+		go func() {
```
AlexKordic, line 128: Smart choice

handlers/upload.go
emranemran: Yup, good catch - missed this as well.

handlers/misttriggers/recording_end.go
emranemran: Note: MistInHLS needs to be fixed - dtsh for playlist is still broken which is why i removed it originally. Following up here: https://github.com/livepeer/catalyst-api/issues/178

PR #179, opened 2022-11-17
transcode: fix bitrate (bps) calculation

emranemran:


thomshutt (approved):

transcode/transcode_test.go
```diff
@@ -129,9 +129,9 @@ func TestItCanTranscode(t *testing.T) {
 	// Confirm the master manifest was created and that it looks like a manifest
 	var expectedMasterManifest = `#EXTM3U
 #EXT-X-VERSION:3
-#EXT-X-STREAM-INF:PROGRAM-ID=0,BANDWIDTH=27428,RESOLUTION=2020x2020,NAME="0-2020p0"
+#EXT-X-STREAM-INF:PROGRAM-ID=0,BANDWIDTH=27,RESOLUTION=2020x2020,NAME="0-2020p0"
```
thomshutt, line 132: The reason the code had changed is because it makes the test look wonky and so I assumed it was wrong. If the change works then could you figure out what's going on here @emranemran - I'm guessing we're passing unrealistic data somewhere else in the test

thomshutt (approved):

transcode/transcode_test.go
emranemran: yup, this one boggled my mind as well. I just chalked it up to running on fumes late at night but looks like there's something else at play here.

emranemran: https://github.com/livepeer/catalyst-api/issues/186 so we don't forget

PR #204, opened 2022-11-21
Implement VOD metrics

### LINKED ISSUE #128: Add VOD metrics to Catalyst API
Initial set should include:

- Request Count
- Success / Error counts, tagged based on the stage in the pipeline where the failure occurred 
- Transcode time per segment

Callbacks
- Number of HTTP retries
- Average duration
- Number of failures (after retries)

External Calls (Catalyst API makes to other systems e.g Broadcaster, Mist)
- Number of HTTP retries
- Average duration
- Number of failures (after retries)

red-0ne:
Implement a metrics service that monitors http calls from clients (CallbackClient, MistClient and local/remote BroadcasterClient) In addition to ObjectStore client which unlike others, is not using `retryablehttp`

The current recorded metrics are:
* Number of HTTP retries for a successful request
* HTTP failure count
* The duration of a successful HTTP request

For the transcoding job the following are monitored:
* Requests count
* Successful requests count
* Transcoding duration per segment

The metrics type are probably to be changed/corrected

Closes #128


emranemran:
Nice work @red-0ne! I looked through the PR and didn't find anything that stands out (also i'm not very familiar with metrics/prometheus so using your PR to learn). Main thing to note here would be to make sure VOD pipeline's timings are not changed due to the new calls for logging metrics. 

As a side request for later and not this PR, could we add a metric that shows file-size vs time spent in pipeline to get renditions? Breaking this down by region would give us some decent insight into vod pipeline performance.

thomshutt (approved):
Great stuff @red-0ne! Let's get this merged and then figure out how to expose the endpoint

red-0ne:
> ... figure out how to expose the endpoint

If you are referring to `/metrics` endpoint then it's already exposed

thomshutt:
@red-0ne right, I meant figure out how we get our Prometheus reading from it

PR #205, opened 2022-11-21
Use big.Int to store timestamps that are potentially very big

thomshutt:
This was the error we were seeing:

```error parsing line 8 of RECORDING_END payload as an int. Line contents: 18446744073709551615. Error: strconv.ParseInt: parsing \"18446744073709551615\": value out of range"```

### DIFF
--- a/handlers/misttriggers/recording_end.go
+++ b/handlers/misttriggers/recording_end.go
@@ -3,6 +3,7 @@ package misttriggers
 import (
 	"fmt"
 	"math"
+	"math/big"
 	"net/http"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: handlers/misttriggers/recording_end.go [hunk 1]
```diff
@@ -180,8 +181,8 @@ type RecordingEndPayload struct {
 	ConnectionStartTimeUnix   int
 	ConnectionEndTimeUnix     int
 	StreamMediaDurationMillis int64
```
thomshutt, line 183: Deliberately not changing this until we see a giant one come in. There's no reason this should exceed `math.MaxInt` though

emranemran (approved):

PR #214, opened 2022-11-22
Fix 'http: wrote more than the declared Content-Length' error

thomshutt:
I thought we'd caught all of these, but looks like we were writing HTTP errors (but not returning) for callback failures


handlers/handlers_test.go
```diff
@@ -414,7 +414,7 @@ func TestWrongContentTypeVODUploadHandler(t *testing.T) {
 	router.POST("/api/vod", catalystApiHandlers.UploadVOD())
 	router.ServeHTTP(rr, req)
 
-	require.Equal(rr.Result().StatusCode, 415)
+	require.Equal(415, rr.Result().StatusCode)
```
thomshutt, line 417: Just because I noticed this while debugging - Expected value should come first

PR #216, opened 2022-11-22
transcode: improve low-bitrate profiles bitrate slightly for better playback

emranemran:
When the input video's height or bitrate combo is too low to meet the ABR ladder, the low-bitrate profile is used in addition to the input video's profile to have some ABR during playback. The low-bitrate profile 1/3's the input video profile resulting in pretty low quality video. Increase this to be 75% of the input video's resolution so that playback still feels good without experiencing artifacts.

\#fix https://github.com/livepeer/catalyst-api/issues/215

### DIFF
--- a/transcode/transcode.go
+++ b/transcode/transcode.go
@@ -261,7 +261,7 @@ func getPlaybackProfiles(iv clients.InputVideo) ([]clients.EncodedProfile, error
 }
 
 func lowBitrateProfile(video clients.InputTrack) clients.EncodedProfile {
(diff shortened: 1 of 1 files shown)


thomshutt:
75% feels a little pointless if the video is already low quality though, maybe 50%?

thomshutt (approved):
@emranemran I've switched it to half for now and will merge (since you're out for a while), but if you think this is wrong then let's discuss when you're back

PR #218, opened 2022-11-23
Progress reporting on timer

thomshutt:
Move from reporting progress on each state change to a global loop over all active jobs that sends callbacks every X seconds.


AlexKordic (approved):
Well done, nice redesign

clients/callback_client.go
```diff
@@ -27,113 +33,173 @@ func NewCallbackClient() CallbackClient {
...
-		return
-	}
-	if err := c.DoWithRetries(req); err != nil {
-		log.Printf("SendRecordingStarted callback %v", err)
-	}
+func recoverer(f func()) {
+	defer func() {
+		if err := recover(); err != nil {
+			log.LogNoRequestID("panic in callback goroutine, recovering", "err", err)
+			go recoverer(f)
```
victorges, line 60: Won't this create an eternal panic loop calling `f` in background? 🤔

thomshutt: Let's continue the discussion under https://github.com/livepeer/catalyst-api/pull/219

PR #219, opened 2022-11-23
Move Sleep inside recoverable function

thomshutt:


thomshutt:
@victorges I created a Go playground version: https://go.dev/play/p/RtV0P6-Mt8ox

victorges:
@thomshutt this is what I'll think will happen in practice: https://go.dev/play/p/SBnynx4f4Py

Meaning:
 - Some specific input will be given which always panics (e.g. hits a nil reference exception)
 - Retrying it won't make it work, it will just panic again the same way
 - So the eternal loop instates itself

The problem with your original example was that the main input to the functions is mutable, which is the `x` var. Each time the function executes the var will have a different value. So when you retry it works the first try (what I don't think would happen in a real scenario).

So I "fixed" (broke) it by moving the `x++` to outside the recovered function and making sure it gets a local copy and not the shared `x` var (so execution always has the same input). On that case, the `go reoverer(f)` will cause an infinite panic-recover loop in background.

clients/callback_client.go
```diff
@@ -46,8 +46,10 @@ func NewPeriodicCallbackClient(callbackInterval time.Duration) *PeriodicCallback
 func (pcc *PeriodicCallbackClient) Start() *PeriodicCallbackClient {
 	go func() {
 		for {
-			recoverer(pcc.SendCallbacks)
-			time.Sleep(pcc.callbackInterval)
+			recoverer(func() {
+				pcc.SendCallbacks()
+				time.Sleep(pcc.callbackInterval)
```
victorges, line 51: Maybe sleep before calling `SendCallbacks` to make sure to still backoff if `SendCallbacks` is panicking

victorges (approved):
LGTM, but I still think the `recoverer` should stop recursing on itself in background.

clients/callback_client.go
```diff
@@ -46,8 +46,10 @@ func NewPeriodicCallbackClient(callbackInterval time.Duration) *PeriodicCallback
 func (pcc *PeriodicCallbackClient) Start() *PeriodicCallbackClient {
```
victorges, line 46: Hmm just thought of something regarding this periodic callback client, which is how do we get its benefits without having "terminal" updates like success or error get delayed up to 10 seconds when they are ideally received ASAP.

IOW today if the processing finishes with a result or error, it will go through the same flow as regular progress updates and only get saved in memory, to be flushed when the background routine gets to it. For terminal updates on the transcoding tho I think we should make the update immediately, as we'd be just busy waiting for a timer to finish the job.

WDYT? Might need to be addressed on a separate issue anyway.

clients/callback_client.go
thomshutt: I thought that too, but then you always delay the first update. I think that would be fine in reality, but we might need to go and update some unit tests.

thomshutt: Decided to stop being lazy and just fix this now

victorges (approved):
LGTM!

PR #226, opened 2022-11-25
Fire first callback immediately

thomshutt:
This fixes a bug where if Arweave/IPFS copy takes too long, Studio times out


victorges (approved):
Nice

PR #228, opened 2022-11-29
pipeline,clients: Implement MediaConvert provider and pipeline

victorges:
This is to implement the core functionality of the MediaConvert pipeline for VOD uploads. It includes:
 - implementation of the MediaConvert transcode provider using AWS SDK (#192)
   - including the logic to copy files from/to S3 (#191) 
 - creation of CLI args to configure pipeline coordinator and the URL of the external transcode provider (#189)
   - including the creation of "transcode provider URL" format and a parser for it. Only implementation now is for MediaConvert
 - move any context about MediaConvert to the MediaConvert provider. Pipeline package only knows external transcoders now

I will work on adding some tests as well, but for now we can also merge this and start testing on staging already! And no problem deploying to prod, as long as we don't put the MediaConvert pipeline in foreground until we're confident about it.


clients/callback_client_status.go
```diff
@@ -123,13 +123,14 @@ func NewTranscodeStatusProgress(url, requestID string, status TranscodeStatus, c
...
 	return TranscodeStatusMessage{
-		URL:       url,
-		RequestID: requestID,
-		Error:     errorMsg,
-		Status:    TranscodeStatusError.String(),
-		Timestamp: config.Clock.GetTimestampUTC(),
+		URL:         url,
+		RequestID:   requestID,
+		Error:       errorMsg,
+		Unretriable: unretriable,
```
thomshutt, line 131: Nice, I realised the other day that we'd been neglecting this field

clients/mediaconvert.go
```diff
@@ -0,0 +1,155 @@
...
+	config := aws.Config{
+		Region:      aws.String("us-east-1"),
+		Credentials: credentials.NewSharedCredentials("", "default"),
+	}
+	sess, err := session.NewSession(&config)
+
+	svc := mediaconvert.New(sess)
+
+	// Add endpoint to the service client
+	svc.Endpoint = "https://vasjpylpa.mediaconvert.us-east-1.amazonaws.com"
```
victorges, line 47: The strings here should all come from the client config instead of being hardcoded. Check the new options struct I created. Also keep in mind that you should use static credentials [like this](https://github.com/livepeer/go-tools/blob/db20184c5f12055422db3c1b1447b6f05f64c99e/drivers/s3.go#L117) instead of shared ones which are not available in production. 

The session and service objects should also not be created on every run. They should be stored in the client struct instead and reused for every request.

clients/mediaconvert.go
```diff
@@ -0,0 +1,155 @@
...
+		if err != nil {
+			fmt.Println(err)
+		}
+		fmt.Println(*job.Job.Status)
+		if *job.Job.Status == "COMPLETE" {
+			fmt.Println("Job is complete")
+			break
+		}
+	}
+	return nil
```
victorges, line 150: All these errors need to be returned instead of logged. Ideally you should wrap them [in `fmt.Errorf` errors](https://github.com/livepeer/dms-api/blob/71bcd9d57c35cc7ca33a83d3bcf3b740bd25c0c6/pipeline/mediaconvert.go#L32) for more relevant error messages, but ok to just return the errors as is.

All the other logs should be either removed or replace with structured logging through the `log` package. [example](https://github.com/livepeer/dms-api/blob/71bcd9d57c35cc7ca33a83d3bcf3b740bd25c0c6/pipeline/coordinator.go#L216)

clients/mediaconvert.go
```diff
@@ -0,0 +1,155 @@
...
+	if err != nil {
+		fmt.Println(err)
+	}
+
+	// print job id
+	fmt.Println(*job.Job.Id)
+
+	// checking if the job is completed (every 3 seconds)
+	for {
+		time.Sleep(3 * time.Second)
```
victorges, line 137: This interval should be higher, something like `10s` should work well, but `15s` should be fine too.

clients/mediaconvert.go
```diff
@@ -0,0 +1,155 @@
...
+							MaxAbrBitrate: aws.Int64(8000000),
+						},
+					},
+				},
+			},
+			TimecodeConfig: &mediaconvert.TimecodeConfig{
+				Source: aws.String("ZEROBASED"),
+			},
+		},
+		Role:  aws.String("arn:aws:iam::055703596054:role/service-role/MediaConvert_Default_Role"),
```
victorges, line 119: This is also a client config and should not be hardcoded.

clients/mediaconvert.go
```diff
@@ -0,0 +1,155 @@
...
+		Queue: aws.String("arn:aws:mediaconvert:us-east-1:055703596054:queues/Default"),
```
victorges, line 120: This is an optional parameter which, for simplicity, I think we should not specify here. Just let it use the default queue (which is already this)

clients/mediaconvert.go
```diff
@@ -0,0 +1,155 @@
...
+		})
+		if err != nil {
+			fmt.Println(err)
+		}
+		fmt.Println(*job.Job.Status)
+		if *job.Job.Status == "COMPLETE" {
+			fmt.Println("Job is complete")
+			break
+		}
+	}
```
victorges, line 149: This loop should call the `input.ReportProgress` function on every poll with the current 0-1 completion ratio of the job.

hjpotter92 (approved):

PR #232, opened 2022-11-30
Return an HTTP 429 when we're over capacity

thomshutt:
Currently very crude, but protects us from being DOSed.

Future iterations will probably be 

1) Include job size and current completion % in the calculation
2) Include machine resource availability (CPU, memory, network) in the calculation

### DIFF
--- a/middleware/capacity.go
+++ b/middleware/capacity.go
@@ -1,0 +1,20 @@
+package middleware
+
+import (
+	"net/http"
+
+	"github.com/julienschmidt/httprouter"
+	"github.com/livepeer/catalyst-api/pipeline"
+)
+
+const MAX_JOBS_IN_FLIGHT = 10
+
+func HasCapacity(vodEngine *pipeline.Coordinator, next httprouter.Handle) httprouter.Handle {
+	return func(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+		if vodEngine.Jobs.Size() >= MAX_JOBS_IN_FLIGHT {
+			w.WriteHeader(http.StatusTooManyRequests)
+			return
+		}
+		next(w, r, ps)
+	}
+}
--- a/middleware/capacity_test.go
+++ b/middleware/capacity_test.go
@@ -1,0 +1,58 @@
+package middleware
+
+import (
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/julienschmidt/httprouter"
+	"github.com/livepeer/catalyst-api/pipeline"
+	"github.com/stretchr/testify/require"
+)
+
+func TestItCallsNextMiddlewareWhenCapacityAvailable(t *testing.T) {
+	// Create a next handler in the middleware chain, to confirm the request was passed onwards
+	var nextCalled bool
+	next := func(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+		nextCalled = true
+	}
+
+	// Set up the HTTP handler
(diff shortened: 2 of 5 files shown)


middleware/capacity.go
```diff
@@ -0,0 +1,20 @@
...
+
+	"github.com/julienschmidt/httprouter"
+	"github.com/livepeer/catalyst-api/pipeline"
+)
+
+const MAX_JOBS_IN_FLIGHT = 10
+
+func HasCapacity(vodEngine *pipeline.Coordinator, next httprouter.Handle) httprouter.Handle {
+	return func(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+		if vodEngine.Jobs.Size() >= MAX_JOBS_IN_FLIGHT {
```
victorges, line 14: This will count both catalyst and MediaConvert jobs, while we only need to limit on catalyst jobs. WDYT of having a function in the cache that returns a slice of all keys instead and here only count catalyst jobs? Might be worth a helper in the coordinator for that

middleware/capacity.go
```diff
@@ -0,0 +1,20 @@
+package middleware
+
+import (
+	"net/http"
+
+	"github.com/julienschmidt/httprouter"
+	"github.com/livepeer/catalyst-api/pipeline"
+)
+
+const MAX_JOBS_IN_FLIGHT = 10
```
victorges, line 10: I think Alex had thought of something shorter, maybe only 3. Wdyt?

pipeline/coordinator.go
```diff
@@ -194,6 +194,10 @@ func (c *Coordinator) TriggerRecordingEnd(p RecordingEndPayload) {
 	})
 }
 
+func (c *Coordinator) NumJobsInFlight() {
+
+}
```
victorges, line 199: 👀

victorges (approved):
LGTM

middleware/capacity.go
thomshutt: Good point, will update

middleware/capacity.go
thomshutt: I think we can definitely handle more than 3, will bring it down to 5 to begin with

pipeline/coordinator.go
thomshutt: Whoops

middleware/capacity_test.go
```diff
@@ -0,0 +1,60 @@
...
+	// Create a lot of in-flight jobs
+	coordinator := pipeline.NewStubCoordinator()
+	for x := 0; x < 5; x++ {
+		coordinator.StartUploadJob(pipeline.UploadJobPayload{
+			RequestID: fmt.Sprintf("request-%d", x),
+		})
+	}
+
+	// Set up the HTTP handler
+	handler := HasCapacity(coordinator, next)
```
victorges, line 49: Hmm I wonder how racy this is since the jobs run in background and the stub handlers return immediately. Does it pass consistently?

victorges (approved):
Nice!

middleware/capacity_test.go
thomshutt: Yep, I think because the stub handlers never trigger the Mist callbacks right? So the jobs stay in flight

victorges: Not sure actually. I think they may return an error which finishes the job and sends a callback

thomshutt: > it's the real handlers with mock clients instead

PR #234, opened 2022-12-01
Separate cucumber from unit tests

### LINKED ISSUE #213: Split out Cucumber tests and regular unit tests on CI
If we have them as separate jobs on Github Actions then we can speed up the overall build process

red-0ne:
Create a separate test job for cucumber tests that run in parallel of unit tests. Seems to be room for improvement by sharing the build  and mod download cache between jobs.

Should close #213

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -7,7 +7,7 @@ on:
       - main
 
 jobs:
-  test:
+  unit-test:
     name: Test the catalyst-api project
     runs-on: ubuntu-20.04
     steps:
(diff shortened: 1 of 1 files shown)


thomshutt (approved):
LGTM, let's get a review from @hjpotter92 too since he's Mr Github Workflow

.github/workflows/test.yaml
```diff
@@ -37,11 +37,37 @@ jobs:
           git diff --exit-code
 
       - name: Run tests with coverage
-        run: go test ./... --short --race --covermode=atomic --coverprofile=coverage.out
+        run: go test $(go list ./... | grep -v /test) --short --race --covermode=atomic --coverprofile=coverage.out
```
hjpotter92, line 40: wouldn't it be better to refer to cucumber specifically here (to avoid possible addition of tests in future)?

```sh
grep -v cucumber
```

hjpotter92:
looks good other than a minor question. feel free to merge

PR #237, opened 2022-12-03
clients/mediaconvert: Handle acceleration settings error

victorges:
There's another possible MediaConvert error code that requires to turn off Acceleration on the job. 
Make sure we handle that one as well.


iameli (approved):
LGTM

PR #238, opened 2022-12-03
mediaconvert: Handle secret acceleration error code

victorges:
Follow-up on #237. 

The error code is not the documented, it's a secret one 🙄 

Also fix a small issue in the external `pipeline.Handler`, which should have an output with
type `object_store` instead of `manifest` .


iameli (approved):
LGTM

PR #239, opened 2022-12-03
handlers,pipeline: Move segmenting target URL logic to mist pipeline

victorges:
Found a bug on the background mist job: Since the coordinator only changes the TargetURL when the job is in background (line 206), the mist job will actually not pick up that change since it uses the segmenting URL instead.

Fix this the right way, by moving the creation of the segmeting URL to the mist pipeline. It is only used there.

The only small issue is that it is still a "global" field in the job info, so I added a comment that it's only used in mist for now. We can do something more sophisticated is this grows (like a "handler data" inside the job info struct that can be of any type)

### DIFF
--- a/handlers/upload.go
+++ b/handlers/upload.go
@@ -7,7 +7,6 @@ import (
 	"mime"
 	"net/http"
 	"net/url"
-	"path"
 	"strings"
 
 	"github.com/julienschmidt/httprouter"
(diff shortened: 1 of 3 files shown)


victorges (commented):
proxying LGTM from Eli, reviewed on my PC

PR #240, opened 2022-12-03
clients/mediaconvert: Auto-rotate input video based on metadata

victorges:
FTR: We have this same issue in the old task-runner and the new catalyst pipelines. We need to implement the same auto-rotate logic based on video metadata on the Catalyst pipeline.


thomshutt (approved):

thomshutt:
Not related to this change, but unit tests are failing

victorges:
Yeah! I fixed them on the other PR

PR #241, opened 2022-12-03
clients: Create some tests for the mediaconvert client

victorges:
As part of tech debt week, I'm adding some unit tests to the MediaConvert client here.

Only some minimal changes to the actual core code, from some things I needed to change
to make the tests work to some minor issues I found on the code and could fix as well.

Added tests for some basic error cases as well as full successful run, using only temp local
files for the tests.

### DIFF
--- a/clients/mediaconvert_test.go
+++ b/clients/mediaconvert_test.go
@@ -16,6 +16,14 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+const dummyHlsPlaylist = `
+#EXTM3U
+
+#EXTINF:10,
+0.ts
+
+#EXT-X-ENDLIST`
+
 func TestOnlyS3URLsToAWSClient(t *testing.T) {
 	require := require.New(t)
 	awsStub := &stubMediaConvertClient{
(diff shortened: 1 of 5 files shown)


clients/mediaconvert_test.go
```diff
@@ -0,0 +1,298 @@
...
+	require.NoError(t, err)
+	require.NoError(t, inputFile.Close())
+
+	// use the random file name as the dir name for the transfer file
+	transferDir = path.Join(inputFile.Name()+"-dir", "transfer")
+	require.NoError(t, os.MkdirAll(transferDir, 0777))
+
+	cleanup = func() {
+		maxRetryInterval, config.DownloadOSURLRetries, pollDelay = oldMaxRetryInterval, oldRetries, oldPollDelay
+		require.NoError(t, os.Remove(inputFile.Name()))
```
thomshutt, line 261: If this fails, it'll not run the next step that removes the dir - this might actually be a case where we want `assert` instead of `require`

clients/mediaconvert_test.go
```diff
@@ -0,0 +1,298 @@
...
+		osTransferBucketURL: mustParseURL("file://" + transferDir),
+		client:              awsStub,
+	}
+	return
+}
+
+func mustParseURL(str string) *url.URL {
+	u, err := url.Parse(str)
+	if err != nil {
+		panic(err)
```
thomshutt, line 276: Why panic instead of passing in the `*testing.T`?

thomshutt:
Screenshot 2023-01-01 at 21 34 31

thomshutt (approved):
Couple of small comments but otherwise looks great!

clients/mediaconvert_test.go
victorges: Doesn't seem like we're using `asset` in `catalyst-api` so I just set the err on a variable to check in the end here.

clients/mediaconvert_test.go
victorges: Just test noobness back when I wrote this helper, will do

PR #265, opened 2022-12-05
pipeline: Measure separate handlers success rate and performance

victorges:
Just improving the core coordinator metric to be more informative. Might be all we need to evaluate pipelines.

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -12,11 +12,11 @@ type ClientMetrics struct {
 }
 
 type CatalystAPIMetrics struct {
-	UploadVODRequestCount    prometheus.Counter
-	UploadVODSuccessCount    prometheus.Counter
-	UploadVODFailureCount    *prometheus.CounterVec
-	UploadVODPipelineResults *prometheus.CounterVec
-	TranscodeSegmentDuration prometheus.Histogram
+	UploadVODRequestCount        prometheus.Counter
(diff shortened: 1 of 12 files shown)


metrics/metrics.go
```diff
@@ -12,11 +12,11 @@ type ClientMetrics struct {
 }
 
 type CatalystAPIMetrics struct {
-	UploadVODRequestCount    prometheus.Counter
-	UploadVODSuccessCount    prometheus.Counter
-	UploadVODFailureCount    *prometheus.CounterVec
-	UploadVODPipelineResults *prometheus.CounterVec
-	TranscodeSegmentDuration prometheus.Histogram
+	UploadVODRequestCount        prometheus.Counter
+	UploadVODSuccessCount        prometheus.Counter
```
thomshutt, line 16: Could we combine SuccessCount and FailureCount into CompletionCount and then have a "success" bool label on it?

thomshutt (approved):

metrics/metrics.go
victorges: Done! https://github.com/livepeer/catalyst-api/pull/265/commits/dcef659a0caed0cf339ddc1e77f1d1ed11296501?diff=split&w=1

I also made it a summary instead, which contains a count but will also measure the request latency! Also refactored the handler a bit so we don't have to repeat measurement code so much (there was one that we missed)

PR #268, opened 2022-12-07
pipeline,handlers: Allow overriding strategy on upload request

victorges:
Created a new field on the request and the simple logic in the controller to use it if provided.


thomshutt (approved):
Really nice!

PR #288, opened 2022-12-12
Implement VOD pipeline metrics

red-0ne:
Initial implementation of VOD pipeline metrics

Not all metrics are collected for MediaConvert strategy. `vod_source_segments` and `vod_source_duration` are unavailable for that flow, same for the labels `source_codec_video` and `source_codec_audio`. We could use some ffprobe or mediainfo go library to populate them.

Please, make sure to review the metrics types https://github.com/livepeer/catalyst-api/pull/288/files#diff-d715f391f2ffa038c3167bfea1adf741c1f5e39f08349fbbacab780a594146a4R120, not sure if these are the right types to use.

### DIFF
--- a/clients/mediaconvert.go
+++ b/clients/mediaconvert.go
@@ -291,40 +291,41 @@ func getFileHTTP(ctx context.Context, url string) (io.ReadCloser, error) {
 	return resp.Body, nil
 }
 
-func copyFile(ctx context.Context, sourceURL, destOSBaseURL, filename string) (*int64, error) {
+func copyFile(ctx context.Context, sourceURL, destOSBaseURL, filename string) (int64, error) {
 	ctx, cancel := context.WithTimeout(ctx, 10*time.Minute)
 	defer cancel()
-	content, err := getFile(ctx, sourceURL)
+	writtenBytes := ByteAccumulatorWriter{count: 0}
+	c, err := getFile(ctx, sourceURL)
 	if err != nil {
-		return nil, fmt.Errorf("download error: %w", err)
+		return writtenBytes.count, fmt.Errorf("download error: %w", err)
 	}
-	defer content.Close()
+	defer c.Close()
+
+	content := io.TeeReader(c, &writtenBytes)
 
 	err = UploadToOSURL(destOSBaseURL, filename, content, 1*time.Minute)
 	if err != nil {
-		return nil, fmt.Errorf("upload error: %w", err)
+		return writtenBytes.count, fmt.Errorf("upload error: %w", err)
 	}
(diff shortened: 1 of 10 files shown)


pipeline/coordinator.go
```diff
@@ -310,10 +329,33 @@ func (c *Coordinator) finishJob(job *JobInfo, out *HandlerOutput, err error) {
...
+
+	metrics.Metrics.VODPipelineMetrics.SourceBytes.
+		WithLabelValues(labels...).
+		Observe(float64(job.sourceBytes))
+
+	metrics.Metrics.VODPipelineMetrics.SourceDuration.
+		WithLabelValues(labels...).
+		Observe(float64(job.sourceDurationMs))
+
+	metrics.Metrics.VODPipelineMetrics.SourceSegments.
```
thomshutt, line 355: Don't we do this same one a few lines above?

pipeline/coordinator_test.go
```diff
@@ -287,6 +291,42 @@ func TestAllowsOverridingStrategyOnRequest(t *testing.T) {
 	require.Equal("catalyst_vod_bg_"+meconJob.RequestID, mistJob.StreamName)
 }
 
+func TestPipelineCollectedMetrics(t *testing.T) {
```
thomshutt, line 294: Very nice to have a test for this stuff!

transcode/transcode.go
```diff
@@ -85,7 +85,7 @@ func init() {
 	LocalBroadcasterClient = b
 }
 
-func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName string, inputInfo clients.InputVideo) ([]clients.OutputVideo, error) {
+func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName string, inputInfo clients.InputVideo, segmentsCounter func()) ([]clients.OutputVideo, error) {
```
thomshutt, line 88: I think I prefer having it return a struct with information about the source (i.e a third return object) to passing in this counting method

clients/mediaconvert.go
```diff
@@ -290,23 +291,43 @@ func getFileHTTP(ctx context.Context, url string) (io.ReadCloser, error) {
...
+	}
+
+	defer info.Body.Close()
+	defer sess.EndSession()
+
+	if err != nil {
+		return nil, err
+	}
+
+	return info.FileInfo.Size, nil
```
thomshutt, line 327: If this is all so that we can get the size, could we have the OS return this instead?

thomshutt (approved):

Issue #311, opened 2022-12-14
(MediaConvert) Retry for "Acceleration fault" errors


yondonfu:
When transcoding with the MC pipeline, MC will occasionally return an "Acceleration fault" error for certain inputs. According to [1], this error indicates that there was unexpected error with the accelerated transcoding of the job. If accelerated transcoding cannot be performed with MC on the input we should retry the input with accelerated transcoding disabled.

We currently track acceleration error codes [here](https://github.com/livepeer/catalyst-api/blob/9d60acffc66e7560b151142b87a5935083f3abfe/clients/mediaconvert.go#L28), but the list of acceleration codes is missing 1550 which is the code for the "Acceleration fault" error according to [1]. Any MC error that [includes an error code from this list](https://github.com/livepeer/catalyst-api/blob/9d60acffc66e7560b151142b87a5935083f3abfe/clients/mediaconvert.go#L168) will cause an `ErrJobAcceleration` to be returned which we already will detect and [retry the job without acceleration](https://github.com/livepeer/catalyst-api/blob/9d60acffc66e7560b151142b87a5935083f3abfe/clients/mediaconvert.go#L100).

[1] https://docs.aws.amazon.com/mediaconvert/latest/ug/mediaconvert_error_codes.html

yondonfu:
Closed by #313

PR #313, opened 2022-12-15
Retry for "Acceleration fault" errors

thomshutt:
Addresses https://github.com/livepeer/catalyst-api/issues/311


yondonfu (approved):
LGTM

It is unclear to me whether the Acceleration fault is a random error that can occur when acceleration is enabled regardless of the input or if it is an error that will occur consistently for a particular input. I wonder it could make sense to retry with acceleration *enabled* and then if we get another Acceleration fault then retry with acceleration *disabled*. The rationale for this tweak would be to avoid the slower transcode for a large input if possible. I think the change in this PR is fine for now though, but worth keeping the above in mind.

thomshutt:
@yondonfu I had the same thought - my reading of the "Contact AWS Support" is that this is more of a catchall for unexpected errors and so I'd guess it will sometimes be consistent and sometimes not.

Issue #231, opened 2022-11-29
Improve Arweave/IPFS S3 copy naming/file type handling


yondonfu:
Whenever we detect an IPFS or Arweave URL, we copy the file to S3 [here](https://github.com/livepeer/catalyst-api/blob/c42c128524a53ba87d22c5d3097d1131dd0235b7/pipeline/mist.go#L25).

There are a few problems here:

1. We always name the temp file `arweave-source.mp4` which can be confusing in logs if the original file is from IPFS and not Arweave
2. We assume that the file is a mp4 file and use the .mp4 extension in the temp file name

Some thoughts on how to address both 1 & 2.

I wonder if we can do a preliminary check here where we skip the copy if we know the file is not a video (this would help if someone gave us an Arweave/IPFS URL for a directory without a path to the file in the directory which we cannot handle). For example, maybe we can check the response headers to determine if it has a `video/*` mime type. If not, we can return an error early instead of having Mist ingest the file from S3 (I'm not sure if Mist will error early in that case)? And for the temp filename maybe we can just use the ID in the Arweave/IPFS URL.

PR #314, opened 2022-12-15
mime type checking and improve arweave/ipfs file naming

mjh1:
Fixes https://github.com/livepeer/catalyst-api/issues/231


pipeline/mist.go
```diff
@@ -75,6 +84,19 @@ func (m *mist) HandleStartUploadJob(job *JobInfo) (*HandlerOutput, error) {
...
 
+func isVideo(source string) (bool, error) {
+	header, err := clients.GetArweaveHeaders(source)
+	if err != nil {
+		return false, err
+	}
+	contentType := header.Get("Content-Type")
+	mediaType, _, err := mime.ParseMediaType(contentType)
+	if err != nil {
+		return false, err
```
mjh1, line 95: @thomshutt do you know if we're ok to require the content-type header or should we fail open instead?

thomshutt: I definitely don't think we should rely on it being set correctly

mjh1: yeah make sense, reverted that bit 👍

clients/arweave_ipfs_s3.go
```diff
@@ -51,3 +51,11 @@ func IsArweaveOrIPFSURL(arweaveOrIPFSURL string) bool {
 
 	return strings.Contains(u.Host, "arweave") || strings.Contains(u.Host, "w3s.link") || strings.Contains(u.Path, "ipfs")
 }
+
+func GetArweaveHeaders(arweaveURL string) (http.Header, error) {
```
thomshutt, line 55: This isn't Arweave-specific, so can probably have a more generic name

pipeline/mist.go
```diff
@@ -39,7 +40,15 @@ func (m *mist) HandleStartUploadJob(job *JobInfo) (*HandlerOutput, error) {
 	// Arweave URLs don't support HTTP Range requests and so Mist can't natively handle them for segmenting
 	// This workaround copies the file from Arweave to S3 and then tells Mist to use the S3 URL
 	if clients.IsArweaveOrIPFSURL(job.SourceFile) {
-		newSourceURL, err := inSameDirectory(job.TargetURL, "source", "arweave-source.mp4")
+		if !isVideo(job.RequestID, job.SourceFile) {
```
thomshutt, line 43: Does it make sense to do this check for all videos (i.e not just Arweave / IPFS ones)?

mjh1: sounds reasonable to me 👍

pipeline/mist.go
```diff
@@ -36,10 +38,17 @@ func (m *mist) HandleStartUploadJob(job *JobInfo) (*HandlerOutput, error) {
 	job.SegmentingTargetURL = segmentingTargetURL.String()
 	log.AddContext(job.RequestID, "segmented_url", job.SegmentingTargetURL)
 
+	if !isVideo(job.RequestID, job.SourceFile) {
+		return nil, fmt.Errorf("source was not a video: %s", job.SourceFile)
```
mjh1, line 42: does anyone know offhand how this error would end up appearing to the user? do we have a special way to signal "bad request" type errors to the user?

thomshutt: Not at the moment, this will just show up in logs

PR #315, opened 2022-12-15
Change vod_duration metric from Summary to Histogram

thomshutt:


thomshutt:
Closing - seems like Summary should be fine for our needs

PR #331, opened 2022-12-22
Redact S3 credentials in logs

leszko:
Remove logging S3 credentials in catalyst.

The drawback of this PR is that I needed to remover logging the `raw` output from the status updates. Alternatively, we could leave it, but then we would need to RegEx all logs to find `s3+https://` URL. Let me know your thoughts.

### DIFF
--- a/clients/callback_client.go
+++ b/clients/callback_client.go
@@ -102,10 +102,9 @@ func (pcc *PeriodicCallbackClient) SendTranscodeStatus(tsm TranscodeStatusMessag
 		}
 	}
 
-	rawStatus, _ := json.Marshal(tsm)
 	log.Log(tsm.RequestID, "Updated transcode status",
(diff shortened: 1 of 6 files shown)


thomshutt (approved):
This looks great, one thing I was thinking in the future was to have the redacting happen on every field that comes in via tha `AddContext` and `Log` methods to avoid the case where someone adds a new log line containing a URL without redacting

thomshutt:
Also, looks like you need to tweak some unit tests

leszko:
> This looks great, one thing I was thinking in the future was to have the redacting happen on every field that comes in via tha `AddContext` and `Log` methods to avoid the case where someone adds a new log line containing a URL without redacting

I changed it to work that way. PTAL.

thomshutt (approved):
LGTM!

PR #332, opened 2022-12-22
Include an app version label in metrics

thomshutt:
This is to let us reliably track what's deployed where and if errors are version-specific.


leszko (approved):
LGTM

PR #339, opened 2022-12-30
Use requestID in the transfer bucket path

leszko:
Introduce `requestID` into the transfer bucket path for MediaConvert. 

Before, we didn't use any randomness in the path. That was not an issue while using assets in Studio, because Studio creates a random ID for each asset. However, it became an issue while introducing Transcode API, where users can define their own output directory. Without any randomness, one request could overwrite another requests.

fix https://github.com/livepeer/catalyst/issues/344

### DIFF
--- a/clients/mediaconvert.go
+++ b/clients/mediaconvert.go
@@ -90,7 +90,7 @@ func (mc *MediaConvert) Transcode(ctx context.Context, args TranscodeJobArgs) er
 	if path.Base(args.HLSOutputFile.Path) != "index.m3u8" {
 		return fmt.Errorf("target URL must be an `index.m3u8` file, found %s", args.HLSOutputFile)
 	}
-	targetDir := path.Join(args.RequestID, getTargetDir(args.HLSOutputFile))
+	targetDir := getTargetDir(args)
 
 	mcInputRelPath := path.Join("input", targetDir, "video")
 	// AWS MediaConvert adds the .m3u8 to the end of the output file name
@@ -421,7 +421,11 @@ func trimBaseDir(osPath, filePath string) string {
 }
 
 // Returns the directory where the files will be stored given an OS URL
-func getTargetDir(url *url.URL) string {
+func getTargetDir(args TranscodeJobArgs) string {
+	var (
+		url       = args.HLSOutputFile
+		requestID = args.RequestID
+	)
 	// remove the file name
 	dir := path.Dir(url.Path)
 	if url.Scheme == "s3" || strings.HasPrefix(url.Scheme, "s3+") {
@@ -434,7 +438,7 @@ func getTargetDir(url *url.URL) string {
 			dir = ""
 		}
 	}
-	return dir
+	return path.Join(requestID, dir)
 }
 
 func contains[T comparable](v T, list []T) bool {
--- a/clients/mediaconvert.go
+++ b/clients/mediaconvert.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/livepeer/catalyst-api/config"
 	"io"
 	"net/http"
 	"net/url"
@@ -90,7 +91,7 @@ func (mc *MediaConvert) Transcode(ctx context.Context, args TranscodeJobArgs) er
 	if path.Base(args.HLSOutputFile.Path) != "index.m3u8" {
 		return fmt.Errorf("target URL must be an `index.m3u8` file, found %s", args.HLSOutputFile)
 	}
-	targetDir := path.Join(args.RequestID, getTargetDir(args.HLSOutputFile))
+	targetDir := path.Join(config.RandomTrailer(10), getTargetDir(args.HLSOutputFile))
 
 	mcInputRelPath := path.Join("input", targetDir, "video")
 	// AWS MediaConvert adds the .m3u8 to the end of the output file name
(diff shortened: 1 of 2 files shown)


thomshutt:
PR makes sense, my only concern is that I'd been thinking of the Request ID as a short-lived thing that just goes into logs and I'd be slightly worried that we'll change the format in the future in a way that's incompatible with it being part of a path. 

Could we instead invoke the https://github.com/livepeer/catalyst-api/blob/main/config/stream_names.go#L25 method and generate a unique X character string here?

leszko:
> PR makes sense, my only concern is that I'd been thinking of the Request ID as a short-lived thing that just goes into logs and I'd be slightly worried that we'll change the format in the future in a way that's incompatible with it being part of a path.
> 
> Could we instead invoke the https://github.com/livepeer/catalyst-api/blob/main/config/stream_names.go#L25 method and generate a unique X character string here?

Ok, updated. PTAL.

thomshutt (approved):
LGTM once the tests are fixed (I'm guessing they're hardcoding an expected path)

clients/mediaconvert.go
```diff
@@ -90,7 +91,7 @@ func (mc *MediaConvert) Transcode(ctx context.Context, args TranscodeJobArgs) er
 	if path.Base(args.HLSOutputFile.Path) != "index.m3u8" {
 		return fmt.Errorf("target URL must be an `index.m3u8` file, found %s", args.HLSOutputFile)
 	}
-	targetDir := getTargetDir(args.HLSOutputFile)
+	targetDir := path.Join(config.RandomTrailer(10), getTargetDir(args.HLSOutputFile))
```
victorges, line 94: AFAIK this will generate a completely random string, not use the request ID. Could we reuse `args.RequestID` here instead?

On another note, I think it would be much easier if we add the request ID to the end of this path, not to the beginning. So when we're looking for specific input/output files we can start from the playback ID of the asset and then find the corresponding request IDs inside. If we start with the request ID (or worse, with this random ID here) it will have an additional layer of indirection where we'll need to go through the logs to find out where each file went.

On a final and less important note, I'd also move this logic to the `getTargetDir` function, as the sole purpose of it is to generate this directory. It should have the logic of adding any randomness to it as well.

victorges (approved):
LGTM as is if you need to merge ASAP, but made a couple suggestions.

clients/mediaconvert.go
victorges: Oh just saw Thom's comment about not wanting to use the request ID here. I guess it makes sense, but it's yet another reason not to organize all the files on a folder with this random string instead, will get nearly impossible to find later (not sure if we're even logging this anywhere).

Personally, these files are temporary and so we don't have any restriction on their format or permanence after the output files are copied back to their final location. It seems fine to me to use the request ID in that case and think of something else if we ever change them to be "file name incompatible". We could perhaps URL-encode the request ID here just in case.

WDYT @thomshutt? If you still think it's better not to use them in these temporary file names, then I'd advocate strongly for my first point, of not using them as the root folder but as the final folder in the path.

leszko: To your points:

1. About using `requestID` vs `randomTrailer()` => I'm ok with both. Request ID is simpler for tracking, but on the other hand it may make the assumption we won't change it, so I get the point from Thom of not using it. @thomshutt please comment on this one.

2. Move logic to `getTargetDir` => ok, I'll move it.

3. Random string at the end of the path => I think it's a good idea. I'll move it.

thomshutt: I think that if there's no other easy way for us to correlate a given transcode to where the file is stored, then I'm okay with us using RequestID here - sorry for all the back and forth here @leszko!

leszko: > 3. Random string at the end of the path => I think it's a good idea. I'll move it.

Actually I just realized it's not that easy to do, because we copy the whole path recursively. Keeping it at the beginning.

victorges: > Actually I just realized it's not that easy to do, because we copy the whole path recursively. Keeping it at the beginning.

Hmm, not sure if I understand. Conceptually, if we had a logic like:
 - `path.Join(args.RequestID, getTargetDir(args.HLSOutputFile))
why can't we make it
 - `path.Join(getTargetDir(args.HLSOutputFile), args.RequestID)`
?

Doesn't really matter if the inner `getTargetDir` is recursive, only matters where we add the request ID in the output of that process. Am I misunderstanding something?

clients/mediaconvert.go
```diff
@@ -434,7 +438,7 @@ func getTargetDir(url *url.URL) string {
 			dir = ""
 		}
 	}
-	return dir
+	return path.Join(requestID, dir)
```
victorges, line 441: Exactly, here! Why doesn't this work?
```suggestion
	return path.Join(dir, requestID)
```

clients/mediaconvert.go
victorges: Just to be clear why I'm insisting on this, I think this debug experience would be pretty neat:
 - Get the task/asset ID from the corresponding asset to investigate, from Studio API
 - From either of those, find the asset playback ID
 - With that, access the S3 transfer bucket on the path `output/hls/<playbackId>`
 - Now here we'll find 1 bucket for each attempt of transcoding the file. Since Studio does 3 retries we'll find 3 folders, each with the corresponding request ID
 - Dig into each request ID ✨ 
 
That in opposition to first having to find all the request IDs for a given playback ID in the logs, then inverting the search in S3 starting with the request ID instead (which their dashboard also makes it harder to do).

leszko: @victorges I think you're right. Updated in https://github.com/livepeer/catalyst-api/pull/355. Please approve.

PR #345, opened 2023-01-03
postgres metric poc

mjh1:


pipeline/coordinator.go
```diff
@@ -361,9 +363,32 @@ func (c *Coordinator) finishJob(job *JobInfo, out *HandlerOutput, err error) {
 		WithLabelValues(labels...).
 		Add(float64(job.transcodedSegments))
 
+	postgresMetrics(job)
```
thomshutt, line 366: We should probably run this as a goroutine so that it's non-blocking

mjh1: yep definitely, i was still in the POC mindset :)

PR #359, opened 2023-01-09
Add explicit 'version' metric

thomshutt:


main.go
```diff
@@ -49,6 +49,10 @@ func main() {
 		HttpReqUrl:      fmt.Sprintf("http://localhost:%d", *mistHttpPort),
 		TriggerCallback: fmt.Sprintf("http://localhost:%d/api/mist/trigger", *port),
 	}
+
+	// Fire a metric to let us track the version of the app we're using
+	metrics.Metrics.Version.WithLabelValues("catalyst-api", config.Version).Inc()
```
victorges, line 54: Should this live in the `Metrics` constructor perhaps?

victorges (approved):
LGTM

main.go
thomshutt: Yeah good call, that's probably a better place

Issue #350, opened 2023-01-04
Retry failed IPFS downloads with multiple HTTP gateways


thomshutt:
We should begin with a list of 3-4 preferred gateways and then when the IPFS download stage fails, iterate through these.

For an example of where we do this on Studio, see: https://github.com/livepeer/task-runner/blob/6853b65ac1ba193e59875e6a04383d73d0744972/task/import.go#L126

@victorges is the best person to consult on questions as he implemented the above and has information on the dedicated Piñata gateway that should be our first option.

victorges:
Other resources from the existing implementation:
 - the config with the list of gateways that we can just copy to the catalyst config: https://github.com/livepeer/livepeer-infra/blob/4da583548d7ea7c1f11c14dbfda1715f2cd8c943/values/task-runner.values.yaml#L9
 - the helper CLI flag parsing code for URLs: https://github.com/livepeer/task-runner/blob/66ef87d7135321eb4fafc8e667a6db092fa77c46/cmd/task-runner/task-runner.go#L48
 - how these flags are then used: https://github.com/livepeer/task-runner/blob/6853b65ac1ba193e59875e6a04383d73d0744972/task/import.go#L126

Notice that we do need to parse the URLs at some point in the code in order to build the full file URLs, since the private Piñata gateway that we have has a query-string in the base URL for the auth token, meaning we can't just string-concat the base URL with the CID to get the full URL. I decided to make the parsing on the CLI-side on task-runner to make the core logic cleaner and also get automatic startup errors if invalid URLs are provided (instead of errors only when they are eventually used). I recommend following the same approach because of that.

PR #360, opened 2023-01-09
Download content addressed sources with retryability

### LINKED ISSUE #141: Support IPFS and Arweave URLs in Catalyst API
Related: https://github.com/livepeer/task-runner/issues/79

For the initial version, we should just pick a free gateway and convert them to HTTP(s) URLs when they first reach Catalyst API.

red-0ne:
Handle Arweave and IPFS content by selecting from configured gateway urls. It will retry a second time if none of the gateways resolved the content.

Closes #141, #350


clients/arweave_ipfs_s3.go
```diff
@@ -2,52 +2,99 @@ package clients
...
-func newDefaultRetryableHttpClient() *http.Client {
-	client := retryablehttp.NewClient()
-	client.RetryMax = 2                          // Retry a maximum of this+1 times
-	client.RetryWaitMin = 200 * time.Millisecond // Wait at least this long between retries
-	client.RetryWaitMax = 1 * time.Second        // Wait at most this long between retries (exponential backoff)
-	client.HTTPClient = &http.Client{
-		// Give up on requests that take more than this long - the file is probably too big for us to process locally if it takes this long
-		// or something else has gone wrong and the request is hanging
-		Timeout: MAX_COPY_DURATION,
+var retryStrategy = backoff.NewConstantBackOff(1 * time.Second)
```
victorges, line 19: Retry strategies are actually not thread-safe, as the backoff lib mutates them between retries (there was a race condition on the one used on object store client). 
Even though this is a constant backoff which might be immutable, I think it's safer to never create these statically and create one per request instead. WDYT?

clients/arweave_ipfs_s3.go
```diff
@@ -2,52 +2,99 @@ package clients
...
 	}
 
-	return nil
+	var opContent io.ReadCloser
+	downloadOperation := func() error {
+		for _, gateway := range gateways {
+			path, err := url.JoinPath(gateway.Path, parsedURL.Host)
+
+			if err != nil {
+				return fmt.Errorf("cannot build gateway path: %w", err)
```
victorges, line 55: WDYT of logging and skipping the gateway instead, just in case it's a bad config? (no idea what could make this `JoinPath` err tho)

Edit: This is not necessary if you switch to the better `JoinPath` function mentioned below (it doesn't even return an error)

clients/arweave_ipfs_s3.go
```diff
@@ -2,52 +2,99 @@ package clients
...
+			}
+
+			if gateway.Query().Has("pinataGatewayToken") {
+				values := gateway.Query()
+				values.Set("pinataGatewayToken", config.LP_PINATA_GATEWAY_TOKEN)
+				gateway.RawQuery = values.Encode()
+			}
```
victorges, line 62: I do not think this is necessary. The gateway URL config should already come with the proper value for the pinata gateway token. I don't think we need a separate config for the token and then re-inject it in the URL.

e.g.: https://github.com/livepeer/livepeer-infra/blob/6f2b075fa3c3a3589867848cf9a6fe78d59f826a/values/task-runner.values.yaml#L10
(this `{{ secrets.... }}` will get replaced by our CI on the deploy with the real thing)

clients/arweave_ipfs_s3.go
```diff
@@ -2,52 +2,99 @@ package clients
...
+
+			url := url.URL{
```
victorges, line 64: You can just use the `path` variable here (likely renaming itself to `url`, `fullURL` or sth like that)

Edit: Ohhhh, just realized now that you are using a different `JoinPath` function above. Instead of using `url.JoinPath(string, string) (string, error)`, you should use the `(*url.URL) JoinPath(string) *url.URL` instead and it will be much cleaner! It already starts from a URL and receives only some additional segments to add to the path, returning a new URL with that. It also never errs! So cleaner logic here.

The whole thing, since the beginning of the `for` loop iteration until this line can become a 1-liner:
```suggestion
			url := gateway.JoinPath(parsedURL.Host)
```

`JoinPath` will also keep the same query string of the URL, so no need to bother about that either.

clients/arweave_ipfs_s3.go
```diff
@@ -2,52 +2,99 @@ package clients
...
+			}
+			resp, err := http.DefaultClient.Get(url.String())
+			if err == nil {
+				if resp.StatusCode >= 300 {
+					resp.Body.Close()
+					return fmt.Errorf("unexpected response: %d", resp.StatusCode)
+				}
+				opContent = resp.Body
+				return nil
+			}
```
victorges, line 78: Can you add a log for the error here just so we don't ignore it?

clients/arweave_ipfs_s3.go
```diff
@@ -2,52 +2,99 @@ package clients
...
+				Scheme:   gateway.Scheme,
+				Host:     gateway.Host,
+				Path:     path,
+				RawQuery: gateway.RawQuery,
+			}
+			resp, err := http.DefaultClient.Get(url.String())
+			if err == nil {
+				if resp.StatusCode >= 300 {
+					resp.Body.Close()
+					return fmt.Errorf("unexpected response: %d", resp.StatusCode)
```
victorges, line 74: This should also not `return` completely and skip the loop, but rather `continue` and try the next gateway. 

**nit:** To make this whole control flow clearer and easier to maintain, it might make sense to create a function that tries 1 gateway download and returns the minimum set of information, only what's necessary in this loop to control the flow and the possible result, if successful. Like `downloadDStorageFromSingleGateway`, then this function here can focus more on the gateway list + retry logic.

clients/arweave_ipfs_s3.go
```diff
@@ -2,52 +2,99 @@ package clients
...
+		return nil, fmt.Errorf("failed to get CID from any gateway: %w", err)
+	} else {
+		return opContent, nil
+
+	}
 }
 
-func IsArweaveOrIPFSURL(arweaveOrIPFSURL string) bool {
-	u, err := url.Parse(arweaveOrIPFSURL)
+func IsContentAddressedResource(dStorage string) bool {
```
victorges, line 93: Arweave is not content-address, so this name is innacurate. WDYT of `IsDStorageURL/Resource`?

clients/mediaconvert.go
```diff
@@ -461,3 +464,19 @@ func contains[T comparable](v T, list []T) bool {
...
+	client.HTTPClient = &http.Client{
+		// Give up on requests that take more than this long - the file is probably too big for us to process locally if it takes this long
+		// or something else has gone wrong and the request is hanging
+		Timeout: MAX_COPY_DURATION,
+	}
+
+	return client.StandardClient()
+}
+
+var retryableHttpClient = newRetryableHttpClient()
```
victorges, line 482: Can you move this to the top of the file? (only the global var, not the helper)

victorges (changes requested):
Looks good overall, but I think there is an issue with the gateway-iterating logic where the function will return/stop on any error instead of trying the next gateway. The other comments are mostly suggestions.

clients/arweave_ipfs_s3.go
red-0ne: Totally agree! Seems that I had trouble with naming things :smile: mixing ContentAddressed and dStorage terms.

clients/arweave_ipfs_s3.go
red-0ne: You mean it is enough having this in `main.go`
```go
URLSliceVarFlag(&config.ImportIPFSGatewayURLs, "import-ipfs-gateway-urls", "...,https://vod-import-gtw.mypinata.cloud/ipfs/?pinataGatewayToken={{secrets.LP_PINATA_GATEWAY_TOKEN }}", "...")
```

and drop this one?
```go
flag.StringVar(&config.LP_PINATA_GATEWAY_TOKEN, "lp-pinata-gateway-token", "", "Token used to access pinata gateways")
```

I can't see where it would be parsed :thinking:

red-0ne: Ok I see now :smile:

thomshutt:
LGTM, it'd be good to add a test for "does it retry with each gateway?" though

clients/arweave_ipfs_s3.go
```diff
@@ -2,52 +2,96 @@ package clients
...
+	if resp.StatusCode >= 300 {
+		resp.Body.Close()
+		log.Log(requestID, "unexpected response from gateway", "status_code", resp.StatusCode, "url", fullURL)
+		return nil
+	}
+
+	return resp.Body
 }
 
-func IsArweaveOrIPFSURL(arweaveOrIPFSURL string) bool {
```
victorges, line 46: Hmmm I just noticed a regression here in case the input is a URL to an IPFS/Arweave gateway. On that case, we also need to copy the file to an S3 bucket in order to get proper `Content-Length`, range requests, etc. If we don't copy these files from their gateways to an S3 bucket, the Mist processing *will* fail (which was the reason to create this copying in the first place)

We could either keep this "gateway URL identification" logic and replace it with the protocol-specififc scheme (like `ipfs://` or `ar://`), but that would potentially make the downloads slower or even not work at all (assuming the gateway provided by the user is the best gateway, possibly where the file resides).

So I think we need to either:
 - Change this function to also detect the old gateway URLs as DStorage URLs (then it might need to make the actual download function support these too), or;
 - Create a new function like `IsDStorageGatewayURL` which is used only on the place that used `IsArweaveOrIPFSURL` to copy it to an S3 before processing (in the Mist pipeline)

WDYT?

clients/arweave_ipfs_s3.go
```diff
@@ -2,52 +2,96 @@ package clients
...
 	}
 
-	return nil
+	if resp.StatusCode >= 300 {
+		resp.Body.Close()
+		log.Log(requestID, "unexpected response from gateway", "status_code", resp.StatusCode, "url", fullURL)
+		return nil
+	}
+
+	return resp.Body
```
victorges, line 87: Nice!

config/config.go
```diff
@@ -27,3 +28,9 @@ var TranscodingParallelJobs int = 2
 var TranscodingParallelSleep time.Duration = 713 * time.Millisecond
 
 var DownloadOSURLRetries uint64 = 10
+
+var ImportIPFSGatewayURLs []*url.URL = []*url.URL{}
+
+var ImportArweaveGatewayURLs []*url.URL = []*url.URL{}
+
+var LP_PINATA_GATEWAY_TOKEN string
```
victorges, line 36: I think this is unused now

pipeline/mist.go
```diff
@@ -41,21 +41,21 @@ func (m *mist) HandleStartUploadJob(job *JobInfo) (*HandlerOutput, error) {
 	if !isVideo(job.RequestID, job.SourceFile) {
 		return nil, fmt.Errorf("source was not a video: %s", job.SourceFile)
 	}
-	// Arweave URLs don't support HTTP Range requests and so Mist can't natively handle them for segmenting
+	// Arweave / IPFS URLs don't support HTTP Range requests and so Mist can't natively handle them for segmenting
 	// This workaround copies the file from Arweave to S3 and then tells Mist to use the S3 URL
-	if clients.IsArweaveOrIPFSURL(job.SourceFile) {
+	if clients.IsDStorageResource(job.SourceFile) {
```
victorges, line 46: Right, here we need to do this copying in case of a gateway URL as well. But maybe not all places that use this helper want that.

clients/arweave_ipfs_s3.go
victorges: Hmm, something more sophisticated but more overengineered would be to use the user-provided gateway as the _first_ gateway that we try, but still try the others. Not sure if it's worth the complexity tho, maybe better to just treat as a regular HTTPS URL for the download, but copy it to the bucket due to the gateway HTTP restritions.

victorges (approved):
LGTM! But I think there's a regression with the `IsArweaveOrIPFSURL` helper, details inline

clients/arweave_ipfs_s3.go
red-0ne: Good catch! Will push a solution

clients/arweave_ipfs_s3.go
```diff
@@ -2,52 +2,138 @@ package clients
...
+		if dStorageType == "" {
+			return nil, fmt.Errorf("unsupported dStorage resource %s", dStorageURL.Scheme)
+		}
+
+		gatewayURL, err := url.Parse(gateway)
+		if err != nil {
+			return nil, fmt.Errorf("invalid URL: %w", err)
+		}
+
+		gateways = append(gateways, gatewayURL)
```
victorges, line 63: **nit:** I was confused a bit thinking there would be something before the user-provided gateway. Just a nit to be clearer here:
```suggestion
		gateways = []*url.URL{gatewayURL}
```

clients/arweave_ipfs_s3.go
```diff
@@ -2,52 +2,138 @@ package clients
...
+	if u.Scheme == SCHEME_ARWEAVE || u.Scheme == SCHEME_IPFS {
+		return true
+	}
+	_, _, dStorageType := parseDStorageGatewayURL(u)
+
+	return dStorageType != ""
+}
+
+func parseDStorageGatewayURL(u *url.URL) (string, string, string) {
+	if strings.Contains(u.Host, "arweave") {
```
victorges, line 125: Not a regression, but IMO this should be an exact check as there is only 1 known arweave gateway that this was trying to match AFAIK.
```suggestion
	if u.Host == "arweave.net" {
```

clients/arweave_ipfs_s3.go
```diff
@@ -2,52 +2,138 @@ package clients
...
+		resource := strings.TrimLeft(u.Path, "/")
+		gateway := strings.ReplaceAll(u.String(), resource, "")
+		return resource, gateway, "ar"
```
victorges, line 128: **nit:** WDYT of reusing the scheme constants here and in the places where you check these values? i.e. `SCHEME_ARWEAVE` here and `SCHEME_IPFS` on the caller

clients/arweave_ipfs_s3.go
```diff
@@ -2,52 +2,138 @@ package clients
...
+	}
+
+	if strings.Contains(u.Host, "w3s.link") || strings.Contains(u.Path, "/ipfs/") {
+		parts := strings.Split(u.Path, "/ipfs/")
+		resource := parts[1]
```
victorges, line 133: **nit:** 
```suggestion
		resource := strings.TrimPrefix(u.Path, "/ipfs/")
```

victorges (approved):
LGTM! Only a couple of nits, no need for re-review if you take any of them

PR #379, opened 2023-01-17
move to fixed bitrate ladder for MC and add mp4 output

mjh1:


clients/broadcaster.go
```diff
@@ -64,6 +70,81 @@ type EncodedProfile struct {
...
+	{
+		Name:    "720p0",
+		FPS:     0,
+		Bitrate: 3_000_000,
+		Width:   1280,
+		Height:  720,
+	},
+}
+
+func GetPlaybackProfiles(iv InputVideo) ([]EncodedProfile, error) {
```
mjh1, line 105: I moved these functions to avoid an import cycle, perhaps I should move them to a whole new package though?

clients/mediaconvert.go
```diff
@@ -147,13 +148,36 @@ func (mc *MediaConvert) Transcode(ctx context.Context, args TranscodeJobArgs) er
 // It expects args to be directly compatible with AWS (i.e. S3-only files).
 func (mc *MediaConvert) coreAwsTranscode(ctx context.Context, args TranscodeJobArgs, accelerated bool) error {
 	log.Log(args.RequestID, "Creating AWS MediaConvert job", "input", args.InputFile, "output", args.HLSOutputFile, "accelerated", accelerated)
-	payload := createJobPayload(args.InputFile.String(), args.HLSOutputFile.String(), mc.role, accelerated)
+
+	// hard coded example input for now until we have the real input info
```
mjh1, line 152: @thomshutt are you ok with this being merged or do I need to wait for the ffprobe stuff? Also we want to implement a content duration check too when we can to limit mp4 output to shorter clips?

thomshutt: I think we'll have to hold off because of the length limit, otherwise it's going to get pretty expensive

emranemran: Yep, let's hold off - we might be tweaking the implementation of it from the user perspective (selectable by user vs autogenerate based on length). I have the ffprobe check working locally and attempting to get it working on staging.

PR #406, opened 2023-01-27
fix duplicate profile

mjh1:
Previously the `lowerQualityThanSrc` condition would match if the heights were equal, which would result in a duplicate profile because we are then adding a profile with height equal to the input [lower down in the function](https://github.com/livepeer/catalyst-api/compare/mh/fix-duplicate-profile?expand=1#diff-7bbc069c84bd3dadf5c408f6c62b025837d72035cd18c8cea4b70a4d12efd1c1R104)

E.g. you'd end up with a set of profiles like:
```
				{Name: "360p0", Width: 640, Height: 360, Bitrate: 800_000},
				{Name: "720p0", Width: 1280, Height: 720, Bitrate: 3_000_000},
				{Name: "720p0", Width: 1280, Height: 720, Bitrate: 3_000_001},
```
MediaConvert errors on this because the output paths clash (due to the duplicate profile name).

### DIFF
--- a/video/profiles.go
+++ b/video/profiles.go
@@ -93,7 +93,7 @@ func GetPlaybackProfiles(iv InputVideo) ([]EncodedProfile, error) {
 	for _, profile := range DefaultTranscodeProfiles {
 		// transcoding job will adjust the width to match aspect ratio. no need to
 		// check it here.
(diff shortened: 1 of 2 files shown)


mjh1:
@thomshutt just want to check the previous behaviour wasn't intentional, we didn't really want to have two renditions with same dimensions but differing bitrates?

thomshutt (approved):

PR #408, opened 2023-01-27
Bump maximum bitrate for default profiles

thomshutt:
Trying this out to see if we can improve our Livepeer pipeline transcodes


thomshutt:
@emranemran I'm going to merge this since it's a small change, but would appreciate a retroactive review

Issue #389, opened 2023-01-24
verify MP4 renditions on S3 before signaling Studio


emranemran:
When the Mist/Mediaconvert transcode job is completed, catalyst-api should check that the MP4 output files are generated before responding to Studio. The Studio logic can then mark these files as "existing" so that SDK knows which renditions are available.

mjh1:
@emranemran should this just be an http HEAD request on the URLs and check for a 2xx?

mjh1:
In this ticket I'll also include implementing fixing [profiles_count field](https://github.com/livepeer/catalyst-api/blob/6c26531a761d5fa4003a8f056df19d3c8a98cb04/pipeline/coordinator.go#L392) for the postgres metrics to be the number of output profiles rather than requested profiles.

Issue #400, opened 2023-01-26
catalyst <-> studio integration for short form videos


gioelecerati:
After a sync with @emranemran , some work is needed to be done on Studio Task Runner and Catalyst for short form videos, specifically:

- add `AutoMP4s: bool` in `api/vod` in both the request studio & the catalyst-api struct
- Add `Mp4Outputs` to TranscodeStatusMessage from Catalyst to Studio
- Potentially, before populating the struct of TranscodeStatusMessage check on the storage if the mp4 is there (may require more time)
- `api/transcode` endpoint need to recognize the AutoMP4s

emranemran:
Adding some context for @mjh1:

> add AutoMP4s: bool in api/vod in both the request studio & the catalyst-api struct
- If this flag is set, then we should enable MP4 output regardless of duration threshold (currently set to 2min)

> Add Mp4Outputs to TranscodeStatusMessage from Catalyst to Studio

- When a job is done, we call: 
`func (c *Coordinator) finishJob(job *JobInfo, out *HandlerOutput, err error) {`
which calls: 
`        job.statusClient.SendTranscodeStatus(tsm)`.
We need to update `TranscodeStatusMessage` and specifically `OutputVideoFile` to include an `MP4OutputLocation` (instead of `MP4Outputs` field that indicates an MP4 was generated for that specific rendition. Since we assume mediaconvert will succeed, perhaps for now we just can auto-generate the paths based on what we know in catalyst-api. A follow-up ticket can be used to verify that the mp4s were successfully placed in the object-store. @mjh1 mind filing a separate ticket if this seems to be a requirement when you make these changes? 

> Potentially, before populating the struct of TranscodeStatusMessage check on the storage if the mp4 is there (may require more time)
- same comment as previous

> api/transcode endpoint need to recognize the AutoMP4s
- Same as first point -- this endpoint should support the same `AutoMP4` field in the request and also update the same job structs that will get passed to the transcode functions.

victorges:
@gioelecerati @emranemran So is the idea to add a new `output_location` on the upload request payload, or to just toggle `AutoMP4s` on the existing `output_location` that we already send? (which is only for HLS today and sends an `index.m3u8` URL for the path)

I think the first (adding a new `output_location`) would need some logic on catalyst side to make sure outputs are generated in the exact same order as the requested `output_locations`, or to figure out some other way to match requests with outputs as currently we only really have the order (so `task-runner` is implemented like that).

[...]

PR #411, opened 2023-01-31
add mp4 outputs to status message

mjh1:
fixes:
https://github.com/livepeer/catalyst-api/issues/389
https://github.com/livepeer/catalyst-api/issues/400


thomshutt (approved):
Looks great, cc @victorges since you'll be the consumer here

PR #413, opened 2023-02-01
Web3.storage

leszko:


clients/mediaconvert.go
```diff
@@ -441,28 +441,11 @@ func copyFile(ctx context.Context, sourceURL, destOSBaseURL, filename, requestID
 
 	content := io.TeeReader(c, &writtenBytes)
 
-	err = UploadToOSURL(destOSBaseURL, filename, content, 1*time.Minute)
+	err = UploadToOSURL(destOSBaseURL, filename, content, 5*time.Minute)
 	if err != nil {
 		return writtenBytes.count, fmt.Errorf("upload error: %w", err)
 	}
 
-	storageDriver, err := drivers.ParseOSURL(destOSBaseURL, true)
```
leszko, line 449: Is there any reason why we would need this logic? It looks redundant to always read the data after it's completed, that's why I removed it.

Nevertheless, let me know if we need it, then I'll figure out how to make it working for web3.storage.

clients/mediaconvert.go
```diff
@@ -537,7 +520,7 @@ func trimBaseDir(osPath, filePath string) string {
 			return filePath
 		}
 	}
-	return strings.TrimPrefix(filePath, baseDir)
+	return strings.TrimLeft(strings.TrimPrefix(filePath, baseDir), "/")
```
leszko, line 523: We need to trim here, otherwise, we're left with files like `/1.ts`.

transcode/manifest.go
```diff
@@ -143,12 +140,17 @@ func GenerateAndUploadManifests(sourceManifest m3u8.MediaPlaylist, targetOSURL s
...
 	}
 
-	err := clients.UploadToOSURL(targetOSURL, MASTER_MANIFEST_FILENAME, strings.NewReader(masterPlaylist.String()), 30*time.Second)
+	err := clients.UploadToOSURL(targetOSURL, MASTER_MANIFEST_FILENAME, strings.NewReader(masterPlaylist.String()), UPLOAD_TIMEOUT)
 	if err != nil {
 		return "", fmt.Errorf("failed to upload master playlist: %s", err)
 	}
 
-	return filepath.Join(targetOSURL, MASTER_MANIFEST_FILENAME), nil
+	res, err := url.JoinPath(targetOSURL, MASTER_MANIFEST_FILENAME)
```
leszko, line 148: This a fix of a bug, before we returned an url in the form of `http:/some-url`, now it's correct `http://some-url` (the `/` was missing).

transcode/transcode.go
```diff
@@ -54,20 +60,23 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
...
-	// Parse the manifest destination of the segmented output specified in the request
-	segmentedOutputManifestURL, err := url.Parse(transcodeRequest.SourceManifestURL)
-	if err != nil {
-		return outputs, segmentsCount, fmt.Errorf("failed to parse transcodeRequest.UploadURL: %s", err)
+	var targetURL *url.URL
+	var err error
+	if transcodeRequest.TargetURL != "" {
+		targetURL, err = url.Parse(transcodeRequest.TargetURL)
+	} else {
+		// TargetURL not defined in the request, fall back the source manifest dir
```
leszko, line 68: This is actually for the backwards-compatibility of the `/api/transcode/file` endpoint. For the `/api/vod`, `TargetURL` is always there.

clients/mediaconvert.go
thomshutt: @mjh1?

transcode/transcode.go
```diff
@@ -120,20 +129,39 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
...
+		return ""
+	}
+	videoUrl, err := osDriver.Publish(context.Background())
+	if err == drivers.ErrNotSupported {
+		return osUrl
+	} else if err == nil {
+		res, _ := url.JoinPath(videoUrl, relPath)
+		return res
+	}
+	return ""
```
thomshutt, line 162: What does this default case (i.e we get some error but not the error about Publish being unsupported) mean here? Could you add an explanatory comment to the method please (also to explain about when Publish is / isn't supported for people unfamiliar for the w3s work)?

handlers/upload.go
```diff
@@ -103,16 +103,18 @@ func (d *CatalystAPIHandlersCollection) handleUploadVOD(w http.ResponseWriter, r
 	var requestID = config.RandomTrailer(8)
 	log.AddContext(requestID, "source", uploadVODRequest.Url)
 
-	// find source segment URL
+	// find output storage URLs for source and target segments
+	var useTargetForSourceOutput bool
 	var tURL string
 	for _, o := range uploadVODRequest.OutputLocations {
+		tURL = o.URL
```
thomshutt, line 110: I've been struggling to parse this logic a bit in my head, so it could be good to split it out into a method on the VOD request like `GetSourceOutputURL()` that we can unit test. In particular, I'm trying to understand if the ordering of OutputLocations now affects which one gets used for the target / what happens in the case where no "source segments" target is set

thomshutt (approved):
Few minor comments but overall LGTM, fits in with the non-dStorage path a lot more nicely that I was expecting!

handlers/upload.go
leszko: Updated + added unit tests for both `getSourceOutputURL()` and `getTargetURL()`.

transcode/transcode.go
leszko: Added a comment.

clients/mediaconvert.go
thomshutt: @red-0ne any chance you could take a look here?

red-0ne: Yes, this can be safely removed. This is legacy code I missed removing after introducing the tee reader for counting file size.

handlers/upload.go
```diff
@@ -64,6 +69,22 @@ func HasContentType(r *http.Request, mimetype string) bool {
...
+		}
+	}
+	return nil, nil
+}
+
+func (r UploadVODRequest) getTargetURL() (*url.URL, error) {
+	if len(r.OutputLocations) < 1 {
+		return nil, errors2.New("no output locations")
+	}
+	return url.Parse(r.OutputLocations[0].URL)
```
victorges, line 85: Should this be a similar logic looking for the `output` with `TranscodedSegments: true`?

handlers/upload.go
```diff
@@ -103,24 +124,23 @@ func (d *CatalystAPIHandlersCollection) handleUploadVOD(w http.ResponseWriter, r
...
+	targetURL, err := uploadVODRequest.getTargetURL()
+	if err != nil {
+		return false, errors.WriteHTTPBadRequest(w, "Invalid request payload", err)
 	}
-	if tURL == "" {
-		return false, errors.WriteHTTPBadRequest(w, "Invalid request payload", fmt.Errorf("no source segment URL in request"))
+	// Hack for web3.storage to distinguish different jobs, before calling Publish()
+	// Can be removed after we address this issue: https://github.com/livepeer/go-tools/issues/16
+	if targetURL.Scheme == "w3s" {
+		targetURL.Host = requestID
```
victorges, line 136: Could this be sth like `log.AddContext(requestID, "w3s-url", targetURL.String())`?

transcode/transcode.go
```diff
@@ -120,20 +129,43 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
...
 	outputs = []clients.OutputVideo{output}
 	// Return outputs for .dtsh file creation
 	return outputs, segmentsCount, nil
 }
 
+// PublishDriverSession tries to publish the given osUrl and returns a publicly accessible video URL.
+// If driver supports `Publish()`, e.g. web3.storage, then return the path to the video.
+// If driver does not support `Publish()`, e.g. S3, then return the input osUrl, video should be accessible with osUrl.
+// In case of any other error, return an empty string; we don't want to propagate osUrl by default, because it may contain sensitive data.
+func PublishDriverSession(osUrl string, relPath string) string {
```
victorges, line 154: Could we obfuscate the errors but still return an error for this function? It's weird even from the usage side for it to just manipulate some strings but is actually performing some I/O underneath that can fail. For example would we want to consider the job successful if we tried doing this publish but the call to web3.storage fails? I think we'd like to retry the job in that case as well (from here we might just need to return/propagate "an" error)

transcode/transcode.go
victorges: Should we check that explicitly in the code, or perhaps replace the comment with that? I found your comment on github more useful than the comment in the code!

transcode/transcode.go
```diff
@@ -120,20 +129,43 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
...
+	osDriver, err := drivers.ParseOSURL(osUrl, true)
+	if err != nil {
+		return ""
+	}
+	videoUrl, err := osDriver.Publish(context.Background())
```
victorges, line 159: Ohh, I see now why we couldn't make the publish process a thing from the OS sessions, not global... Hope we can refactor that soon to avoid leaking data! Either between tasks and unfinished leftovers etc.

victorges (approved):
LGTM in general

transcode/transcode.go
leszko: Updated the comment.

handlers/upload.go
leszko: Yeah, I think you're right. I can change it that way.

leszko: With the only difference is that we currently don't accept requests without `TranscodedSegments: true`, so it'll return an error.

transcode/transcode.go
leszko: Yes, it's a good point. Let me change it.

transcode/transcode.go
leszko: Yes, completely agree. We can do it as a separate activity. Tracked here: https://github.com/livepeer/go-tools/issues/16

handlers/upload.go
leszko: Good idea, added.

PR #414, opened 2023-02-01
remove logic to let MC try dstorage urls

mjh1:
if copying from dstorage failed we were then attempting to get a presigned s3 url from a non existent object

### DIFF
--- a/clients/callback_client_status.go
+++ b/clients/callback_client_status.go
@@ -97,7 +97,7 @@ type OutputVideoFile struct {
 
 type OutputVideo struct {
 	Type     string            `json:"type"`
-	Manifest string            `json:"manifest"`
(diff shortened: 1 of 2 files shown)


clients/callback_client_status.go
```diff
@@ -97,7 +97,7 @@ type OutputVideoFile struct {
 
 type OutputVideo struct {
 	Type     string            `json:"type"`
-	Manifest string            `json:"manifest"`
+	Manifest string            `json:"manifest,omitempty"`
```
mjh1, line 100: as suggested by Gioele to omit the blank manifest field for mp4 outs

thomshutt (approved):

emranemran (approved):

PR #436, opened 2023-02-07
Remove ucan flag

leszko:
The env variable `W3_PRINCIPAL_KEY` will be used instead.


thomshutt:
What does a failure look like when we're trying to do Web3.storage and this isn't set? 

i.e do we need some sort of checking for the env var to provide an early / clear error message?

leszko:
> What does a failure look like when we're trying to do Web3.storage and this isn't set?
> 
> i.e do we need some sort of checking for the env var to provide an early / clear error message?

Yes, the check is here: https://github.com/livepeer/go-tools/blob/main/drivers/drivers.go#L280

It's not as early as starting catalyst-api, but I think it's early enough, it'll fail the first time you try to create a driver. We could add an additional check in catalyst-api, but I'm not sure we want to introduce that coupling, just to check if the env variable is there.

thomshutt (approved):
Looks great, thanks for confirming!

Issue #418, opened 2023-02-02
update mp4 output logic to respect the automp4 flag from studio


emranemran:
From Victor: 
``` 
I don’t think we should generate MP4s in the new Transcode API. The default should be turned off. As this is a playback optimization feature right now, not a transcode/output option, we shouldn’t generate these additional files in the user-provided object-store. That seems like an annoying behavior (e.g. I’m requesting for HLS now I’m getting a bunch of MP4s I don’t want consuming my storage).
```

PR #441, opened 2023-02-08
introduce auto_mp4 request flag

mjh1:
only generate mp4s if auto_mp4 request flag is set and input duration is less than 2mins

Fixes https://github.com/livepeer/catalyst-api/issues/418

### DIFF
--- a/handlers/upload.go
+++ b/handlers/upload.go
@@ -25,6 +25,7 @@ type UploadVODRequestOutputLocationOutputs struct {
 	SourceMp4          bool `json:"source_mp4"`
 	SourceSegments     bool `json:"source_segments"`
 	TranscodedSegments bool `json:"transcoded_segments"`
+	AutoMP4            bool `json:"auto_mp4"`
 }
 
 type UploadVODRequestOutputLocation struct {
@@ -84,6 +85,15 @@ func (r UploadVODRequest) getTargetURL() (*url.URL, error) {
 		}
 	}
 	return nil, fmt.Errorf("no output_location with transcoded_segments")
+}
+
+func (r UploadVODRequest) isAutoMP4() bool {
+	for _, o := range r.OutputLocations {
+		if o.Outputs.AutoMP4 {
+			return true
+		}
+	}
+	return false
 }
 
 func (d *CatalystAPIHandlersCollection) UploadVOD() httprouter.Handle {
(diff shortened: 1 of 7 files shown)


emranemran:
Changes LGTM. @victorges and @gioelecerati: mind reviewing when you get a chance? We'll also hold off on the corresponding studio changes before merging.

gioelecerati (approved):
LGTM, 
this is the commit on the task-runner side if you want to take a look https://github.com/livepeer/task-runner/pull/132/commits/bef71b859355ad30469cffefd2490f773081bce6

handlers/upload.go
```diff
@@ -86,6 +87,15 @@ func (r UploadVODRequest) getTargetURL() (*url.URL, error) {
 	return nil, fmt.Errorf("no output_location with transcoded_segments")
 }
 
+func (r UploadVODRequest) isAutoMP4() bool {
+	for _, o := range r.OutputLocations {
+		if o.Outputs.AutoMP4 {
+			return true
```
victorges, line 93: WDYT of returning the location of the `URL` of the MP4 location instead and saving it on the args object? Otherwise we're completely ignoring it here and assuming it has to be the same location as the HLS playlist etc.

If it requires too much changes in the core logic (like splitting the logic to copy the files etc which might be cumbersome), another idea would be returning this "is auto mp4" boolean from the `getTargetURL` function instead. So we explicitly only support the same output location for both locations, and just return whether the "target url" location has `auto_mp4` turned on.

victorges (approved):
LGTM either way

handlers/upload.go
mjh1: yeah i was considering doing the latter tbh, since we do currently only support a single output location. i'll go with that

handlers/upload.go
```diff
@@ -77,13 +78,13 @@ func (r UploadVODRequest) getSourceOutputURL() (*url.URL, error) {
 	return nil, nil
 }
 
-func (r UploadVODRequest) getTargetURL() (*url.URL, error) {
+func (r UploadVODRequest) getTargetOutput() (UploadVODRequestOutputLocation, error) {
 	for _, o := range r.OutputLocations {
 		if o.Outputs.TranscodedSegments {
-			return url.Parse(o.URL)
+			return o, nil
```
victorges, line 84: Nice! Me gusta :)

victorges (approved):
LGTM

PR #449, opened 2023-02-10
add retries to probing

mjh1:


thomshutt (approved):
Unit test would be good, but fine with having it as a separate PR

PR #451, opened 2023-02-10
probe: add check for jpeg/mpeg type of inputs

emranemran:
Any single jpeg input file may get classified as an mjpeg in the probing logic. These inputs are not supported currently. This change adds a check for mjpeg type of inputs and fails the probe check.

### DIFF
--- a/video/probe.go
+++ b/video/probe.go
@@ -16,8 +16,7 @@ type Prober interface {
 	ProbeFile(url string) (InputVideo, error)
 }
 
-type Probe struct {
-}
+type Probe struct{}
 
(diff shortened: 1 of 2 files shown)


thomshutt:
@emranemran Looks good, I also added some unit tests (and from running them noticed that we'd also return the mjpeg error if the source video codec string was empty, so changed it to an equality rather than `Contains` check)

thomshutt (approved):

emranemran:
> @emranemran Looks good, I also added some unit tests (and from running them noticed that we'd also return the mjpeg error if the source video codec string was empty, so changed it to an equality rather than `Contains` check)

Got it. Nice catch.

PR #457, opened 2023-02-13
round to an even height and width to fix MC errors

mjh1:
Mediaconvert only accepts even values for width and height so we have to round these values.

We were seeing the following errors in the logs:
`Invalid resolution [400 x 239], only even values are supported. video_description [1]`

### DIFF
--- a/video/profiles.go
+++ b/video/profiles.go
@@ -93,11 +93,11 @@ func GetPlaybackProfiles(iv InputVideo) ([]EncodedProfile, error) {
 		profiles = []EncodedProfile{lowBitrateProfile(video)}
 	}
 	profiles = append(profiles, EncodedProfile{
-		Name:    strconv.FormatInt(video.Height, 10) + "p0",
+		Name:    strconv.FormatInt(nearestEven(video.Height), 10) + "p0",
 		Bitrate: video.Bitrate,
 		FPS:     0,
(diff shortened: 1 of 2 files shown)


thomshutt:
Could you add a description to the ticket about why odd numbers are a problem for Mediaconvert please?

video/profiles.go
```diff
@@ -93,11 +93,11 @@ func GetPlaybackProfiles(iv InputVideo) ([]EncodedProfile, error) {
...
 	}
 	profiles = append(profiles, EncodedProfile{
-		Name:    strconv.FormatInt(video.Height, 10) + "p0",
+		Name:    strconv.FormatInt(nearestEven(video.Height), 10) + "p0",
 		Bitrate: video.Bitrate,
 		FPS:     0,
-		Width:   video.Width,
-		Height:  video.Height,
+		Width:   nearestEven(video.Width),
+		Height:  nearestEven(video.Height),
```
emranemran, line 100: Does rounding cause any issues with the Mist/Livepeer pipeline?

thomshutt: I think we only pass one of width / height in there and it scales on the transcoder so I don't think this should be an issue

mjh1: oh yeah kinda forgot that, we only pass height to MC too

PR #458, opened 2023-02-13
extract s3 upload and probe into a function

mjh1:
Moving the logic which copies the input to s3 and probes it into a separate file. We'll then be able to use this for the mist pipeline too.

### DIFF
--- a/clients/upload.go
+++ b/clients/upload.go
@@ -1,0 +1,64 @@
+package clients
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/url"
+
+	"github.com/livepeer/catalyst-api/log"
+	"github.com/livepeer/catalyst-api/video"
+)
+
+type SourceUpload struct {
+	S3    S3
+	Probe video.Prober
+}
+
+func (s *SourceUpload) Upload(args TranscodeJobArgs, s3HTTPTransferURL *url.URL) (TranscodeJobArgs, error) {
+	if s3HTTPTransferURL == nil {
+		return TranscodeJobArgs{}, errors.New("s3HTTPTransferURL was nil")
+	}
+	s3URL, err := url.Parse("s3://" + s3HTTPTransferURL.Path)
+	if err != nil {
+		return TranscodeJobArgs{}, fmt.Errorf("failed to parse s3 url: %w", err)
+	}
+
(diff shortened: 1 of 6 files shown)


clients/upload.go
```diff
@@ -0,0 +1,64 @@
+package clients
```
emranemran, line 1: Would suggest calling this file something else like `filecopy.go` so that it's not clashing with `handlers/upload.go`. It would be a little less confusing this way.

clients/upload.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/livepeer/catalyst-api/log"
+	"github.com/livepeer/catalyst-api/video"
+)
+
+type SourceUpload struct {
+	S3    S3
+	Probe video.Prober
+}
+
+func (s *SourceUpload) Upload(args TranscodeJobArgs, s3HTTPTransferURL *url.URL) (TranscodeJobArgs, error) {
```
emranemran, line 18: Can we call this something more explicit like `CopySrcToDst` or something to that effect so that it doesn't clash with other functions like `UploadOSURL` etc?

emranemran (changes requested):

clients/upload.go
mjh1: Ah good point, will do

clients/upload.go
mjh1: makes sense 👍

log/logger.go
```diff
@@ -82,7 +88,7 @@ func redactKeyvals(keyvals ...interface{}) []interface{} {
 
 func RedactURL(str string) string {
 	strLower := strings.ToLower(str)
-	if !strings.HasPrefix(strLower, "s3+http") && !strings.HasPrefix(strLower, "http") {
+	if !strings.HasPrefix(strLower, "s3+http") && !strings.HasPrefix(strLower, "http") && !strings.HasPrefix(strLower, "s3") {
```
thomshutt, line 91: Can we get rid of the first clause now you've added the more general one?

thomshutt (approved):

log/logger.go
mjh1: oh yeah ha :)

emranemran (approved):

PR #460, opened 2023-02-13
Increase retries and timeouts for saving data

leszko:


clients/mediaconvert.go
```diff
@@ -481,7 +481,7 @@ func copyFile(ctx context.Context, sourceURL, destOSBaseURL, filename, requestID
 
 	content := io.TeeReader(c, &writtenBytes)
 
-	err = UploadToOSURL(destOSBaseURL, filename, content, 5*time.Minute)
+	err = UploadToOSURL(destOSBaseURL, filename, content, 30*time.Minute)
```
leszko, line 484: We need this one for mp4 files, not only for web3.strorage since mp4 can be very big. In the long run,we should make the timeout dependent on the file size.

emranemran: Will this cause issues with Studio timeouts? 
https://github.com/livepeer/studio/blob/23b1aa453bbef83981ad2547169c7149f0436e35/packages/api/src/store/task-table.ts#L7-L9
If we're stuck in `UploadToOSURL`, is our callback to Studio `ReportProgress` called concurrently? 
cc @victorges

clients/object_store_client.go
```diff
@@ -115,14 +115,25 @@ func PublishDriverSession(osUrl string, relPath string) (string, error) {
...
-	videoUrl, err := osDriver.Publish(context.Background())
-	if err == drivers.ErrNotSupported {
-		// driver does not support Publish(), video will be accessible with osUrl
-		return osUrl, nil
-	} else if err != nil {
-		// error while publishing the video
-		return "", err
+
+	var videoUrl string
+	err = backoff.Retry(func() error {
```
emranemran, line 120: Should this exponential retry logic also be copied to `getFileHTTP`?

emranemran (approved):

clients/object_store_client.go
leszko: I think that `getFileHTTP` has its own retry logic. But I'll update its parameters to be the same as here.

clients/mediaconvert.go
leszko: That is a very good point! I'll move this change into a separate PR and we can discuss it there (since it's not strictly related to saving data, but more to bigger files => mp4)

PR #463, opened 2023-02-15
Implement unified input copying and probing

mjh1:
This PR covers moving the input file copying and probing from the mediaconvert pipeline to the Coordinator so that we have access to the probe data ahead of any of the pipeline specific code. The object store copy of the input file can then be used as the input for mist as well as mediaconvert.

The bulk of the change lives in `pipeline/coordinator.go`, along with some refactoring in `clients/input_copy.go`. There are then a fair amount of changes to move the copy & probe related tests from `clients/mediaconvert_test.go` to `pipeline/coordinator_test.go`


pipeline/coordinator.go
```diff
@@ -320,6 +369,21 @@ func (c *Coordinator) runHandlerAsync(job *JobInfo, handler func() (*HandlerOutp
 	})
 }
 
+func (c *Coordinator) runHandlerAsync2(job *JobInfo, handler func() (*HandlerOutput, error)) {
```
mjh1, line 372: (needs a better name) This is the similar to `runHandlerAsync()` but only calls `finishJob()` on errors to notify studio of any errors in the copying & probing stage.

main.go
```diff
@@ -28,6 +28,7 @@ func main() {
 	promPort := flag.Int("prom-port", 2112, "Prometheus metrics port")
 	pprofPort := flag.Int("pprof-port", 6061, "Pprof listen port")
 	sourceOutputUrl := flag.String("source-output", "", "URL for the video source segments used if source_segments is not defined in the upload request")
+	transferUrl := flag.String("transfer-location", "", "URL to store input files temporarily while processing")
```
emranemran: so is this the s3 url for mediaconvert and gcs url for mist?

pipeline/coordinator.go
```diff
@@ -254,18 +302,71 @@ func (c *Coordinator) startOneUploadJob(p UploadJobPayload, handler Handler, for
...
+			s3URL := &url.URL{
+				Scheme: c.TransferURL.Scheme,
+				Host:   c.TransferURL.Host,
+				Path:   path.Join(c.TransferURL.Path, mcInputRelPath),
+			}
+			// same thing is done in the constructor in mediaconvert.go in terms of constructing the url with region as host
+			osTransferURL := &url.URL{
+				Scheme: c.TransferURL.Scheme,
+				User:   c.TransferURL.User,
+				Host:   "us-east-1", // weird but compatible with drivers.ParseOSURL
```
emranemran: is it correct to always use this host?

pipeline/coordinator.go
```diff
@@ -254,18 +302,71 @@ func (c *Coordinator) startOneUploadJob(p UploadJobPayload, handler Handler, for
...
 	si.ReportProgress(clients.TranscodeStatusPreparing, 0)
 
 	c.Jobs.Store(streamName, si)
 	log.Log(si.RequestID, "Wrote to jobs cache")
 
 	c.runHandlerAsync(si, func() (*HandlerOutput, error) {
+		if c.TransferURL == nil {
+			log.Log(si.RequestID, "no unified transfer")
+		} else {
+			targetDir := getTargetDir(si.RequestID, si.TargetURL)
```
emranemran: I think all the logic from here on in this `else` case would be nicer to move away to a `CopyAndProbe` fcn?

pipeline/mist.go
```diff
@@ -64,29 +68,22 @@ func (m *mist) HandleStartUploadJob(job *JobInfo) (*HandlerOutput, error) {
...
-	// This workaround copies the file from Arweave to S3 and then tells Mist to use the S3 URL
-	if clients.IsDStorageResource(job.SourceFile) {
-		if !isVideo(job.RequestID, job.SourceFile) {
-			return nil, fmt.Errorf("source was not a video: %s", job.SourceFile)
-		}
-		sourceURL, err := url.Parse(job.SourceFile)
-		if err != nil {
-			return nil, fmt.Errorf("failed to parse source as URL: %w", err)
-		}
+	sourceURL, err := url.Parse(job.SourceFile)
```
emranemran: For the mist case, is it able to handle a pre-signed URL from s3? I can't recall where we landed on that discussion.

pipeline/coordinator.go
```diff
@@ -197,6 +199,62 @@ func NewStubCoordinatorOpts(strategy Strategy, statusClient clients.TranscodeSta
 // This has the main logic regarding the pipeline strategy. It starts jobs and
 // handles processing the response and triggering a fallback if appropriate.
 func (c *Coordinator) StartUploadJob(p UploadJobPayload) {
+	si := &JobInfo{
+		UploadJobPayload: p,
+		//StreamName:       streamName,
```
emranemran, line 203: delete?

pipeline/coordinator.go
```diff
@@ -197,6 +199,62 @@ func NewStubCoordinatorOpts(strategy Strategy, statusClient clients.TranscodeSta
...
+		statusClient: c.statusClient,
+		startTime:    time.Now(),
+		//result:           make(chan bool, 1),
```
emranemran, line 206: delete?

emranemran (changes requested):

emranemran:
Left a few questions and have a few generic comments: 
- could we squash the first 2 commits? It was a getting a bit confusing to go commit by commit and have some new code deleted in the second commit. Would love to re-review after a squash so make sure I understand the new logic. 
- Mind adding a bit more explanation to the commit descriptions? A summary of the changes and reasons 'why' we need the change would add a lot of context for future changes as this PR is quite large now. 
- What kind of testing was done for the Mist pipeline? 
- What's the plan to gain confidence in staging with these changes? I think we need some soak time on staging with different file types and ensure we don't push to prod until we validate. WDYT?

pipeline/mist.go
mjh1: That's a good point, thanks for the reminder. I have tested and it is working for mist, but i need to double check I understand fully.

mjh1:
> Left a few questions and have a few generic comments:
> 
> * could we squash the first 2 commits? It was a getting a bit confusing to go commit by commit and have some new code deleted in the second commit. Would love to re-review after a squash so make sure I understand the new logic.
> * Mind adding a bit more explanation to the commit descriptions? A summary of the changes and reasons 'why' we need the change would add a lot of context for future changes as this PR is quite large now.
> * What kind of testing was done for the Mist pipeline?
> * What's the plan to gain confidence in staging with these changes? I think we need some soak time on staging with different file types and ensure we don't push to prod until we validate. WDYT?

Ah sorry yes the first commit was a bit of a throwaway first attempt, pointless to keep now so I'll squash it.

For mist (and also external) it was just a simple transcode test for a 30 second video.

That's right we will need a fair amount of testing on staging, Thom mentioned he has a script from before to test a bunch of different video types.

pipeline/mist.go
mjh1: Ok i've ensured that we now send mist the same type of URL we used to (rather than the presigned one). e.g. what we do already here: https://github.com/livepeer/catalyst-api/blob/2d89524e60a33e95453e2583d09e5e0efe37c513/pipeline/mist.go#L87

clients/input_copy.go
```diff
@@ -20,59 +20,80 @@ const MAX_COPY_FILE_DURATION = 30 * time.Minute
...
-	presignedInputFileURL, err := s.S3.PresignS3(s3URL.Host, s3URL.Path)
+	signedURL, err = signURL(osTransferURL)
 	if err != nil {
-		return TranscodeJobArgs{}, fmt.Errorf("error creating s3 url: %w", err)
+		return
 	}
 
-	log.Log(args.RequestID, "starting probe", "s3url", s3URL)
-	inputVideoProbe, err := s.Probe.ProbeFile(presignedInputFileURL)
+	log.Log(requestID, "starting probe", "source", inputFile.String(), "dest", osTransferURL.String(), "presigned", signedURL)
```
emranemran, line 56: This doesn't leak S3 creds right?

clients/s3.go
```diff
@@ -12,19 +12,19 @@ type S3 interface {
...
 }
 
 func (c *S3Client) PresignS3(bucket, key string) (string, error) {
-	req, _ := c.s3.GetObjectRequest(&s3.GetObjectInput{
+	req, _ := c.S3.GetObjectRequest(&s3.GetObjectInput{
 		Bucket: &bucket,
 		Key:    &key,
 	})
-	return req.Presign(5 * time.Minute)
+	return req.Presign(60 * time.Minute)
```
emranemran, line 23: Why the increase to 60? Also, maybe better to define as a const somewhere common since I see the 60min reference in other places in the code.

pipeline/coordinator.go
```diff
@@ -188,6 +199,46 @@ func NewStubCoordinatorOpts(strategy Strategy, statusClient clients.TranscodeSta
 // This has the main logic regarding the pipeline strategy. It starts jobs and
 // handles processing the response and triggering a fallback if appropriate.
 func (c *Coordinator) StartUploadJob(p UploadJobPayload) {
+	// A bit hacky - this is effectively a dummy job object to allow us to reuse the runHandlerAsync and
```
emranemran, line 202: I'm not sure I fully understand this here - mind elaborating why it's hacky? Also, if the InputCopyToS3 fails for some reason, does it get reflected in the async handler correctly? Maybe we should trigger this condition in a test to see what happens?

pipeline/mist.go
```diff
@@ -64,30 +62,6 @@ func (m *mist) HandleStartUploadJob(job *JobInfo) (*HandlerOutput, error) {
...
-			return nil, fmt.Errorf("failed to parse source as URL: %w", err)
-		}
-
-		newSourceURL, err := inSameDirectory(*sourceOutputUrl, MIST_SEGMENTING_SUBDIR, path.Base(sourceURL.Path))
-		if err != nil {
-			return nil, fmt.Errorf("cannot create location for source copy: %w", err)
-		}
-		log.AddContext(job.RequestID, "new_source_url", newSourceURL.String())
-
-		if err := clients.CopyDStorageToS3(job.SourceFile, newSourceURL.String(), job.RequestID); err != nil {
```
emranemran, line 84: So where does the DStorage copy logic live now?

emranemran:
Left a couple more questions. Also, I think this PR could use a test plan outlined here so that we can all brainstorm and be on the same page on what specific things will need to be tested on staging via manual tests vs w/ Studio in the loop.

pipeline/mist.go
mjh1: Since we're using the unified copy now it gets taken care of there, `input_copy.go` handles reading from dstorage https://github.com/livepeer/catalyst-api/pull/463/files#diff-d93900b04efde86efa555721e97460e1caf9b798bbdae4a68a1f4105c0500160R227

pipeline/coordinator.go
mjh1: The tests I added to coordinator_test.go cover some error scenarios e.g. https://github.com/livepeer/catalyst-api/pull/463/files#diff-d93900b04efde86efa555721e97460e1caf9b798bbdae4a68a1f4105c0500160R227
I'll check if there are any other cases that could be added

clients/s3.go
mjh1: I think I changed that when I thought we would be sending the presigned urls to mist and didn't know how long it would take mist to process. But since that's not happening I'll reduce it, I think 10 mins makes sense in case there are multiple retries for the probing etc.

pipeline/coordinator.go
```diff
@@ -188,6 +199,46 @@ func NewStubCoordinatorOpts(strategy Strategy, statusClient clients.TranscodeSta
...
+		statusClient:     c.statusClient,
+		startTime:        time.Now(),
+
+		numProfiles:    len(p.Profiles),
+		state:          "segmenting",
+		catalystRegion: os.Getenv("MY_REGION"),
+	}
+	si.ReportProgress(clients.TranscodeStatusPreparing, 0)
+
+	c.runHandlerAsync(si, func() (*HandlerOutput, error) {
```
thomshutt, line 215: Why are we using runHandlerAsync here rather than a standard `go func(){}`?

mjh1: So that we can re-use the logic to report errors to studio via the callback handler.

thomshutt: Ah yep, gotcha

mjh1: @victorges would you mind sanity checking my usage of this function as I believe you wrote this stuff originally? It's tested in staging successful and failed ingests and I can see the callbacks are working but are there any edge cases you think I should check?

emranemran (approved):

clients/mediaconvert.go
```diff
@@ -27,10 +27,9 @@ var pollDelay = 10 * time.Second
 var retryableHttpClient = newRetryableHttpClient()
 
 const (
-	rateLimitedPollDelay  = 15 * time.Second
-	maxMP4OutDuration     = 2 * time.Minute
-	mp4OutFilePrefix      = "static"
-	maxInputFileSizeBytes = 10 * 1024 * 1024 * 1024 // 10 GiB
+	rateLimitedPollDelay = 15 * time.Second
+	maxMP4OutDuration    = 2 * time.Minute
```
emranemran, line 31: Should this be defined elsewhere and not just in clients/mediaconvert.go since we'll use it for livepeer pipeline as well?

mjh1: will do in a follow up PR

mjh1: ok thanks

PR #467, opened 2023-02-16
Move copy functions to input_copy

mjh1:
Moving copy functions to `input_copy.go` as they'll now be used for the mist pipeline too.


thomshutt (approved):
LGTM, could you add a description about why you're moving this please

PR #470, opened 2023-02-17
Add a pprof listener for tracing

mjh1:
We're having some errors copying output files to storage so adding a pprof listener to be able to retrieve profiling/tracing data from the pods to investigate.

### DIFF
--- a/main.go
+++ b/main.go
@@ -15,6 +15,7 @@ import (
 	"github.com/livepeer/catalyst-api/config"
 	"github.com/livepeer/catalyst-api/metrics"
 	"github.com/livepeer/catalyst-api/pipeline"
+	"github.com/livepeer/catalyst-api/pprof"
 	"github.com/livepeer/livepeer-data/pkg/mistconnector"
 )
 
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -47,6 +49,9 @@ func main() {
 	go func() {
 		log.Fatal(metrics.ListenAndServe(*promPort))
 	}()
+	go func() {
```
leszko, line 52: We do want to run it on prod as well, right? I guess that should be fine, they mention it should not add more than 5% CPU usage, but just want to double-check.

leszko (approved):
LGTM.

Just one comment/question.

main.go
mjh1: @leszko could you link to where you saw the reference to 5% cpu usage? I could be wrong but to my understanding it shouldn't add any extra cpu usage other than having an extra http listener, the actual profiling doesn't occur until you make a request to the handler to fetch some pprof data.

mjh1: https://stackoverflow.com/questions/26545159/how-big-is-pprof-import-overhead-in-go

leszko: Check this out: https://stackoverflow.com/questions/64057727/is-it-ok-to-use-golang-pprof-on-production-without-effecting-performance

PR #478, opened 2023-02-21
Fix retry mechanism

### LINKED ISSUE #473: Fix UploadToOSURL to correctly propagate the error message in case of timeout
Context: https://discord.com/channels/423160867534929930/1076224090941161492/1076419581549228112

leszko:
I discovered that our current upload retry mechanism is fundamentally wrong and therefore it **does not work**. This PR fixes the issue.

fix #473 

One non-backward-compatible change is that we'll no longer have the upload retries metrics.

---------

**Explanation**

We retry saving the data here:

https://github.com/livepeer/catalyst-api/blob/main/clients/object_store_client.go#L75

But what we pass is `data io.Reader` which is not something you can retry, because if some data was already read from the reader, then it's not there anymore. Therefore it does not make sense to have any retry mechanism in the `UploadToOSURL()` function. 

This PR moves the retry mechanism from the `UploadToOSURL()` function upper to the place where it can actually be retried.

### DIFF
--- a/clients/arweave_ipfs_s3.go
+++ b/clients/arweave_ipfs_s3.go
@@ -16,7 +16,7 @@ import (
 const SCHEME_IPFS = "ipfs"
 const SCHEME_ARWEAVE = "ar"
 
-var dstorageRetryBackoff = backoff.NewConstantBackOff(1 * time.Second)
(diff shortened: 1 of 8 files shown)


thomshutt (approved):

thomshutt:
LGTM - any idea why the unit tests were passing?

PR #482, opened 2023-02-22
Reduce the number of goroutines for the directory copy

mjh1:
We have been seeing a lot of connection reset errors when copying the output files, trying a lower number of goroutines to see if it improves the situation since 10 per job seemed a lot. I tested with different numbers of goroutines, even just a single routine and it didn't seem to have a big impact on the time to copy.

### DIFF
--- a/clients/mediaconvert.go
+++ b/clients/mediaconvert.go
@@ -161,8 +161,11 @@ func (mc *MediaConvert) Transcode(ctx context.Context, args TranscodeJobArgs) ([
 
 	mcOutputBaseDir := mc.osTransferBucketURL.JoinPath(mcOutputRelPath, "..")
 	ourOutputBaseDir := args.HLSOutputFile.JoinPath("..")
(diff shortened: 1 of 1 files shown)


mjh1:
@emranemran Just realised that a fix went in from @leszko for the retry mechanism used here, so perhaps this PR is not necessary. What do you guys think, should I just mark this as draft until we see what https://github.com/livepeer/catalyst-api/pull/478 effect has?

leszko:
> @emranemran Just realised that a fix went in from @leszko for the retry mechanism used here, so perhaps this PR is not necessary. What do you guys think, should I just mark this as draft until we see what https://github.com/livepeer/catalyst-api/pull/478 effect has?

I think you're right. We can keep it as a draft and see if it's needed.

thomshutt:
Shall we close this?

mjh1:
Yep 👍

PR #484, opened 2023-02-23
Set a maximum bitrate for our output renditions

mjh1:
- mediaconvert was throwing an error when an input had a very high bitrate, over 288000000
- renamed constants to match the golang convention

### DIFF
--- a/video/profiles.go
+++ b/video/profiles.go
@@ -107,6 +107,10 @@ func GetPlaybackProfiles(iv InputVideo) ([]EncodedProfile, error) {
 	return profiles, nil
 }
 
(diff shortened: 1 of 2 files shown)


video/profiles.go
```diff
@@ -104,10 +109,10 @@ func GetPlaybackProfiles(iv InputVideo) ([]EncodedProfile, error) {
 
 func lowBitrateProfile(video InputTrack) EncodedProfile {
```
thomshutt, line 110: This method's getting (slash already way) pretty complicated to follow - could you add a bit of a descriptive comment please? i.e why do we halve the bitrate to begin with

thomshutt (approved):

PR #491, opened 2023-02-27
concatenate segments

emranemran:
The segments returned from the T are first stored in the order they are
received, and then sorted according to the segment index, and then
concatenated into a single .ts file. A nested map with an outer and
inner table is used to temporarily store the segment data streams.

TODOs (will use follow-up PR for these):
 * write concatenated ts segment to disk
 * fix/add tests

### DIFF
--- a/transcode/transcode.go
+++ b/transcode/transcode.go
@@ -109,9 +109,20 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
 	// transcodedStats hold actual info from transcoded results within requested constraints (this usually differs from requested profiles)
 	transcodedStats := statsFromProfiles(transcodeProfiles)
 
+	renditionList := video.TRenditionList{RenditionSegmentTable: make(map[string]*video.TSegmentList)}
+	// only populate video.TRenditionList map if MP4 is enabled via override or short-form video detection
+	if transcodeRequest.GenerateMP4 {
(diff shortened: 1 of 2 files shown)


mjh1 (commented):
overall lgtm 👍

emranemran:
Updates: 
* gated mp4 generation on GenerateMP4 flag

transcode/transcode.go
```diff
@@ -109,9 +110,20 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
...
+	renditionList := video.TRenditionList{RenditionSegmentTable: make(map[string]video.TSegmentList)}
+	// only populate video.TRenditionList map if MP4 is enabled via override or short-form video detection
+	if transcodeRequest.GenerateMP4 {
+		for _, profile := range transcodeProfiles {
+			renditionList.AddRenditionSegment(profile.Name,
+				video.TSegmentList{
+					SegmentDataTable: make(map[int][]byte),
+				})
+		}
+	}
```
mjh1, line 122: it could be nice to just late initialise these inside `video/media.go` when calls are made to `GetSegmentList` further down

emranemran:
Updates: 
* Addressed some good lint errors where I was copying a mutex around. All fixed now. 
* Unit tests will be added in a follow-up PR.

mjh1 (approved):

PR #492, opened 2023-02-27
Fix data race

mjh1:
We were seeing intermittent data race failures in our unit tests. This backoff implementation is advertised as not thread safe so we need a new instance each time. https://github.com/cenkalti/backoff/blob/v4.2.0/tries.go#L10-L12

### DIFF
--- a/clients/input_copy.go
+++ b/clients/input_copy.go
@@ -20,8 +20,6 @@ const MAX_COPY_FILE_DURATION = 30 * time.Minute
 const MaxInputFileSizeBytes = 10 * 1024 * 1024 * 1024 // 10 GiB
 const PresignDuration = 10 * time.Minute
 
-var RETRY_BACKOFF = backoff.WithMaxRetries(newExponentialBackOffExecutor(), 5)
-
 type InputCopier interface {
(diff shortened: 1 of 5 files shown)


mjh1:
Damn, just noticed there are a few more occurrences that need fixing too.

mjh1:
Ok fixed all of them now 👍

clients/input_copy.go
```diff
@@ -155,3 +153,7 @@ type StubInputCopy struct{}
 func (s *StubInputCopy) CopyInputToS3(requestID string, inputFile, osTransferURL *url.URL) (inputVideoProbe video.InputVideo, signedURL string, err error) {
 	return video.InputVideo{}, "", nil
 }
+
+func RetryBackoff() backoff.BackOff {
+	return backoff.WithMaxRetries(newExponentialBackOffExecutor(), 5)
```
emranemran, line 158: Maybe better to create two functions? 
`func RetryExponentialBackoff()` and `func RetryConstantBackoff()`

emranemran (approved):

thomshutt (approved):

PR #494, opened 2023-02-27
transcode: commonize mp4 flag between pipelines

emranemran:
The AutoMP4 field in the studio request and the input file's duration is checked to enable MP4 generation -- this is currently only implemented for the mediaconvert pipeline. This commit moves the check and decision to HandleStartUploadJob such that it applies to both MediaConvert and Livepeer pipelines.

TODOs: 
* fix tests

### DIFF
--- a/pipeline/coordinator.go
+++ b/pipeline/coordinator.go
@@ -39,6 +39,11 @@ const (
 	StrategyFallbackExternal Strategy = "fallback_external"
 )
 
+const (
+	// Only mp4s of maxMP4OutDuration will have MP4s generated for each rendition
+	maxMP4OutDuration = 2 * time.Minute
+)
+
 func (s Strategy) IsValid() bool {
 	switch s {
 	case StrategyCatalystDominance, StrategyExternalDominance, StrategyBackgroundExternal, StrategyBackgroundMist, StrategyFallbackExternal:
@@ -61,6 +66,7 @@ type UploadJobPayload struct {
 	Profiles              []video.EncodedProfile
 	PipelineStrategy      Strategy
 	AutoMP4               bool
+	GenerateMP4           bool
 	InputFileInfo         video.InputVideo
 	SignedSourceURL       string
 }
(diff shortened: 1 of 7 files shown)


pipeline/external.go
```diff
@@ -33,7 +37,12 @@ func (e *external) HandleStartUploadJob(job *JobInfo) (*HandlerOutput, error) {
 		InputFile:     sourceFileUrl,
 		HLSOutputFile: job.TargetURL,
 		Profiles:      job.Profiles,
-		AutoMP4:       job.AutoMP4,
+		GenerateMP4: func(automp4 bool, duration float64) bool {
+			if automp4 && duration <= maxMP4OutDuration.Seconds() {
+				return true
+			}
+			return false
```
mjh1, line 44: I think this would make sense to be moved into pipeline/coordinator.go `StartUploadJob()`. Currently being in pipeline/external.go it means it still only applies for mediaconvert, there is a separate version of this `HandleStartUploadJob()` function in pipeline/mist.go

emranemran: Doh. I missed the Handler interface completely. Fixing...

pipeline/coordinator.go
```diff
@@ -39,6 +39,11 @@ const (
 	StrategyFallbackExternal Strategy = "fallback_external"
 )
 
+const (
+	// Only mp4s of maxMP4OutDuration will have MP4s generated for each rendition
+	maxMP4OutDuration = 2 * time.Minute
```
gioelecerati, line 44: We may consider making this part of the TranscodeJobArgs instead of a fixed value into catalyst-api? Not necessarily in this pr
LGTM, in any case

gioelecerati (approved):

pipeline/coordinator.go
```diff
@@ -231,6 +237,12 @@ func (c *Coordinator) StartUploadJob(p UploadJobPayload) {
 		p.SourceFile = newSourceURL.String()
 		p.SignedSourceURL = signedURL
 		p.InputFileInfo = inputVideoProbe
+		p.GenerateMP4 = func(automp4 bool, duration float64) bool {
```
mjh1, line 240: i've never seen an inline function used like this before, pretty neat 👍

mjh1 (approved):

PR #495, opened 2023-02-28
Increase presigned url expiry

mjh1:
I realised that there is some retry logic for the probing ahead of the presigned URL being sent to MediaConvert, so I think it's safer to increase the validity period a bit in case it expires by the time MediaConvert reads it.


thomshutt (commented):
Shouldn't be any harm in having this be even longer

thomshutt (approved):

PR #499, opened 2023-03-01
Fix IPFS resource ID parsing

mjh1:
We were ignoring the path of the incoming IPFS urls which meant we were fetching directory/contents pages rather than the intended files. From testing it appears that if we concatenate the host and path we get the correct result, for IPFS urls without any path this continues to work fine.

(Also removed an unused function in main.go)

### DIFF
--- a/clients/arweave_ipfs_s3.go
+++ b/clients/arweave_ipfs_s3.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"path"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


mjh1:
Verified my assumption about the urls here: https://docs.ipfs.tech/how-to/address-ipfs-on-web/#native-urls

Maybe we should be passing queryparams through too

main.go
```diff
@@ -97,15 +97,6 @@ func parseURL(s string, dest **url.URL) error {
 	return nil
 }
 
-func URLVarFlag(fs *flag.FlagSet, dest **url.URL, name, value, usage string) {
```
thomshutt, line 100: I'm surprised linting hadn't caught this

thomshutt (approved):

clients/arweave_ipfs_s3.go
```diff
@@ -47,7 +48,7 @@ func DownloadDStorageFromGatewayList(u string, requestID string) (io.ReadCloser,
 		resourceID = dStorageURL.Host
 	} else if dStorageURL.Scheme == SCHEME_IPFS {
 		gateways = config.ImportIPFSGatewayURLs
-		resourceID = dStorageURL.Host
+		resourceID = path.Join(dStorageURL.Host, dStorageURL.Path)
```
victorges, line 51: Ah! Nice catch

victorges (approved):
LGTM

emranemran (approved):

PR #501, opened 2023-03-02
transmux ts segments into mp4

emranemran:
Final PR to actually transmux concatenated ts segments and upload. This is a first iteration and goal is to keep it simple first and then add more features like asynchronous processing via ffmpeg and/or better memory management (if required).

### DIFF
--- a/transcode/transcode.go
+++ b/transcode/transcode.go
@@ -1,10 +1,13 @@
 package transcode
 
 import (
+	"bufio"
 	"bytes"
 	"fmt"
 	"net/url"
+	"os"
 	"path"
+	"path/filepath"
 	"strings"
 	"sync"
 	"time"
@@ -18,6 +21,7 @@ import (
 )
 
 const UPLOAD_TIMEOUT = 5 * time.Minute
+const TRANSMUX_STORAGE_DIR = "/tmp/transmux_stage"
 
 type TranscodeSegmentRequest struct {
 	SourceFile        string                 `json:"source_location"`
@@ -140,20 +144,6 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
 		return outputs, segmentsCount, err
 	}
 
-	// TODO: remove this in follow-up PR to write files to disk for transmuxing
-	/*
-		if transcodeRequest.GenerateMP4 {
-			for rlist, slist := range renditionList.RenditionSegmentTable {
-				table := slist.SegmentDataTable
-				for s, d := range table {
-					fmt.Println("debug-segments:", rlist, ":", strconv.Itoa(s)+".ts", "size:", len(d))
-				}
-				for _, k := range slist.GetSortedSegments() {
-					fmt.Println("debug-segments:", k, len(slist.SegmentDataTable[k]))
-				}
-			}
-		}
-	*/
 	// Build the manifests and push them to storage
 	manifestURL, err := GenerateAndUploadManifests(sourceManifest, targetTranscodedRenditionOutputURL.String(), transcodedStats)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


transcode/transcode.go
```diff
@@ -165,6 +156,92 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
 		return outputs, segmentsCount, err
 	}
 
+	// transmux received segments from T into a single mp4
+	if transcodeRequest.GenerateMP4 {
+		for rlist, slist := range renditionList.RenditionSegmentTable {
+			// create folder
+			if _, err := os.Stat(TRANSMUX_STORAGE_DIR); os.IsNotExist(err) {
+				os.MkdirAll(TRANSMUX_STORAGE_DIR, 0700)
+			}
```
mjh1, line 165: we could just omit the existence check since `MkdirAll` just does a noop if it already exists

transcode/transcode.go
```diff
@@ -165,6 +156,92 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
...
+
+			f2, err := os.Open(fileName+".mp4")
+			if err != nil {
+				log.Log(transcodeRequest.RequestID, "XXX: error opening mp4", "file", f2.Name(), "err", err)
+				break
+			}
+			defer f2.Close()
+
+			// clean-up intermediary .ts file
+			err = os.Remove(f.Name())
```
mjh1, line 224: we should do this in a defer function so that even if there's an error further up we'll still clean this up

transcode/transcode.go
```diff
@@ -165,6 +156,92 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
...
+			err = ffmpeg.Input(f.Name()).
+				Output(fileName+".mp4", ffmpeg.KwArgs{"c": "copy", "bsf:a": "aac_adtstoasc"}).
+				OverWriteOutput().ErrorToStdOut().Run()
+			if err != nil {
+				log.Log(transcodeRequest.RequestID, "XXX: error transmuxing", "file", f.Name(), "err", err)
+				f.Close()
+				continue
+			}
+
+			f2, err := os.Open(fileName+".mp4")
```
mjh1, line 216: do we need to delete this in a defer function too?

transcode/transcode.go
```diff
@@ -165,6 +156,92 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
...
+			defer f2.Close()
+
+			// clean-up intermediary .ts file
+			err = os.Remove(f.Name())
+			if err != nil {
+				log.Log(transcodeRequest.RequestID, "XXX: error removing concat ts file")
+			}
+			
+			// TODO: upload mp4 and delete from local on success
+		        targetMP4URL, err := url.JoinPath(targetTranscodedRenditionOutputURL.String(), rlist)
```
mjh1, line 230: you can just do `targetTranscodedRenditionOutputURL.JoinPath(rlist).String()` no error to check then either which is nice

transcode/transcode.go
```diff
@@ -165,6 +156,92 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
...
+				}
+				totalBytes = totalBytes + int64(segBytes)
+			}
+			// Verify the total bytes written for a given rendition matches the total # of bytes we received from T
+			renditionIndex := getProfileIndex(transcodeProfiles, rlist)
+			fmt.Println("XXX: ", renditionIndex, "transcodeProfiles", transcodeProfiles, "rlist", rlist, "transcodedStats", transcodedStats)
+			var rendBytesWritten int64 = -1
+			for _, v := range transcodedStats {
+				if v.Name == rlist {
+					rendBytesWritten = v.Bytes
```
mjh1, line 196: it doesn't look like this `Bytes` value will be filled in.. the stats are populated here: https://github.com/livepeer/catalyst-api/blob/ae7536c9a4ff9f95db94646ab83208af69cf9908/transcode/transcode.go#L255

emranemran: Actually comes from here: 
https://github.com/livepeer/catalyst-api/blob/b003b72d49822911cd0df143fcb8a86cb6e6c650/transcode/transcode.go#L251

And during testing, I can see it get populated as expected as we get results from transcoder.

transcode/transcode.go
```diff
@@ -165,6 +155,68 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
...
+			}
+
+			// b. create a single .ts file for a given rendition by concatenating all segments in order
+			concatTsFileName := filepath.Join(TRANSMUX_STORAGE_DIR, transcodeRequest.RequestID+"_"+rendition+".ts")
+			totalBytes, err := video.ConcatTS(concatTsFileName, segments)
+			if err != nil {
+				log.Log(transcodeRequest.RequestID, "error concatenating .ts", "file", concatTsFileName, "err", err)
+				continue
+			}
+			defer os.Remove(concatTsFileName)
```
mjh1, line 175: we just need to move this line onto the line above the call to `video.ConcatTS` so that if there's an error from that function on any of the individual writes then we will still clean up the file. It doesn't matter if we try to delete it before it gets created as we're not doing anything with the error from `os.Remove` anyway.

mjh1 (commented):
One small change needed but otherwise lgtm

transcode/transcode.go
```diff
@@ -165,6 +155,68 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
 		return outputs, segmentsCount, err
 	}
 
+	// Transmux received segments from T into a single mp4
+	if transcodeRequest.GenerateMP4 {
+		for rendition, segments := range renditionList.RenditionSegmentTable {
+			// a. create folder to hold transmux-ed files in local storage temporarily
+			err := os.MkdirAll(TRANSMUX_STORAGE_DIR, 0700)
+			if err != nil && !os.IsExist(err) {
```
mjh1, line 163: `!os.IsExist(err)`? is that meant to be `!os.IsExist(TRANSMUX_STORAGE_DIR)`

transcode/transcode.go
```diff
@@ -165,6 +155,68 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
...
+				return clients.UploadToOSURL(targetMP4URL, filepath.Base(mp4OutputFile.Name()), bufio.NewReader(mp4OutputFile), UPLOAD_TIMEOUT)
+			}, clients.UploadRetryBackoff())
+			if err != nil {
+				log.Log(transcodeRequest.RequestID, "failed to upload mp4", "file", mp4OutputFile.Name())
+			}
+		}
+	}
+
 	manifestURL = strings.ReplaceAll(manifestURL, targetTranscodedRenditionOutputURL.String(), playbackBaseURL)
 	output := video.OutputVideo{Type: "object_store", Manifest: manifestURL}
```
mjh1, line 221: Will you be adding the mp4 urls to the output struct in a follow up PR?

transcode/transcode.go
emranemran: Good catch - fixed.

transcode/transcode.go
emranemran: Nope, actually takes an `err`: https://pkg.go.dev/os#IsExist

transcode/transcode.go
emranemran: Yeah this change will go in when we make catalyst the primary pipeline.

transcode/transcode.go
mjh1: oh! nice

mjh1 (approved):

PR #503, opened 2023-03-03
Allow custom segment size to be specified on the API

thomshutt:
The change itself is simple, but the PR is slightly bigger because it also updates the integration tests to get as far as the segmenting call to Mist so that we can assert the segment size parameter is correctly passed through.


handlers/schemas/UploadVOD.yaml
```diff
@@ -2,7 +2,6 @@ type: "object"
 properties:
   url:
     type: "string"
-    format: "uri"
```
thomshutt, line 5: This is so that we can use local file paths in testing

thomshutt: TODO: Check ObjectStore and see if we can use a file:// syntax to avoid changing this

emranemran: Wouldn't this allow end-users to pass a local file using file:// to the studio api?

emranemran (approved):

pipeline/mist.go
```diff
@@ -16,7 +17,6 @@ import (
 )
 
 const MIST_SEGMENTING_SUBDIR = "source"
-const TARGET_SEGMENT_LENGTH_SECS = "5"
```
mjh1, line 19: why are we going with 10 seconds as the default now rather than 5?

test/steps/http.go
```diff
@@ -45,27 +60,39 @@ func (s *StepContext) SetAuthHeaders() {
...
+					"source_segments": true,
+					"transcoded_segments": true
 				}
-			},
+			}
+		]
+	}`, sourceFilename)
+}
+
+func (s *StepContext) getCustomSegmentSizeUploadRequestPayload(sourceFilename string) string {
```
mjh1, line 80: maybe we should make use of the struct definitions, build a struct with whatever optional fields we need then marshal to json?

pipeline/mist.go
thomshutt: We were doing 10 for Mediaconvert and I think longer segments suits our architecture better on the Livepeer side (fewer requests to transcoders)

test/steps/http.go
thomshutt: Yeah good call, I'd convinced myself that it wasn't worth it until we need to change one more thing but probably best to just get it done now.

handlers/schemas/UploadVOD.yaml
thomshutt: Studio rejects it 
```
{
  "errors": [
    "{\"keyword\":\"pattern\",\"dataPath\":\".url\",\"schemaPath\":\"#/properties/url/pattern\",\"params\":{\"pattern\":\"^(https?|ipfs|ar)://\"},\"message\":\"should match pattern \\\"^(https?|ipfs|ar)://\\\"\"}"
  ]
}
```

thomshutt: `file://` seems to work on our side though, so I've updated the tests and added this validation back in

test/steps/upload_request.go
```diff
@@ -0,0 +1,43 @@
...
+	TranscodedSegments bool `json:"transcoded_segments,omitempty"`
+}
+
+type OutputLocation struct {
+	Type    string `json:"type,omitempty"`
+	URL     string `json:"url,omitempty"`
+	Outputs Output `json:"outputs,omitempty"`
+}
+
+type UploadRequest struct {
```
mjh1, line 16: Wouldn't it be ok to just use the real struct defs: https://github.com/livepeer/catalyst-api/blob/9cb23e73717cd4f34c90d5a12606e7cb49d21348/handlers/upload.go#L38 ?

thomshutt: It would, but I want to keep the test and real code separate - this will hopefully help catch weird cases around missing vs. blank parameters etc.

PR #509, opened 2023-03-07
Check codecs are compatible with mist

mjh1:
Only send to mist if the input codecs are known to be compatible to avoid unnecessary failures


thomshutt (approved):
Really nice, I think this all looks right and I like that you made it continue using the passed in strategy in the happy path rather than hardcoding `StrategyFallbackExternal` and that `StrategyCatalystDominance` always takes precedence for the tests (which were both things I hadn't thought of).

Could you just add a couple of comments on this method to explain that please, since I think without all this context it might not be obvious

PR #512, opened 2023-03-08
Set a default bitrate if we don't find one

mjh1:
This will allow us to continue processing videos where the only issue is a missing bitrate in the probe output

### DIFF
--- a/video/probe.go
+++ b/video/probe.go
@@ -37,7 +37,7 @@ func (p Probe) ProbeFile(url string) (iv InputVideo, err error) {
 	return parseProbeOutput(data)
 }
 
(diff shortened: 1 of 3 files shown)


video/probe.go
```diff
@@ -37,7 +37,7 @@ func (p Probe) ProbeFile(url string) (iv InputVideo, err error) {
 	return parseProbeOutput(data)
 }
 
-func parseProbeOutput(probeData *ffprobe.ProbeData) (InputVideo, error) {
+func parseProbeOutput(probeData *ffprobe.ProbeData) (iv InputVideo, err error) {
```
thomshutt, line 40: Let's not used named returns, since it can make the flow confusing to reason about

thomshutt (approved):
Other than my comment, this looks great

PR #514, opened 2023-03-09
Increase max input file size from 10Gb -> 30Gb

thomshutt:


config/config.go
```diff
@@ -34,6 +34,9 @@ const MaxSegmentSizeSecs = 20
 // the future.
 const MAX_JOBS_IN_FLIGHT = 8
 
+// The maximum allowed input file size
+const MaxInputFileSizeBytes = 30 * 1024 * 1024 * 1024 // 10 GiB
```
mjh1, line 38: the comment needs updating to 30 :)

mjh1 (approved):

config/config.go
thomshutt: 😄 good spot

PR #517, opened 2023-03-10
Log input video/audio track info

emranemran:
Log input video/audio track info so that we can identify any raciness in the ffprobe behavior to root
cause cases where the input file uses mp4 + aac but still ends up in the
mediaconvert pipeline instead of solely using the mist pipeline.

### DIFF
--- a/clients/input_copy.go
+++ b/clients/input_copy.go
@@ -62,21 +62,23 @@ func (s *InputCopy) CopyInputToS3(requestID string, inputFile, osTransferURL *ur
 		return
 	}
 	log.Log(requestID, "probe succeeded", "source", inputFile.String(), "dest", osTransferURL.String())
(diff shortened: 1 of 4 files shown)


clients/input_copy.go
```diff
@@ -62,21 +62,23 @@ func (s *InputCopy) CopyInputToS3(requestID string, inputFile, osTransferURL *ur
 		return
 	}
 	log.Log(requestID, "probe succeeded", "source", inputFile.String(), "dest", osTransferURL.String())
-	videoTrack, err := inputVideoProbe.GetVideoTrack()
+	videoTrack, err := inputVideoProbe.GetTrack(video.TrackTypeVideo)
```
thomshutt, line 65: Isn't `GetVideoTrack` a shorthand for `GetTrack(video.TrackTypeVideo)`?

thomshutt (approved):

PR #518, opened 2023-03-10
Add mp4 override flag

emranemran:


handlers/upload.go
```diff
@@ -178,6 +179,7 @@ func (d *CatalystAPIHandlersCollection) handleUploadVOD(w http.ResponseWriter, r
 		RequestID:             requestID,
 		Profiles:              uploadVODRequest.Profiles,
 		PipelineStrategy:      uploadVODRequest.PipelineStrategy,
+		ForceMP4:              targetOutput.Outputs.ForceMP4,
 		AutoMP4:               targetOutput.Outputs.AutoMP4,
```
emranemran, line 183: @gioelecerati @victorges - I think all these MP4 fields are starting to get a bit confusing now. Do you guys think having an `AutoMP4` field is still required from the Studio perspective?

gioelecerati: I think we could remove AutoMp4 potentially, considering it always true for assets < 2min or even better make it as a arg config parameter of catalyst instead of a parameter specified by Task Runner in the outputLocation, wdyt?

emranemran: I'm in favor of completely removing AutoMP4. I think the decision tree from Studio would look like: 
- ForceMP4 is `true`: generate mp4s regardless of length
- ForceMP4 is `false`: generate mp4s based on what catalyst-api sets as default (i.e. video duration < 2min)

@victorges any thoughts on this?

victorges: I think we don't want the "auto mp4s" option to be always true, especially for the lower-level transcode API in which we are writing to user-owned storage. We never create MP4s in that case and it would come off a surprise to users if we started generating things that they don't ask for in their own storage. This would be also more expensive for them as they'd pay for that storage without making use of them.
e.g. I want an IPFS CAR with an HLS playlist of my video now a bunch of MP4s would show up there without me asking for them and my bill is 2x higher (I guess).

One thing that we could do would be replacing the multiple confusing booleans with some kind of "enum" instead. So we could have perhaps `transcoded_mp4: "auto/on/off"` or something like that. WDYT?

pipeline/coordinator.go
```diff
@@ -240,12 +241,12 @@ func (c *Coordinator) StartUploadJob(p UploadJobPayload) {
...
 		p.InputFileInfo = inputVideoProbe
-		p.GenerateMP4 = func(automp4 bool, duration float64) bool {
-			if automp4 && duration <= maxMP4OutDuration.Seconds() {
+		p.GenerateMP4 = func(forcemp4, automp4 bool, duration float64) bool {
+			if forcemp4 || (automp4 && duration <= maxMP4OutDuration.Seconds()) {
 				return true
 			}
 			return false
-		}(p.AutoMP4, p.InputFileInfo.Duration)
+		}(p.ForceMP4, p.AutoMP4, p.InputFileInfo.Duration)
```
victorges, line 249: Not a change from this PR but curious about the function indirection. Why not?:

```suggestion
		p.GenerateMP4 = p.ForceMP4 || (p.AutoMP4 && p.InputFileInfo.Duration <= maxMP4OutDuration.Seconds())
```

victorges (approved):
LGTM either way

pipeline/coordinator.go
emranemran: I had a couple more things happening in this function but since removed - will clean this up in next PR.

PR #522, opened 2023-03-14
starting point for new playback endpoint

mjh1:


playback/manifests_test.go
```diff
@@ -0,0 +1,89 @@
...
+	require.NoError(t, err)
+	config.PrivateBucketURL = privateBucket
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := Manifest(tt.req)
+			if tt.expectedErr != "" {
+				require.ErrorContains(t, err, tt.expectedErr)
+			} else {
+				require.NoError(t, err)
+				fmt.Println(got)
```
gioelecerati, line 79: I think this print got left while debugging

handlers/playback.go
```diff
@@ -0,0 +1,101 @@
...
+			handleError(err, req, requestID, w)
+			return
+		}
+
+		key := req.Form.Get(playback.ManifestKeyParam)
+		if err := checkKey(key); err != nil {
+			handleError(err, req, requestID, w)
+			return
+		}
+		manifest, err := playback.Manifest(playback.PlaybackRequest{
```
gioelecerati, line 30: I am not sure how go behave, is this reader returned from the Manifest function getting closed here? Or do we need to return a `io.ReadCloser` instead and defer here manifest.Close()?

gioelecerati (approved):
LGTM with comments

handlers/playback.go
mjh1: we're ok because the underlying object is a bytes.Buffer which only implements reader: https://github.com/grafov/m3u8/blob/master/writer.go#L73

PR #533, opened 2023-03-17
fix fallback job primary-key db entry

emranemran:
coordinator: use fb_ prefix for request-id if fallback pipeline is being used

### DIFF
--- a/pipeline/coordinator.go
+++ b/pipeline/coordinator.go
@@ -324,6 +324,9 @@ func (c *Coordinator) startOneUploadJob(p UploadJobPayload, handler Handler, for
 		// this will prevent the callbacks for this job from actually being sent
 		p.CallbackURL = ""
 	}
+	if p.InFallbackMode {
+		p.RequestID = fmt.Sprintf("fb_%s", p.RequestID)
+	}
 	streamName := config.SegmentingStreamName(p.RequestID)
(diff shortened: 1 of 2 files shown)


pipeline/coordinator.go
```diff
@@ -318,7 +318,7 @@ func checkMistCompatibleCodecs(strategy Strategy, iv video.InputVideo) Strategy
 // failures (today this means re-running the job in another pipeline). If it's
 // set to true, error callbacks from this job will not be sent.
 func (c *Coordinator) startOneUploadJob(p UploadJobPayload, handler Handler, foreground, hasFallback bool) <-chan bool {
-	if !foreground {
+	if !foreground || p.InFallbackMode {
```
thomshutt, line 321: Isn't this going to break things by stopping callbacks to Studio in fallback mode though?

thomshutt: Also maybe we should do something like `p.RequestID = fmt.Sprintf("fb_%s", p.RequestID)` to distinguish fallback requests from background and foreground ones

emranemran: Yes...yes it will. Fixing.

emranemran:
Oops just noticed the failures - I'll need to fix the tests.

PR #538, opened 2023-03-22
Access control gate API calls

mjh1:
This builds on the playback endpoint to add in a middleware which makes the Gate API calls to confirm whether playback is allowed or not.
The other main change is serving both manifests and media from a single endpoint, this was done to be able to serve static mp4s, it would've been too messy to try to rewrite these urls with a different path before hitting the playback endpoint here. So now I check the content-type and file extension to determine whether the file is a manifest.

### DIFF
--- a/playback/manifests.go
+++ b/playback/manifests.go
@@ -23,6 +23,7 @@ type Request struct {
 	PlaybackID string
 	File       string
 	AccessKey  string
+	Range      string
 }
 
 type Response struct {
@@ -31,7 +32,7 @@ type Response struct {
 }
 
 func Handle(req Request) (*Response, error) {
-	f, err := osFetch(req.PlaybackID, req.File)
+	f, err := osFetch(req.PlaybackID, req.File, req.Range)
 	if err != nil {
 		return nil, err
 	}
@@ -95,9 +96,9 @@ func appendAccessKey(uri, accessKey string) (string, error) {
 	return variantURI.String(), nil
 }
 
-func osFetch(playbackID, file string) (*drivers.FileInfoReader, error) {
+func osFetch(playbackID, file, byteRange string) (*drivers.FileInfoReader, error) {
 	osURL := config.PrivateBucketURL.JoinPath("hls").JoinPath(playbackID).JoinPath(file)
-	f, err := clients.GetOSURL(osURL.String())
+	f, err := clients.GetOSURL(osURL.String(), byteRange)
 	if err != nil {
 		var awsErr awserr.Error
 		if errors.As(err, &awsErr) && awsErr.Code() == s3.ErrCodeNoSuchKey {
--- a/api/http.go
+++ b/api/http.go
@@ -57,7 +57,7 @@ func NewCatalystAPIRouter(cli config.Cli, vodEngine *pipeline.Coordinator, bal b
 
 	router.GET("/ok", withLogging(catalystApiHandlers.Ok()))
 
-	// Manifest endpoint
+	// Playback endpoint
 	router.GET("/asset/hls/:playbackID/*file",
 		withLogging(
 			withGatingCheck(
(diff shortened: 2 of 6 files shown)


mjh1:
Marked as draft until https://github.com/livepeer/catalyst/pull/437 is merged and the dependency is updated here. Still ready for review though, no other changes I'm aware of.

middleware/gating.go
```diff
@@ -0,0 +1,70 @@
...
+func NewGatingHandler(gateURL string) *GatingHandler {
+	return &GatingHandler{
+		AccessControl: accesscontrol.NewPlaybackAccessControl(gateURL),
+	}
+}
+
+func (h *GatingHandler) GatingCheck(next httprouter.Handle) httprouter.Handle {
+	return func(w http.ResponseWriter, req *http.Request, params httprouter.Params) {
+		var requestID = config.RandomTrailer(8)
+		req.Header.Set("requestID", requestID)
```
victorges, line 35: Perhaps a common helper like `getOrSetRequestId(req.Header)` would be useful across middlewares/handlers.

playback/manifests.go
```diff
@@ -5,37 +5,51 @@ import (
...
-	reader, err := osFetch(req.PlaybackID, req.File)
+func Handle(req Request) (*Response, error) {
+	f, err := osFetch(req.PlaybackID, req.File)
 	if err != nil {
 		return nil, err
 	}
-	defer reader.Close()
 
-	p, listType, err := m3u8.DecodeFrom(reader, true)
+	if !strings.HasSuffix(req.File, "m3u8") && f.ContentType != "application/x-mpegurl" {
```
victorges, line 39: Perhaps skip the content-type check. We don't want to risk trying to parse other kinds of files as manifests (e.g. metadata or src files) and failing. Also would add `.` prefix to m3u8, wdyt?
```suggestion
	if !strings.HasSuffix(req.File, ".m3u8") {
```

victorges (commented):
Looking good

test/cucumber_test.go
```diff
@@ -110,20 +111,3 @@ func InitializeScenario(ctx *godog.ScenarioContext) {
 		return ctx, nil
 	})
 }
-
-func TestFeatures(t *testing.T) {
```
mjh1, line 114: removed this to stop tests running twice once with `make test` and again with `make integration-test`

gioelecerati (approved):

middleware/gating.go
```diff
@@ -0,0 +1,52 @@
...
+			deny(w)
+			return
+		}
+
+		next(w, req, params)
+	}
+}
+
+func deny(w http.ResponseWriter) {
+	catErrs.WriteHTTPUnauthorized(w, "denied", errors.New("access control failed"))
```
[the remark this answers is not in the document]
mjh1: Good call, i'm going to raise a separate PR since this one is already quite large and it's not just like a one-line change

handlers/accesscontrol/access-control.go
```diff
@@ -101,70 +102,43 @@ func (ac *AccessControlHandlersCollection) handleUserNew(payload []byte) []byte
...
-	}
-
-	body, err := json.Marshal(PlaybackAccessControlRequest{Type: "jwt", Pub: pubKey, Stream: playbackID})
-	if err != nil {
-		glog.Errorf("Unable to get playback access control info, JSON marshalling failed. playbackId=%v pubkey=%v", playbackID, pubKey)
-		return []byte("false")
-	}
-
-	playbackAccessControlAllowed, err := ac.GetPlaybackAccessControlInfo(playbackID, pubKey, body)
+	playbackAccessControlAllowed, err := ac.IsAuthorized(playbackID, requestURL)
```
victorges, line 105: Neeeat

victorges (commented):
Gave a post-merge skim and looks great!

PR #541, opened 2023-03-24
add mist cleanup script to catalyst-api

emranemran:
This PR integrates the mist-cleanup script that will be called periodically to clean up /dev/shm. A separate PR will be used to actually schedule and call this script. 

Best to review on a per commit basis. @Thulinma: FYI so that you can see the diff from your original script. 

Related Catalyst PR to install the script in `build/` dir: https://github.com/livepeer/catalyst/pull/441

### DIFF
--- a/scripts/mist-cleanup.sh
+++ b/scripts/mist-cleanup.sh
@@ -7,49 +7,49 @@ containsElement () {
   return 1
 }
 
-SEMS=`ls -tr /dev/shm/sem.MstTRKS* /dev/shm/sem.MstUser* /dev/shm/sem.MstInpt* /dev/shm/sem.MstPull* 2> /dev/null`
-PULS=`ls -tr /dev/shm/sem.MstPull* 2> /dev/null`
+SEMS=$(ls -tr /dev/shm/sem.MstTRKS* /dev/shm/sem.MstUser* /dev/shm/sem.MstInpt* /dev/shm/sem.MstPull* 2> /dev/null)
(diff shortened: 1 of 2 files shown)


Thulinma:
The lines originally ending in ` > /dev/null` were not meant to be logged - those would end up being _extremely_ spammy if left in. (They were used for debugging/checking the script while it was being written only.)

thomshutt (approved):
Structurally looks fine, will let Jaron address the content

Thulinma (dismissed):
As long as the lines that originally went to `/dev/null` don't end up in the log, everything else looks good to me. The other changes to the script should not harm anything as far as I can see.

scripts/mist-cleanup.sh
```diff
@@ -0,0 +1,82 @@
...
+  STRM=${P:17} #Strip the beginning
+  if containsElement "$STRM" "$ACTS" ; then
+    log "Yes: ${STRM} (${P})"
+  else
+    log "STATE: ${P}"
+    rm "${P}"
+  fi
+done
+
+log "Done. Deleted ${DATA_DEL} data pages"
```
pwilczynskiclearcode, line 82: Is it worth calculating and displaying number of deleted semaphores and state pages?

emranemran:
> As long as the lines that originally went to `/dev/null` don't end up in the log, everything else looks good to me. The other changes to the script should not harm anything as far as I can see.

OK, added logging levels to turn those lines off.

PR #542, opened 2023-03-24
add methods to shell-out with a schedule

emranemran:
Periodic shell scripts will be used to handle work that will happen
out-of-band at the system level. The first use-case here is to call
the `mist-cleanup` script which will clean leftover Mist resources from
/dev/shm.


main.go
```diff
@@ -85,6 +86,16 @@ func main() {
 		log.Println(handlers.DebugEndpointListenAndServe(*debugPort, vodEngine))
 	}()
 
+	// Start cron style apps to run periodically
+	app := "mist-cleanup"
+	// schedule mist-cleanup every 2hrs with a timeout of 5min
```
emranemran, line 91: @Thulinma: any concerns/comments with the mist-cleanup script running every 2hrs? What do you suggest?

middleware/shell.go
```diff
@@ -0,0 +1,69 @@
...
+		for {
+			select {
+			case <-ticker.C:
+				err := s.Run()
+				if err != nil {
+					log.Println("Failed to start cmd:", s.Cmd)
+					break
+				}
+			}
+		}
```
emranemran, line 48: Getting a `  Error: S1000: should use for range instead of for { select {} } (gosimple)` linter error which seems like a false positive for this use case here? Any suggestions @mjh1 @thomshutt?

main.go
Thulinma: Every two hours is fine - I was personally thinking every hour, but that choice is pretty much arbitrary. As long as we don't run it more than every 15m and not less than once a day, I think we're probably good.

middleware/shell.go
mjh1: found this: https://github.com/dominikh/go-tools/issues/503
I think what it's really complaining about is the fact that's there's just a single case so I think it could just be replaced with my suggestion above

middleware/shell.go
```diff
@@ -0,0 +1,69 @@
...
+		defer ticker.Stop()
+		for {
+			select {
+			case <-ticker.C:
+				err := s.Run()
+				if err != nil {
+					log.Println("Failed to start cmd:", s.Cmd)
+					break
+				}
+			}
```
mjh1, line 47:
```suggestion
			<-ticker.C
			err := s.Run()
			if err != nil {
				log.Println("Failed to start cmd:", s.Cmd)
				break
			}
```

mjh1: Just tested that and seems to work

mjh1: Oh and you can just do `for range ticker.C {` to get rid of one more line

middleware/shell.go
```diff
@@ -0,0 +1,69 @@
...
+	"sync"
+	"time"
+)
+
+type Shell struct {
+	mu           sync.Mutex
+	Cmd          string
+	Args         []string
+	IntervalSecs int64
+	TimeoutSecs  int64
```
mjh1, line 17: could you make these `time.Duration`s all the way through so that it's clear from the calling code what we're setting (rather than potential confusion over the unit)

middleware/shell.go
emranemran: good call - i think we can use the `range` style you suggested and get around this.

middleware/shell.go
```diff
@@ -0,0 +1,69 @@
...
+					log.Println("Failed to start cmd:", s.Cmd)
+					break
+				}
+			}
+		}
+	}()
+	return ticker
+}
+
+func (s *Shell) Run() error {
```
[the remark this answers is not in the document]
emranemran: Kept it external in case we need to run a cmd just once and exit instead of something in the background via RunBg.

middleware/shell.go
emranemran: Yeah good suggestion, will do.

emranemran:
Updates: 
* addressed comments
* a few minor changes in the logging syntax so that we can grep it easily on loki

mjh1 (approved):

PR #543, opened 2023-03-24
Fix confusing error message

mjh1:
In a few places we're passing in the full destination URL as the `osURL` param with `filename` blank. This meant the error message made it appear like there was a bug with filename being empty


thomshutt (approved):
Thanks for fixing this, it's been bugging me in the logs for so long!

PR #546, opened 2023-03-28
VOD MP4 separate from HLS

leszko:
Separate HLS and MP4 outputs as described in the [tech doc](https://www.notion.so/livepeer/VOD-MP4-separate-from-HLS-34efd71b091a491ea4e785dda3bf9873).

Now, it's possible to request generating only HLS or only MP4 outputs from catalyst.

### Non-backward-compatible changes:
- Catalyst VOD API (as defined in [tech doc](https://www.notion.so/livepeer/VOD-MP4-separate-from-HLS-34efd71b091a491ea4e785dda3bf9873))
- Changed the source bucket path for transfer from `transfer/requestID` to `requestID/transfer` (to be consistent with `source` paths)
- Use always the bucket defined by the`source-output` flag for storing source segments

### Deployment plan

Since this PR introduces some non-backward-compatible changes between catalyst-api and task-runner, I included a temporary commit https://github.com/livepeer/catalyst-api/pull/546/commits/b057cf3107a5b7c761a068e80a976e1c5408a5a8 which provides the backward-compatibility. Therefore the deployment plan is as follows:
1. Merge this PR and deploy into prod
2. Merge task-runner PR (https://github.com/livepeer/task-runner/pull/176) and deploy into prod
3. Revert https://github.com/livepeer/catalyst-api/pull/546/commits/b057cf3107a5b7c761a068e80a976e1c5408a5a8


thomshutt (approved):

emranemran:
Overall refactor LGTM. A few comments: 
* Just thought of a potential issue - requesting MP4s in the Mist only pipeline might be an issue if the duration exceeds 2min. Mainly because the segments are all saved locally and transmuxed into an mp4 which *may* eat up some filesystem/memory resources. We've tested this heavily with short-form video and so far it's been working alright but needs to be considered carefully for video duration that is longer than a few mins.
* How did testing on staging go with these changes (i.e. ignoring the task-runner bits and curl-ing catalyst-api directly)? Did you encounter any issues with larger files?
* Can you share curl request you're using to test just so that we can verify the new fields that were added?  Will also help reference it and make changes to our local test setups quickly.
* minor nit: would be nicer to group the commits together into specific changes - would have made review a bit easier.

leszko:
> * Just thought of a potential issue - requesting MP4s in the Mist only pipeline might be an issue if the duration exceeds 2min. Mainly because the segments are all saved locally and transmuxed into an mp4 which _may_ eat up some filesystem/memory resources. We've tested this heavily with short-form video and so far it's been working alright but needs to be considered carefully for video duration that is longer than a few mins. This is not an issue in the mediaconvert pipeline since it happens on the "aws cloud".

Yeah, I see, the segments are stored in `/tmp`. How about changing Helm Charts and mounting a persistent volume into `/tmp`? I think that should solve the issue for now. And later we should think of a better solution that will not store files in any intermediary files in FS Wdyt? @emranemran @thomshutt 

Saying that, it's not related to this PR, because this PR only changes the API, it does make any changes to the transcoding process. In other words, if `"mp4":"enabled"` now makes the catalyst crash, then the same would happen with the current `"force_mp4": true`.

> * How did testing on staging go with these changes (i.e. ignoring the task-runner bits and curl-ing catalyst-api directly)? Did you encounter any issues with larger files?

It's not yet in staging, I tested it locally. 

> * Can you share curl request you're using to test just so that we can verify the new fields that were added?  Will also help reference it and make changes to our local test setups quickly.

Yeah, sure. Here's the curl command you an use:

```bash
curl --location 'http://localhost:7979/api/vod' \
--header 'Authorization: Bearer IAmAuthorized' \
--data-raw '{
    "url": "https://storage.googleapis.com/thom-vod-testing/mustwork/Pexels%20Videos%203444.mp4",
    "callback_url": "YOUR_WEBHOOK_URL",
    "output_locations": [
        {
            "type": "object_store",
            "url": "YOUR_S3_URL",
            "outputs": {
                "hls": "enabled",
                "mp4": "enabled"
            }
        }
    ]
}'
```

> * minor nit: would be nicer to group the commits together into specific changes - would have made review a bit easier.

Sorry for that, actually it's hard to do now, because the branch has already the `main` merged in. But I'll keep it in mind for the future PRs!

handlers/upload.go
```diff
@@ -22,11 +22,13 @@ import (
...
-	TranscodedSegments bool `json:"transcoded_segments"`
-	ForceMP4           bool `json:"force_mp4"`
-	AutoMP4            bool `json:"auto_mp4"`
+	SourceMp4          bool   `json:"source_mp4"`
+	SourceSegments     bool   `json:"source_segments"`
+	TranscodedSegments bool   `json:"transcoded_segments"`
+	ForceMP4           bool   `json:"force_mp4"`
+	AutoMP4            bool   `json:"auto_mp4"`
+	HLS                string `json:"hls"`
+	MP4                string `json:"mp4"`
```
mjh1, line 31: I can't quite tell what the default behaviour is if both of these are empty. Or is an error returned?

emranemran:
> Saying that, it's not related to this PR, because this PR only changes the API, it does make any changes to the transcoding process. In other words, if "mp4":"enabled" now makes the catalyst crash, then the same would happen with the current "force_mp4": true.

Yeah agree here - the current issue is that Mist fails segmenting for larger files so we never make it to the step where we can transmux the ts segments into an mp4 - hence this path is not validation until we get the Mist failures under control. 

> Yeah, I see, the segments are stored in /tmp. How about changing Helm Charts and mounting a persistent volume into /tmp? I think that should solve the issue for now. And later we should think of a better solution that will not store files in any intermediary files in FS Wdyt? @emranemran @thomshutt

I'll file a ticket so we don't forget. Once the Mist segmenting issues are resolved, we need to revisit this logic to understand how larger files behave and if required, make some changes. I don't think persistent volume would do much here in the worst case - instead we need to refactor to find a way that offsets having to concatenate the segments locally and then uploading.

leszko:
> > Saying that, it's not related to this PR, because this PR only changes the API, it does make any changes to the transcoding process. In other words, if "mp4":"enabled" now makes the catalyst crash, then the same would happen with the current "force_mp4": true.
> 
> Yeah agree here - the current issue is that Mist fails segmenting for larger files so we never make it to the step where we can transmux the ts segments into an mp4 - hence this path is not validation until we get the Mist failures under control.
> 
> > Yeah, I see, the segments are stored in /tmp. How about changing Helm Charts and mounting a persistent volume into /tmp? I think that should solve the issue for now. And later we should think of a better solution that will not store files in any intermediary files in FS Wdyt? @emranemran @thomshutt
> 
> I'll file a ticket so we don't forget. Once the Mist segmenting issues are resolved, we need to revisit this logic to understand how larger files behave and if required, make some changes. I don't think persistent volume would do much here in the worst case - instead we need to refactor to find a way that offsets having to concatenate the segments locally and then uploading.

Good, I've seen you created a Linear issue. Thanks for that!

PR #553, opened 2023-03-30
Dont send videos with rotation to mist

mjh1:
Rotation was found to be ignored by mist so this ensures we only send videos like that to mediaconvert


clients/mediaconvert_test.go
```diff
@@ -291,39 +291,6 @@ func Test_MP4OutDurationCheck(t *testing.T) {
 	}
 }
 
-func TestProbe(t *testing.T) {
```
mjh1, line 294: moved this to probe_test.go, it was missed in a previous refactor

go.mod
```diff
@@ -174,3 +174,5 @@ require (
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace gopkg.in/vansante/go-ffprobe.v2 => github.com/mjh1/go-ffprobe v0.0.0-20230330150500-3b664b71952e
```
thomshutt, line 178: Did you raise a PR against the upstream repo?

mjh1: not yet as i was planning on expanding it to include all the different side data. but through a bit of searching it looks like there may be too much to include in named struct fields :( can't find a definitive list of all variations.
Found this typescript version: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/ffprobe/index.d.ts

But then chatgpt says there are way more options ha:
Screenshot 2023-03-31 at 12 24 06

So might have to be a `map[string]interface{}` kind of thing

thomshutt: Yeah, that sounds like a good idea and then we can leave it to someone else to do the full list if needed

thomshutt (approved):

PR #559, opened 2023-03-30
mist-cleanup: set timeout of script to 15min instead of 5min

emranemran:
Depending on how many pages need to be deleted, this process can take several mins upwards of 5mins to complete. Raising this to 15min instead.


Thulinma (approved):
Yeah, 15m is all good too. 👍

PR #565, opened 2023-04-03
mist-cleanup: disable script since it hogged resources and affected s…

emranemran:
…treaming

### DIFF
--- a/main.go
+++ b/main.go
@@ -19,7 +19,7 @@ import (
 	"github.com/livepeer/catalyst-api/cluster"
 	"github.com/livepeer/catalyst-api/config"
 	mistapiconnector "github.com/livepeer/catalyst-api/mapic"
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -19,7 +19,6 @@ import (
 	"github.com/livepeer/catalyst-api/cluster"
 	"github.com/livepeer/catalyst-api/config"
 	mistapiconnector "github.com/livepeer/catalyst-api/mapic"
-	"github.com/livepeer/catalyst-api/middleware"
```
hjpotter92, line 22: should be just commented out (so revert commit would be just uncommenting, and not searching for what 'middleware' package was

emranemran: Yep good catch, fixed.

hjpotter92 (approved):

thomshutt (approved):

PR #567, opened 2023-04-03
Verify type field is set on Gate API call

mjh1:
This is to add test coverage for the change in bf5f658f

### DIFF
--- a/test/steps/init.go
+++ b/test/steps/init.go
@@ -32,9 +32,7 @@ func (s *StepContext) StartStudioAPI(listen string) error {
 		if err != nil {
 			panic(err)
 		}
-		if gateRequest["type"] == "" {
-			panic(fmt.Errorf("type field should not be empty on gate API request: %+v", gateRequest))
-		}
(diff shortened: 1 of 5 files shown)


test/steps/init.go
```diff
@@ -21,6 +21,18 @@ func (s *StepContext) StartStudioAPI(listen string) error {
...
 	})
 	router.POST("/api/access-control/gate", func(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+		body, err := io.ReadAll(req.Body)
+		if err != nil {
+			panic(err)
+		}
+		gateRequest := make(map[string]interface{})
+		err = json.Unmarshal(body, &gateRequest)
+		if err != nil {
+			panic(err)
```
thomshutt, line 31: Would it work to return an HTTP error response instead? No worries if not, just thinking about the readability of the logs when this happens

mjh1: Added a specific step to check 👍

thomshutt (approved):

PR #573, opened 2023-04-04
mapic: Grab SendAudio from config directly

victorges:
Haven't checked if there are other configs we're
ignoring, but doing this in a rush so this already fixes the known problem.


iameli (approved):
Yep, that looks right

PR #575, opened 2023-04-05
Don't use Mist for large files

thomshutt:
We've been seeing issues with Mist's handling of large files so until that's resolved, let's route those to the backup pipelines.


mjh1 (commented):
LGTM

PR #578, opened 2023-04-06
Skeleton code + tests for segmenting with ffmpeg

thomshutt:
Wanted to get this out there for review so that we get agreement on the pipeline naming etc. and make future PRs smaller and more readable.

On naming - should we just call it the "livepeer" pipeline? We're a little hamstrung by having called the other one the "catalyst" pipeline and wanting to have both of these in the code until we gain confidence in this new one.

### DIFF
--- a/handlers/schemas/UploadVOD.yaml
+++ b/handlers/schemas/UploadVOD.yaml
@@ -19,11 +19,11 @@ properties:
     description:
       Force to use a specific strategy in the Catalyst pipeline. If not
       specified, the default strategy that Catalyst is configured for will be
-      used. This field only available for admin users, and is only used for E2E
-      testing.
+      used. This field only available for admin users.
     enum:
       - catalyst
       - external
+      - ffmpeglivepeer
       - background_external
       - background_mist
       - fallback_external
--- a/handlers/schemas/UploadVOD.yaml
+++ b/handlers/schemas/UploadVOD.yaml
@@ -23,7 +23,7 @@ properties:
     enum:
       - catalyst
       - external
-      - ffmpeglivepeer
+      - catalyst_ffmpeg
       - background_external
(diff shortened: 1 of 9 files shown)


handlers/schemas/UploadVOD.yaml
```diff
@@ -19,11 +19,11 @@ properties:
     description:
       Force to use a specific strategy in the Catalyst pipeline. If not
       specified, the default strategy that Catalyst is configured for will be
-      used. This field only available for admin users, and is only used for E2E
-      testing.
+      used. This field only available for admin users.
     enum:
       - catalyst
       - external
+      - ffmpeglivepeer
```
emranemran, line 26: i think `livepeer` sounds better without having to reference ffmpeg. What if we build some parts of the pipeline with gstreamer in the future?

emranemran (changes requested):

mjh1 (approved):

victorges (commented):
Cool

thomshutt:
@victorges I don't love `catalyst_v2` just because of the mental layer of indirection between name and implementation (plus assumption that one definitely supersedes the other) but I think `catalyst_ffmpeg` is nice - let's go with that

handlers/schemas/UploadVOD.yaml
thomshutt: I don't love catalyst_v2 just because of the mental layer of indirection between name and implementation (plus assumption that one definitely supersedes the other) but I think catalyst_ffmpeg is nice - let's go with that

thomshutt: @emranemran that's a good point, but I think in that case we'll do the same approach of a parallel pipeline that eventually gets merged in / becomes dominant

PR #589, opened 2023-04-11
Makefile: install golangci-lint and godog

iameli:
Needed for a working `make` on first run.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,7 @@ fmt:
 
 .PHONY: integration-test
 integration-test:
+	go install github.com/cucumber/godog/cmd/godog@v0.12.5
 	cd test && godog run --strict --stop-on-failure 2> ./logs/test.log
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -21,10 +21,12 @@ fmt:
 
 .PHONY: integration-test
 integration-test:
+	go install github.com/cucumber/godog/cmd/godog@v0.12.5
 	cd test && godog run --strict --stop-on-failure 2> ./logs/test.log
 
 .PHONY: lint
 lint:
+	go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.52.2
```
thomshutt, line 29: Can we do @latest? Obviously more risk of breakage, but particularly for the linter it'd be good to be up to date

iameli: Sure. Rebased.

thomshutt (approved):

PR #591, opened 2023-04-11
Fix MP4 range requests for access control playback

mjh1:
Seeking in the player was broken because we weren't signaling that we support range requests. I have added an http HEAD handler for this and implemented a custom OPTIONS handler.

### DIFF
--- a/api/http.go
+++ b/api/http.go
@@ -59,15 +59,16 @@ func NewCatalystAPIRouter(cli config.Cli, vodEngine *pipeline.Coordinator, bal b
 	router.GET("/ok", withLogging(catalystApiHandlers.Ok()))
 
 	// Playback endpoint
-	router.GET("/asset/hls/:playbackID/*file",
-		withLogging(
-			withCORS(
-				withGatingCheck(
-					handlers.PlaybackHandler(),
-				),
+	playback := withLogging(
+		withCORS(
+			withGatingCheck(
+				handlers.PlaybackHandler(),
 			),
 		),
 	)
+	router.GET("/asset/hls/:playbackID/*file", playback)
+	router.HEAD("/asset/hls/:playbackID/*file", playback)
+	router.OPTIONS("/asset/hls/:playbackID/*file", withLogging(handlers.PlaybackOptionsHandler()))
 
 	// Handling incoming playback redirection requests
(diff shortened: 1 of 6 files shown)


thomshutt (approved):

handlers/playback.go
```diff
@@ -36,14 +37,32 @@ func PlaybackHandler() httprouter.Handle {
...
 
+		w.Header().Set("accept-ranges", "bytes")
 		w.Header().Set("content-type", response.ContentType)
+		if response.ContentLength != nil {
+			w.Header().Set("content-length", fmt.Sprintf("%d", *response.ContentLength))
+		}
+		w.Header().Set("etag", response.ETag)
+		w.WriteHeader(http.StatusOK)
+
+		if req.Method == "HEAD" {
```
victorges, line 48: **nit:** 
```suggestion
		if req.Method == http.MethodHead {
```

api/http.go
```diff
@@ -59,15 +59,16 @@ func NewCatalystAPIRouter(cli config.Cli, vodEngine *pipeline.Coordinator, bal b
...
+	playback := withLogging(
+		withCORS(
+			withGatingCheck(
+				handlers.PlaybackHandler(),
 			),
 		),
 	)
+	router.GET("/asset/hls/:playbackID/*file", playback)
+	router.HEAD("/asset/hls/:playbackID/*file", playback)
+	router.OPTIONS("/asset/hls/:playbackID/*file", withLogging(handlers.PlaybackOptionsHandler()))
```
victorges, line 71: I think we might need the `withCORS` here as well, wdyt?

handlers/playback.go
```diff
@@ -36,14 +37,32 @@ func PlaybackHandler() httprouter.Handle {
 		}
 		defer response.Body.Close()
 
+		w.Header().Set("accept-ranges", "bytes")
```
victorges, line 40: Should this be set on OPTIONS response as well?

victorges (approved):
nice

api/http.go
mjh1: oh yep, think you're right

handlers/playback.go
mjh1: I don't think it's needed as it's only to check supported http methods right? but wouldn't do any harm i guess

PR #599, opened 2023-04-13
Add transcoding to FFMPEG pipeline

thomshutt:
Still TODO in future PRs:

- Create "fallback" version of this pipeline
- Check if there are any pipeline-specific metrics we want to add
- Check for callbacks in integration tests

### DIFF
--- a/pipeline/coordinator.go
+++ b/pipeline/coordinator.go
@@ -115,11 +115,20 @@ type JobInfo struct {
 	startTime    time.Time
 	result       chan bool
 
-	sourceBytes           int64
-	sourceSegments        int
-	sourceDurationMs      int64
-	sourceCodecVideo      string
-	sourceCodecAudio      string
+	sourceBytes        int64
+	sourceSegments     int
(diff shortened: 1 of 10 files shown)


pipeline/coordinator.go
```diff
@@ -384,6 +393,14 @@ func (c *Coordinator) startOneUploadJob(p UploadJobPayload, handler Handler, for
 		catalystRegion:        os.Getenv("MY_REGION"),
 		sourceCodecVideo:      videoCodec,
 		sourceCodecAudio:      audioCodec,
+		sourceWidth:           videoTrack.Width,
```
emranemran, line 396: Should we just reuse the video.InputVideo struct here? I think most of these fields exist there as well.

emranemran (approved):

pipeline/coordinator.go
thomshutt: Yeah, I was thinking that as I was working on it - it's basically 1:1 now we're getting everything from ffprobe

PR #605, opened 2023-04-14
Switch 'fallback' strategy to FFMPEG->External

thomshutt:
Once we get this onto Staging, it will mean the FFMPEG become the default first pipeline and we can begin testing in earnest.

Still TODO in future PRs: 
- Check if there are any pipeline-specific metrics we want to add
- Check for callbacks in integration tests


mjh1 (commented):
Lgtm

PR #617, opened 2023-04-17
Add support for HLS input in livepeer pipeline

emranemran:
Some useful info: 

* The ffprobe step requires a publicly accessible manifest URL to be able to probe. 
* The mediaconvert pipeline requires manifests to list the full URL path to the segments instead of relative paths. 

Given the above, I thought it's reasonable for us to convert the GCP recordings bucket to a public bucket to keep implementation simple. 

The other alternative is to teach ffprobe to access s3 like URLs (which I think it can) and modify the manifest on the fly for mediaconvert fallback. This alternative felt more error prone so I opted for the simpler method - we do still need to modify the URLs in a few places as you can see in the PR but they're minimal. Suggestions welcome.


pipeline/ffmpeg.go
```diff
@@ -30,26 +30,36 @@ func (f *ffmpeg) HandleStartUploadJob(job *JobInfo) (*HandlerOutput, error) {
...
+		if err := video.Segment(job.SignedSourceURL, destinationURL, job.TargetSegmentSizeSecs); err != nil {
+			return nil, err
+		}
+	} else {
+		// If hls input is detected (i.e. a recording), override the SegmentingTargetURL
+		// to use the SourceFile manifest since segmentation step is skipped.
+		// And also make it an OS compatible URL so that later stages can download it
+		if strings.HasPrefix(job.SourceFile, "http") {
+			job.SegmentingTargetURL = "s3+" + job.SourceFile
+		}
```
emranemran, line 60: I'm not convinced that this is the best place to do this. It feels a bit error prone...

clients/input_copy.go
```diff
@@ -101,6 +101,14 @@ func isDirectUpload(inputFile *url.URL) bool {
 		(inputFile.Scheme == "https" || inputFile.Scheme == "http")
 }
 
+func isHLSInput(inputFile *url.URL) bool {
```
leszko, line 104: Please add unit test for this function.

clients/input_copy.go
```diff
@@ -101,6 +101,14 @@ func isDirectUpload(inputFile *url.URL) bool {
...
+	ext := strings.LastIndex(inputFile.Path, ".")
+	if ext == -1 {
+		return false
+	}
+	return inputFile.Path[ext:len(inputFile.Path)] == ".m3u8"
```
leszko, line 109:
```suggestion
	return inputFile.Path[ext:] == ".m3u8"
```

clients/input_copy.go
```diff
@@ -34,7 +34,7 @@ type InputCopy struct {
 
 // CopyInputToS3 copies the input video to our S3 transfer bucket and probes the file.
 func (s *InputCopy) CopyInputToS3(requestID string, inputFile *url.URL) (inputVideoProbe video.InputVideo, signedURL string, osTransferURL *url.URL, err error) {
```
leszko, line 36: We copy here only the manifest file. Don't we need to copy segment files as well?

transcode/manifest.go
```diff
@@ -176,5 +176,15 @@ func manifestURLToSegmentURL(manifestURL, segmentFilename string) (string, error
 		return "", fmt.Errorf("failed to parse segment filename when converting to segment URL: %s", err)
 	}
 
+	// If a manifest is using URL paths instead of relative paths, then override the
```
leszko, line 179: This is not clear to me, why do we change absolute => relative paths only for Google GCP? What about other S3 storages, AWS, Storj?

leszko (changes requested):
Added my comments.

Other than that, I wonder how difficult it would be to add the non-workaround solution, meaning, to pass the presigned S3 url to `ffprobe`? Then, the HLS input would work for other scenarios, like in Transcode API, for Storj. @thomshutt @emranemran WDYT?

clients/input_copy.go
thomshutt: I think that for V1, we can assume Studio / Recording is the only user of this API if that makes things simpler. I think that then we shouldn't have to worry about copying anything around since we know it'll be on Google Storage

clients/input_copy.go
emranemran: Done

clients/input_copy.go
emranemran: Fixed.

clients/input_copy.go
emranemran: * We're not copying anything here if it's a directupload or hls-input - we can source the manifest directly and parse it and send it to the transcode session. 
* Copying segments around should be out of scope for recordings v2 since it is required for object stores other than GCS. So I submitted VID-144 as a backlog to address this once we ship the first version. Also see my comments below for additional context.

emranemran:
@leszko: answering your qs here since they're all related: 

> Other than that, I wonder how difficult it would be to add the non-workaround solution, meaning, to pass the presigned S3 url to ffprobe? Then, the HLS input would work for other scenarios, like in Transcode API, for Storj. @thomshutt @emranemran WDYT?
It's not difficult but requires refactoring certain bits to handle non-GCS object stores. For these ones, we'll have to copy all segments to GCS first before attempting to transcode. IMO we should address this post-launch of recordings v2 so that we don't get bogged down (similar to what we did for mp4 output support in transcode API). 

> This is not clear to me, why do we change absolute => relative paths only for Google GCP? What about other S3 storages, AWS, Storj?
* For mediaconvert to support HLS input (e.g. fallback scenario), the spec requires a manifest URL with the absolute paths to the segments in that manifest otherwise it fails.
* For our own livepeer transcoding pipeline, we already download the manifest and parse through the relative paths to convert it to an OS compatible URL which then gets passed to the transcode session. 
* So given the above two constraints, it made sense to change *our* HLS input spec to say that the manifest needs to have full URL paths (including when we expand support to Storj, etc). Then we can update this logic here and remove the check for GCP and also handle both cases i.e manifests from segmenting that contain relative paths and manifests from HLS inputs that will contain full paths. In addition, to support all OS, we need to implement the copy to GCS logic for all segments and then modify the manifests and the paths and then submit for transcoding -- this was getting out of scope to get recordings out the door. So for now, we're assuming it will be GCP only for the recordings workflow.

clients/input_copy.go
```diff
@@ -95,6 +92,89 @@ func (s *InputCopy) CopyInputToS3(requestID string, inputFile *url.URL) (inputVi
...
 
+func isHLSInput(inputFile *url.URL) bool {
+	ext := strings.LastIndex(inputFile.Path, ".")
+	if ext == -1 {
+		return false
+	}
+	return inputFile.Path[ext:] == ".m3u8"
+}
+
+func getDestLoc(srcManifestUrl, dstTransferUrl *url.URL, srcSegmentUrl string) string {
```
emranemran, line 103: @leszko, lmk if there's a cleaner way for this implementation - it works but I think it could be simpler and I got lost in the world of golang's url lib and resorted to a very basic implementation. 

As an example, this is what this function is supposed to do given the following args: 

srcManifestUrl := `https://link.storjshare.io/raw/jwbmejfygkb6mdrarfplvc46zeka/cakesource/rec/b40efldl0rzdcm7b/2023.04.19.14.04.01/output.m3u8`

srcSegmentUrl := `https://link.storjshare.io/raw/jwbmejfygkb6mdrarfplvc46zeka/cakesource/rec/b40efldl0rzdcm7b/2023.04.19.14.04.01/blah/hello/world/3.ts`

dstTransferUrl := `s3+https://USER:PASS@storage.googleapis.com/lp-us-catalyst-vod-monster/hls/source/hcfecbea/transfer/output.m3u8`

The function should return the target transfer directory where *each* segment from the storj src URL can be copied to. In this case, the function should return: 
```s3+https://GOOG1ECL5LAFILBTTALJ6EFZTFJQSBC6QVRJWXEKROJW6Y2R7RZ25WPE2VNVA:foWv%2F3xcMT1+q4g1SLDCYtJ47y3USZ%2FMKGD+HmbK@storage.googleapis.com/lp-us-catalyst-vod-monster/hls/source/hcfecbea/transfer/blah/hello/world/3.ts```

**Comment by emranemran** [comment]
**Updates**
* Got rid of the original changes changes to only handle backup-recordings and updated the PR to support generic HLS input in catalyst-api
* TODOs: 
** cleanup of code + comments
** add/fix tests

File: clients/input_copy.go [hunk 7]
```diff
@@ -95,6 +92,89 @@ func (s *InputCopy) CopyInputToS3(requestID string, inputFile *url.URL) (inputVi
...
+	p := url.QueryEscape(pass)
+
+	parsedURL3.User = url.UserPassword(u, p)
+
+	newURL := parsedURL3.Scheme + "://" + u + ":" + p + "@" + parsedURL3.Host + dir + "/" + relPath
+
+	return newURL
+}
+
+func copyFile(requestID string, srcInputUrl, dstOutputUrl *url.URL) (size int64, err error) {
```
leszko, line 138: Will this also work if the input file is the `s3+https://` URL. E.g. `s3+https://<ACCESS_KEY>:<SECRET_KEY>@gateway.storjshare.io/testinput/video.mp4`

leszko (commented):
@emranemran I tried the following url and it didn't work, could you check?

URL: `https://storage.googleapis.com/lp-us-catalyst-recordings-monster/hls/adfdu549xe1wiah0/ec2e3ab2-8889-4474-9430-823bf12d81b2/output.m3u8`

Result:
```
  "error": "failed to create signed url for https://storage.googleapis.com/lp-us-catalyst-recordings-monster/hls/adfdu549xe1wiah0/ec2e3ab2-8889-4474-9430-823bf12d81b2/source/0.ts: failed to parse OS url: unrecognized OS scheme: https",
```

clients/input_copy.go
```diff
@@ -95,6 +92,89 @@ func (s *InputCopy) CopyInputToS3(requestID string, inputFile *url.URL) (inputVi
...
+	ext := strings.LastIndex(inputFile.Path, ".")
+	if ext == -1 {
+		return false
+	}
+	return inputFile.Path[ext:] == ".m3u8"
+}
+
+func getDestLoc(srcManifestUrl, dstTransferUrl *url.URL, srcSegmentUrl string) string {
+
+	parsedURL1, _ := url.Parse(srcManifestUrl.String())
```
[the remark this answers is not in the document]
emranemran: Yep, still on my TODO list.

clients/input_copy.go
emranemran: I'll test and confirm but it should b/c any mp4 input will not go through any of the new code paths being introduced here.

emranemran: Yes, this will work as long as the s3+https url has user creds.

clients/input_copy.go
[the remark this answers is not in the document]
emranemran: Done

clients/input_copy.go
```diff
@@ -95,6 +92,89 @@ func (s *InputCopy) CopyInputToS3(requestID string, inputFile *url.URL) (inputVi
...
+		}
+	}
+
+	// Extract the relative path by removing the common prefix
+	relPath := path2[i:]
+	if strings.HasPrefix(relPath, "/") {
+		relPath = relPath[1:]
+	}
+
+	parsedURL3, _ := url.Parse(dstTransferUrl.String())
```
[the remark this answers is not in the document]
emranemran: Done

clients/input_copy.go
emranemran: Done

clients/input_copy.go
emranemran: The code was updated to handle generic HLS input in catalyst.

emranemran:
Updates: 
* addressed all comments. 

> @emranemran I tried the following url and it didn't work, could you check?
Works now and tested with the exact URL.

clients/input_copy.go
```diff
@@ -95,9 +91,101 @@ func (s *InputCopy) CopyInputToS3(requestID string, inputFile *url.URL) (inputVi
...
+	}
+	return inputFile.Path[ext:] == ".m3u8"
+}
+
+// Given a source manifest URL (e.g. https://storage.googleapis.com/foo/bar/output.m3u8) and
+// a source segment URL (e.g. https://storate.googleapis.com/foo/bar/0.ts), generate a target
+// OS-compatible transfer URL for each segment that uses the destination transfer URL for the source manifest
+// (e.g. if destination transfer URL is:
+// https://USER:PASS@storage.googleapi.com/hello/world/transfer/output.m3u8
+// then detination transfer URL for each segment will be:
```
leszko, line 107:
```suggestion
// then destination transfer URL for each segment will be:
```

leszko (approved):
LGTM 🎉 

Tested with the Studio PR and it worked fine (after some changes in task-runner).

PR #620, opened 2023-04-21
Remove the Mist VOD pipeline

thomshutt:
After a week with the FFMPEG pipeline running in the foreground, I don't think we're going to be going back to this so I'm removing it to make the codebase more manageable. It'll still be in the Git history if we ever need to refer back to it.


pipeline/coordinator.go
```diff
@@ -28,16 +28,12 @@ import (
 type Strategy string
 
 const (
-	// Only execute the Catalyst (Mist) pipeline.
-	StrategyCatalystDominance Strategy = "catalyst"
-	// Only execute the external pipeline.
+	// Only execute the external pipeline
 	StrategyExternalDominance Strategy = "external"
 	// Only execute the FFMPEG / Livepeer pipeline
 	StrategyCatalystFfmpegDominance Strategy = "catalyst_ffmpeg"
```
leszko, line 34: I'm not sure we need to have `catalyst_ffmpeg`, maybe we should just keep the old name `catalyst`?

If you decide to keep `catalyst_ffmpeg`, then please update the related configs here: https://github.com/livepeer/catalyst/tree/main/config

leszko (approved):
Good work @thomshutt 🚀 Really a great improvement with `ffmpeg` and it's great to look at the PRs removing so much code.

Added one comments, other then that LGTM.

config/cli.go
```diff
@@ -22,7 +22,6 @@ type Cli struct {
 	MistUser                  string
 	MistPassword              string
 	MistPort                  int
-	MistHttpPort              int
```
mjh1, line 25: Will anything break if the infra is still trying to pass this param?

mjh1 (approved):

config/cli.go
thomshutt: Good question, I'll check on Staging and add it back in if needed

thomshutt: Looks like we weren't actually using it anywhere

PR #623, opened 2023-04-24
Fix callback stale check

mjh1:
Previously we were comparing seconds to milliseconds so stale callback messages were essentially never timing out with this check. I've fixed the check by switching to `time.Time` types to make it clearer.

### DIFF
--- a/clients/callback_client.go
+++ b/clients/callback_client.go
@@ -15,7 +15,7 @@ import (
 	"github.com/livepeer/catalyst-api/metrics"
 )
 
-const MAX_TIME_WITHOUT_UPDATE = 30 * time.Minute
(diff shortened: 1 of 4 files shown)


mjh1:
@thomshutt damn just realised we may need to increase this from 30 mins right? I think you noticed earlier the copy step can take a long time without the app updating the status?

victorges:
> we may need to increase this from 30 mins right? I think you noticed earlier the copy step can take a long time without the app updating the status?

One alternative is reporting progress during the copy step as well. On `task-runner` we have a `Reader` wrapper that reports progress (as long as it knows the full size): https://github.com/livepeer/task-runner/blob/4f2b0514ec4e4c5dc9e50d6e20e2dcfb82d6af1e/task/upload.go#L442

PR #626, opened 2023-04-25
[SRE-89] Add geolocation redirect handler for `/webrtc` paths

hjpotter92:


iameli (approved):
LGTM but let's cross-reference [VID-48] here which would be to implement a generic way to support all Mist protocols here

PR #630, opened 2023-04-26
Probe segments coming out of the ffmpeg segmenting process

mjh1:
We need to do this to verify the segments are not corrupt in any way. We found with one input video this was happening (playbackID 97b1rnzi74ps1g58)

### DIFF
--- a/pipeline/ffmpeg.go
+++ b/pipeline/ffmpeg.go
@@ -117,6 +117,18 @@ func (f *ffmpeg) HandleStartUploadJob(job *JobInfo) (*HandlerOutput, error) {
 
 	job.sourceSegments = len(sourceManifest.GetAllSegments())
 
+	sourceSegmentURLs, err := transcode.GetSourceSegmentURLs(transcodeRequest.SourceManifestURL, sourceManifest)
+	for _, seg := range sourceSegmentURLs {
+		probeURL, err := clients.SignURL(seg.URL)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create signed url for %s: %w", seg.URL, err)
+		}
+		_, err = video.Probe{}.ProbeFile(probeURL)
+		if err != nil {
+			return nil, fmt.Errorf("probe failed for segment %s: %w", seg.URL, err)
+		}
+	}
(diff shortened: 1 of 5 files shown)


mjh1:
Tested the problematic video on staging and verified the fix 👍 it plays fine if it goes through the mediaconvert fallback: https://storage.googleapis.com/livepeer-test-vod/test-max/vod/2023-04-26T10%3A34%3A52.181Z/index.m3u8

pipeline/ffmpeg.go
```diff
@@ -135,6 +135,21 @@ func (f *ffmpeg) HandleStartUploadJob(job *JobInfo) (*HandlerOutput, error) {
 
 	job.sourceSegments = len(sourceManifest.GetAllSegments())
 
+	sourceSegmentURLs, err := transcode.GetSourceSegmentURLs(transcodeRequest.SourceManifestURL, sourceManifest)
+	if err != nil {
+		return nil, fmt.Errorf("error fetching source segment URLs: %w", err)
+	}
+	for _, seg := range sourceSegmentURLs {
```
thomshutt, line 142: Do we need to probe all of them? Could we maybe do the first and last or something?

pipeline/ffmpeg.go
```diff
@@ -115,7 +116,20 @@ func (f *ffmpeg) HandleStartUploadJob(job *JobInfo) (*HandlerOutput, error) {
...
+	job.sourceSegments = len(sourceSegments)
+
+	if job.sourceSegments > 0 {
+		firstSeg := sourceSegments[0]
+		lastSeg := sourceSegments[job.sourceSegments-1]
+
+		if err := probeSourceSegment(firstSeg, transcodeRequest.SourceManifestURL); err != nil {
+			return nil, err
+		}
+		if err := probeSourceSegment(lastSeg, transcodeRequest.SourceManifestURL); err != nil {
```
thomshutt, line 129: We could be more efficient and check it's not a single segment video before doing this, but let's 🚢 🇮🇹  for now

thomshutt (approved):

PR #631, opened 2023-04-26
Clean up local source files on startup

thomshutt:
This PR is a followup to https://github.com/livepeer/catalyst-api/pull/625 and deals with the case where an app restart (e.g following a deploy) happens while we're mid-segmenting, which will mean that the post-segmenting cleanup code never gets run.

It also splits the segmenting out into its own method to allow the regular `defer` for cleanup happen sooner (currently it happens after transcoding is also complete).

### DIFF
--- a/pipeline/ffmpeg.go
+++ b/pipeline/ffmpeg.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/url"
 	"os"
+	"path/filepath"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pipeline/ffmpeg.go
```diff
@@ -15,11 +16,21 @@ import (
 	"github.com/livepeer/catalyst-api/video"
 )
 
+const LocalSourceFilePattern = "sourcevideo*"
```
thomshutt, line 19: I changed the pattern here from `source*` to `sourcevideo*` just to try and reduce the chance of catching some other unrelated file that's in the tmp directory

thomshutt: This does mean that following the deploy where this goes out, we should probably do manual cleanup of any files with the old `source*` pattern

mjh1 (approved):

PR #632, opened 2023-04-26
Don't send a duplicate Completion callback

thomshutt:
This is sent by the coordinator with all information that Studio needs, so we shouldn't be sending it here

### DIFF
--- a/pipeline/ffmpeg.go
+++ b/pipeline/ffmpeg.go
@@ -125,9 +125,6 @@ func (f *ffmpeg) HandleStartUploadJob(job *JobInfo) (*HandlerOutput, error) {
 
 	job.transcodedSegments = transcodedSegments
 
-	// Transcoding Finished
-	job.ReportProgress(clients.TranscodeStatusCompleted, 1)
-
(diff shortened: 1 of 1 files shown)


pipeline/ffmpeg.go
```diff
@@ -125,9 +125,6 @@ func (f *ffmpeg) HandleStartUploadJob(job *JobInfo) (*HandlerOutput, error) {
 
 	job.transcodedSegments = transcodedSegments
 
-	// Transcoding Finished
-	job.ReportProgress(clients.TranscodeStatusCompleted, 1)
```
victorges, line 129: Hmm nice catch. Maybe what we wanted here was progress 1 but status just "transcoding"?

victorges (approved):
Lgtm

pipeline/ffmpeg.go
thomshutt: Yeah, I think that will actually happen inside the Transcode loop already though. This was really just a leftover from me PoC-ing the ffmpeg pipeline

PR #638, opened 2023-04-28
mapic: Add STREAM_BUFFER handling for stream health reporting

victorges:
Basically works by parsing the Mist STREAM_BUFFER payload into a single
JSON object and pushing it to a "stream heath hook url" configured on cli.
The Studio API will handle updating the stream health with what comes in
that trigger (e.g. can easily start handling the END trigger soon).

This was meant to be catalyst-api code, but in the midst of the implementation
I realized it has no connection with Mist anymore, so it's easier to plug this into 
mist-api-connector instead for now (especially due to code/config familiarity).
Nice that it's all in the same repo which makes it easier. I also implemented it
in a way that makes sense for catalyst-api (no hard deps on Studio API) so
it's easy to migrate in the future.

### DIFF
--- a/mapic/misttriggers/stream_buffer.go
+++ b/mapic/misttriggers/stream_buffer.go
@@ -6,7 +6,6 @@ import (
 	"net/http"
 	"strings"
 
-	"github.com/livepeer/catalyst-api/errors"
 	"github.com/livepeer/catalyst-api/log"
 )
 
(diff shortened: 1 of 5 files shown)


mapic/misttriggers/stream_buffer.go
```diff
@@ -0,0 +1,192 @@
...
+		return nil, nil
+	}
+
+	var tracksAndIssues map[string]interface{}
+	err := json.Unmarshal(data, &tracksAndIssues)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing stream details JSON: %w", err)
+	}
+
+	issues, ok := tracksAndIssues["issues"].(string)
```
mjh1, line 175: I'm wondering if we need to check the key exists first before checking type, does this not panic with a nil pointer?

mjh1 (approved):
Cucumber tests seem to be failing but otherwise lgtm

mapic/misttriggers/stream_buffer.go
victorges: I had that doubt myself, but it does work! Map fetches will never panic, they just return the zero value for the type, while interface casts work on completely `nil` interfaces _as long as you use the 2 value ok syntax_. 

For curiosity: The only caveat is when the `interface{}` is not `nil` but holds a nullable-type and a `nil` value. Then you could need to check if the value itself is also not nil additionally from the `ok` (and you can't use single-value syntax or that would explode)

playground: https://go.dev/play/p/WcZ_u4puwfW

mapic/misttriggers/stream_buffer_test.go
```diff
@@ -109,7 +109,8 @@ func TestTriggerStreamBufferE2E(t *testing.T) {
 		err := json.NewDecoder(r.Body).Decode(&receivedPayload)
 		if err != nil {
 			w.WriteHeader(http.StatusBadRequest)
-			w.Write([]byte("error unmarshalling payload"))
+			_, err := w.Write([]byte("error unmarshalling payload"))
+			require.NoError(t, err)
```
victorges, line 113: Thanks!

PR #642, opened 2023-05-01
signed multistream targets

iameli:
Opening draft now; completed PR ready for review tomorrow.

Tech debt incurred:
* I'm including go-livepeer as a library to make use of its `eth.AccountManager`. This mandates `CGO_ENABLED=0` to avoid linking ffmpeg, which is in turn required by `go test -race` so that's off too. Possible fixes: refactor on go-livepeer to split out ffmpeg packages from non-ffmpeg packages, move eth.AccountManager and its supporting libraries to livepeer/go-tools, or rewrite entirely on this end.


cluster/cluster.go
```diff
@@ -234,21 +246,57 @@ func ResolveNodeURL(c Cluster, streamURL string) (string, error) {
 	return u2.String(), nil
 }
 
+func (c *ClusterImpl) BroadcastEvent(signed *events.SignedEvent) error {
+	unverified, err := signed.UnverifiedEvent()
+	payload, err := json.Marshal(unverified)
+	if err != nil {
+		return fmt.Errorf("error marshalling event payload: %w", err)
+	}
+	err = c.serf.UserEvent("UserAction", payload, false)
```
iameli, line 255: Should be a constant.

cluster/cluster.go
```diff
@@ -234,21 +246,57 @@ func ResolveNodeURL(c Cluster, streamURL string) (string, error) {
...
+	if err != nil {
+		return fmt.Errorf("error broadcasting serf event: %w", err)
+	}
+	return nil
+}
+
+func (c *ClusterImpl) handleUserAction(evt serf.UserEvent) {
+	err := func() error {
+		if evt.Name != "UserAction" {
```
iameli, line 264: Constant.

cluster/cluster.go
```diff
@@ -234,21 +246,57 @@ func ResolveNodeURL(c Cluster, streamURL string) (string, error) {
...
 				return
-			case e := <-c.eventCh:
-				select {
-				case <-ctx.Done():
-					return
-				case inbox <- e:
-					// Event is now in the inbox
-				default:
-					// Overflow event gets dropped
+			case e := <-c.serfCh:
```
iameli, line 287: Yuck, that's a lot of nesting. Refactor.

events/actions.go
```diff
@@ -0,0 +1,42 @@
...
+	"encoding/json"
+)
+
+type Action interface {
+	Type() string
+	SignerAddress() string
+}
+
+// Exports this action to a map
+func ActionToMap(a any) (map[string]any, error) {
```
iameli, line 13: The implementations here are kinda funny, but the more I think about it they're kind of right... I mean, there's a technically-unnecessary JSON serialization, but we do want to populate the structs based on the JSON tags os this is _almost_ correct behavior.

events/events_test.go
```diff
@@ -0,0 +1 @@
+package events
```
iameli, line 1: Need some.

main.go
```diff
@@ -216,6 +240,26 @@ func reconcileBalancer(ctx context.Context, bal balancer.Balancer, c cluster.Clu
 	}
 }
 
+// Validate new events and pass them to state machine
+func reconcileMachine(ctx context.Context, c cluster.Cluster, signer events.Signer, machine *state.Machine) error {
```
leszko, line 244: 1. Where do you distribute the state (or replay the events) when a new Catalyst node joins? How do you make it aware of the current state in the cluster?

2. Do we have any persistence for the state machine? What if all Catalyst nodes are down?

iameli-streams:
https://github.com/livepeer/catalyst-api/assets/136638730/1f71068a-0396-43c2-b870-95a6ad644ffb

PR #645, opened 2023-05-02
mapic: FIx stream_buffer payload parsing

victorges:
Turns out that the Mist docs I used were outdated. Instead of having only the `issues` field
additional on that JSON, we actually have some other int/float metrics that need to be removed
as well so we can parse the rest as track objects. 

Updated the code with the assumption now that any object field is a track, and the rest are
either issues or "extra" fields we don't really care. Still include those in a separate `extra`
object for logging on the API.

For ctx, this is the error the code is getting now: `error parsing stream details tracks: json: cannot unmarshal number into Go value of type misttriggers.TrackDetails`
Which happens because we fields like this in the root object of the payload:
```
  "buffer": 34667,
  "jitter": 51,
  "maxkeepaway": 7500,
```

Also found out that, when issues are present, the actual human friendly message goes under a `human_issues` field,
not just `issues`. Fixed that as well and passing it to the API.

### DIFF
--- a/mapic/misttriggers/stream_buffer.go
+++ b/mapic/misttriggers/stream_buffer.go
@@ -213,7 +213,7 @@ func ParseMistStreamDetails(streamState string, data []byte) (*MistStreamDetails
 
 	var tracks map[string]TrackDetails
 	if err = json.Unmarshal(tracksJSON, &tracks); err != nil {
-		return nil, fmt.Errorf("eror parsing stream details tracks: %w", err)
+		return nil, fmt.Errorf("error parsing stream details tracks: %w", err)
 	}
 
 	return &MistStreamDetails{tracks, issues, humanIssues, extra}, nil
--- a/mapic/misttriggers/stream_buffer.go
+++ b/mapic/misttriggers/stream_buffer.go
@@ -70,6 +70,8 @@ func TriggerStreamBuffer(cli *config.Cli, req *http.Request, lines []string) err
 		streamHealth.IsHealthy = streamHealth.IsHealthy && details.Issues == ""
 		streamHealth.Tracks = details.Tracks
 		streamHealth.Issues = details.Issues
+		streamHealth.HumanIssues = details.HumanIssues
+		streamHealth.Extra = details.Extra
 	}
 
 	err = PostStreamHealthPayload(cli.StreamHealthHookURL, cli.APIToken, streamHealth)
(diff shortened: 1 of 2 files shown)


mapic/misttriggers/stream_buffer.go
```diff
@@ -82,12 +84,16 @@ func TriggerStreamBuffer(cli *config.Cli, req *http.Request, lines []string) err
...
+	StreamName string `json:"stream_name"`
+	SessionID  string `json:"session_id"`
+	IsActive   bool   `json:"is_active"`
+
+	IsHealthy   bool   `json:"is_healthy"`
+	Issues      string `json:"issues,omitempty"`
+	HumanIssues string `json:"human_issues,omitempty"`
+
+	Tracks map[string]TrackDetails `json:"tracks,omitempty"`
+	Extra  map[string]interface{}  `json:"extra,omitempty"`
```
iameli, line 96: I enjoy spelling this like
```suggestion
	Extra  map[string]any  `json:"extra,omitempty"`
```

mapic/misttriggers/stream_buffer_test.go
```diff
@@ -13,13 +13,13 @@ import (
 
 var (
 	streamBufferPayloadFull = []string{
-		"stream1", "FULL", `{"track1":{"codec":"h264","kbits":1000,"keys":{"B":"1"},"fpks":30,"height":720,"width":1280}}`,
+		"stream1", "FULL", `{"track1":{"codec":"h264","kbits":1000,"keys":{"B":"1"},"fpks":30,"height":720,"width":1280},"jitter":420}`,
 	}
 	streamBufferPayloadIssues = []string{
-		"stream1", "RECOVER", `{"track1":{"codec":"h264","kbits":1000,"keys":{"B":"1"},"fpks":30,"height":720,"width":1280},"issues":"Stream is feeling under the weather"}`,
+		"stream1", "RECOVER", `{"track1":{"codec":"h264","kbits":1000,"keys":{"B":"1"},"fpks":30,"height":720,"width":1280},"issues":"The aqueous linear entity, in a manner pertaining to its metaphorical state of existence, appears to be experiencing an ostensibly suboptimal condition that is reminiscent of an individual's disposition when subjected to an unfavorable meteorological phenomenon","human_issues":"Stream is feeling under the weather"}`,
```
iameli, line 19: madlad

mapic/misttriggers/stream_buffer.go
iameli: Yeah it was introduced with generics, `any` is just an alias for `interface{}`

mapic/misttriggers/stream_buffer.go
```diff
@@ -82,12 +84,16 @@ func TriggerStreamBuffer(cli *config.Cli, req *http.Request, lines []string) err
...
-	Issues     string                  `json:"issues,omitempty"`
+	StreamName string `json:"stream_name"`
+	SessionID  string `json:"session_id"`
+	IsActive   bool   `json:"is_active"`
+
+	IsHealthy   bool   `json:"is_healthy"`
+	Issues      string `json:"issues,omitempty"`
+	HumanIssues string `json:"human_issues,omitempty"`
+
+	Tracks map[string]TrackDetails `json:"tracks,omitempty"`
```
iameli, line 95: Not blocking release, but I kind of think we shouldn't in general use `json.Unmarshal` with stuff coming out of Mist; I don't like the idea of them adding a field and our code breaks. Unless there's a way to signify `ignoreExtra` to the JSON unmarhsaller

iameli (approved):
Just nits, looks fine

mapic/misttriggers/stream_buffer.go
victorges: JSON unmarshalling does ignore extra fields! It just doesn't ignore type mismatches, which is why the previous code was breaking (trying to parse a number as an object). Jaron guaranteed that the only "object fields" that we'll have in this payload will be tracks, and that they make sure not to break compatibility with existing clients, so I believe we are safe now!

iameli: ohhhh cool

PR #649, opened 2023-05-03
handler: vod-decryption: added public key endpoint

gioelecerati:


config/cli.go
```diff
@@ -51,6 +51,8 @@ type Cli struct {
 	Tags                      map[string]string
 	RetryJoin                 []string
 	EncryptKey                string
+	VodEncryptPublicKey       string
+	VodEncryptPrivateKey      string
```
victorges, line 55: **nit:** IMO either call this `VodDecrypt...` or `VodEncryption...`. Catalyst-api doesn't really encrypt anything on this current version.

handlers/accesscontrol/encryption.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+func (ec *EncryptionHandlersCollection) PublicKeyHandler() httprouter.Handle {
+	return func(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusOK)
+
+		responseData := map[string]string{
+			"publicKey": ec.publicKey,
+			"nodeName":  ec.nodeName,
```
victorges, line 30: I think catalyst-api uses `snake_case` so we should be consistent here
```suggestion
			"public_key": ec.publicKey,
			"node_name":  ec.nodeName,
```

victorges (commented):
Looking good

clients/input_copy.go
```diff
@@ -33,14 +38,33 @@ type InputCopy struct {
...
+func (s *InputCopy) CopyInputToS3(requestID string, inputFile *url.URL, encryptedKey string, vodEncryptPrivateKey string) (inputVideoProbe video.InputVideo, signedURL string, osTransferURL *url.URL, err error) {
+
+	// Create a temporary local file to write to
+	localSourceFile, err := os.CreateTemp(os.TempDir(), LocalSourceFilePattern)
+	if err != nil {
+		glog.Errorf("failed to create local file on vod upload: %s", err)
+		return
+	}
+	defer localSourceFile.Close()
+	defer os.Remove(localSourceFile.Name()) // Clean up the file as soon as we're done segmenting
```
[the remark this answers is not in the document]
victorges: I think that ideally we would refactor the file downloading to be in a single place instead, on the outermost layer (runner) and make sure that we create&delete a single local file for the job. This would help reduce all the repeated downloading of the file that seems to happen today inside catalyst (only in this context: download once for s3 copy, once for probe, once for ffmpeg segmenting).

If that is too complex of a refactor to make now, I thought that this current implementation is _already_ an improvement over the previous state: instead of downloading the file once for copying to s3 and once for probing it, we download only once and use it for both (with the added decryption). So it's already an improvement. 

I think this setup is also better than another download&upload just to decrypt the file, followed by other downloads to copy to s3, probe, segment etc. So the fact that we are not re-using the file from here for segmenting is a separate issue to be optimized later, but it is not a regression related to the previous state (it seems more like an improvement, correct me if I'm wrong).

WDYT?

crypto/aes.go
```diff
@@ -0,0 +1,155 @@
...
+	encryptedKey, err := base64.StdEncoding.DecodeString(encryptedKeyB64)
+	if err != nil {
+		glog.Fatalf("Error decoding base64 encoded key: %v", err)
+	}
+
+	// Decrypt the key with the RSA private key
+	key, err := rsa.DecryptOAEP(sha256.New(), rand.Reader, privateKey, encryptedKey, nil)
+
+	if err != nil {
+		log.Fatalf("Error decrypting key: %v", err)
```
victorges, line 73: reminder to change this into a returned error. We shouldn't crash the app on a job

crypto/decrypt.go
```diff
@@ -0,0 +1,65 @@
...
+	"crypto/rsa"
+	"crypto/sha256"
+	"encoding/base64"
+	"fmt"
+	"os"
+
+	"github.com/golang/glog"
+)
+
+func pkcs7Unpad(data []byte) ([]byte, error) {
```
victorges, line 16: I saw that the lib we're using in the other implementation does a couple more checks on the padding block, like [if the padded bytes have the right value](https://github.com/d1str0/pkcs7/blob/master/pkcs7.go#L83). Would be good to add some of those checks to this impl, if we want to avoid the external dependency for the simple logic.

crypto/aes.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	return DecryptAESCBCWithIV(reader, privateKey, encryptedKeyFile, iv)
+}
+
+func DecryptAESCBCWithIV(reader io.ReadCloser, privateKey *rsa.PrivateKey, encryptedKeyB64 string, iv []byte) (io.ReadCloser, error) {
+
+	encryptedKey, err := base64.StdEncoding.DecodeString(encryptedKeyB64)
+	if err != nil {
+		glog.Fatalf("Error decoding base64 encoded key: %v", err)
```
victorges, line 66: Also needs to be a returned error. Fatal will crash the process. Might be worth going round the PR diff to check if there aren't any other fatal logs like these

clients/input_copy.go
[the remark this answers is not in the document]
mjh1: Ah ok, I see what you're saying but the only thing to highlight is that the probing happens with range requests so it only has to pull down small amounts of the file, so for direct uploads with this code we're now copying the whole file to disk which would make performance worse. If it's a direct upload and there's no encryption we should avoid the copy to disk here I think.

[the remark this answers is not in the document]
victorges: I see, that makes sense. What do you think should be the right approach here then? Maybe deal with the encryption as part of the "copy to s3" part as streaming decryption? So the condition for copying to S3 becomes "is not a direct upload or has encryption". It would only affect (more copies) in the case of "direct upload + encryption", in which the file is already on s3 and we're streaming it back and up only for decryption. This shouldn't be too bad tho since I think it will be the rarest kind of input (most of the times it will be encrypted in some dStorage).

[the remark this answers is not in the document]
mjh1: > I see, that makes sense. What do you think should be the right approach here then? Maybe deal with the encryption as part of the "copy to s3" part as streaming decryption? So the condition for copying to S3 becomes "is not a direct upload or has encryption". It would only affect (more copies) in the case of "direct upload + encryption", in which the file is already on s3 and we're streaming it back and up only for decryption. This shouldn't be too bad tho since I think it will be the rarest kind of input (most of the times it will be encrypted in some dStorage).

Yeah that sounds right to me 👍

crypto/aes.go
```diff
@@ -0,0 +1,154 @@
...
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+
+	"github.com/d1str0/pkcs7"
+	"github.com/golang/glog"
+)
+
+func LoadPrivateKey(privateKeyBase64 string) (*rsa.PrivateKey, error) {
```
victorges, line 22: This function (and any other rsa helpers, which might be useful) should probably go on a separate `rsa.go` file. Or we can rename this file to just `decryption.go` or sth

pipeline/coordinator.go
```diff
@@ -69,12 +69,17 @@ type UploadJobPayload struct {
...
 	GenerateMP4           bool
+	Encryption            *EncryptionPayload
 	InputFileInfo         video.InputVideo
 	SignedSourceURL       string
 	InFallbackMode        bool
 	LivepeerSupported     bool
 }
 
+type EncryptionPayload struct {
+	EncryptedKey string
```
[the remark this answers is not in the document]
gioelecerati: They encryption key itself is encrypted with the catalyst public key and sent by the user!

clients/input_copy.go
```diff
@@ -213,6 +274,25 @@ func CopyFile(ctx context.Context, sourceURL, destOSBaseURL, filename, requestID
 	return
 }
 
+func CopyReaderFile(ctx context.Context, readerFile io.Reader, destOSBaseURL, filename, requestID string) (writtenBytes int64, err error) {
+	err = backoff.Retry(func() error {
```
victorges, line 278: This `retry` will not work since you're using the same reader/stream instance. It's possible that it fails the first time and then it only uploads a partial file on the next attempt.

So you must have the download+decryption logic inside the retry, which is why I suggested you pass in a lambda func to handle it transparently for the copying logic here.

clients/input_copy.go
```diff
@@ -63,8 +70,62 @@ func (s *InputCopy) CopyInputToS3(requestID string, inputFile *url.URL) (inputVi
...
+		if err != nil {
+			err = fmt.Errorf("failed to copy file(s): %w", err)
+			return
+		}
+		log.Log(requestID, "Copied", "bytes", size, "source", inputFile.String(), "dest", decryptedFileUrl)
+
+		signedURL, err = SignURL(osTransferURL)
+		if err != nil {
+			return
+		}
```
victorges, line 108: Would be nice to reduce the code repetition here with the S3 copy below. It could be just 1 `if` on line 114 like
```go
if encryptedKey != "" {
  size, err = CopyFile(context.Background(), sourceOutputURL.String(), osTransferURL.String(), "", requestID, decryptFileReader(encryptedKey))
} else {
  size, err = CopyFile(context.Background(), sourceOutputURL.String(), osTransferURL.String(), "", requestID, nil) 
}
```

clients/input_copy.go
```diff
@@ -63,8 +70,62 @@ func (s *InputCopy) CopyInputToS3(requestID string, inputFile *url.URL) (inputVi
...
+	}
+
+	if !isDirectUpload(inputFile) {
```
victorges, line 111: This condition is not correct since the file may have already been copied above. In case of non-direct uploads (most of the encryption use case) we will copy the decrypted file to S3, then copy it again, but encrypted. Overriding the decryption work.

I think the ideal fix here would be moving the decryption work to this code block below and have a simple condition here 
`!isDirectUpload(inputFile) || encryptionKey != ""`

clients/input_copy.go
```diff
@@ -63,8 +70,62 @@ func (s *InputCopy) CopyInputToS3(requestID string, inputFile *url.URL) (inputVi
...
+	if encryptedKey != "" {
+		c, e := getFile(context.Background(), requestID, inputFile.String())
+
+		if e != nil {
+			glog.Errorf("error getting file: %w", err)
+			return
+		}
+
+		var decryptionKey *rsa.PrivateKey
+		if decryptionKey, err = crypto.LoadPrivateKey(VodDecryptPrivateKey); err != nil {
```
victorges, line 82: nit: How expensive is this loading of the private key? We have it a static config, so ideally should only need to load it once right?

crypto/decryption.go
```diff
@@ -0,0 +1,147 @@
...
+	// Decrypt the key with the RSA private key
+	key, err := rsa.DecryptOAEP(sha256.New(), rand.Reader, privateKey, encryptedKey, nil)
+
+	if err != nil {
+		glog.Errorf("Error decrypting key: %v", err)
+	}
+
+	if err != nil {
+		return nil, fmt.Errorf("error decoding base16 key: %w", err)
+	}
```
victorges, line 70:
```suggestion
	if err != nil {
		glog.Errorf("Error decrypting key: %v", err)
		return nil, fmt.Errorf("error decoding base16 key: %w", err)
	}
```

victorges (commented):
Still need some fixes on the streaming logic, but almost there!

victorges (commented):
Still need some fixes on the streaming logic, but almost there!

victorges (commented):
Still need some fixes on the streaming logic, but almost there!

victorges (commented):
Still need some fixes on the streaming logic, but almost there!

clients/input_copy.go
```diff
@@ -32,9 +36,24 @@ type InputCopy struct {
...
 
+type DecryptionKeys struct {
+	decryptKey   rsa.PrivateKey
+	encryptedKey string
+}
+
 // CopyInputToS3 copies the input video to our S3 transfer bucket and probes the file.
-func (s *InputCopy) CopyInputToS3(requestID string, inputFile *url.URL) (inputVideoProbe video.InputVideo, signedURL string, osTransferURL *url.URL, err error) {
-	if isDirectUpload(inputFile) {
+func (s *InputCopy) CopyInputToS3(requestID string, inputFile *url.URL, encryptedKey string, VodDecryptPrivateKey *rsa.PrivateKey) (inputVideoProbe video.InputVideo, signedURL string, osTransferURL *url.URL, err error) {
```
victorges, line 45: What if we just got the `*DecryptionKeys` arg here? Will probably be cleaner overall since there's already a conditional on the `coordinator` today

clients/input_copy.go
```diff
@@ -197,12 +218,25 @@ func CopyFile(ctx context.Context, sourceURL, destOSBaseURL, filename, requestID
 		byteAccWriter := ByteAccumulatorWriter{count: 0}
 		defer func() { writtenBytes = byteAccWriter.count }()
 
+		var c io.ReadCloser
 		c, err := getFile(ctx, requestID, sourceURL, dStorage)
+
 		if err != nil {
-			return fmt.Errorf("download error: %w", err)
+			glog.Errorf("error getting file: %w", err)
+			return err
```
victorges, line 226: I think the old behavior of wrapping the error with a more specific error is cleaner. It will be logged on the outer layer
```suggestion
			return fmt.Errorf("download error: %w", err)
```

clients/input_copy.go
```diff
@@ -197,12 +218,25 @@ func CopyFile(ctx context.Context, sourceURL, destOSBaseURL, filename, requestID
...
 		}
+
 		defer c.Close()
 
+		if decryptor != nil {
+			decryptedFile, err := crypto.DecryptAESCBC(c, &decryptor.decryptKey, decryptor.encryptedKey)
```
victorges, line 232: Going even further, maybe `DecryptionKeys` should be in the `crypto` package and be the args for this func.

clients/input_copy.go
```diff
@@ -197,12 +218,25 @@ func CopyFile(ctx context.Context, sourceURL, destOSBaseURL, filename, requestID
...
+			if err != nil {
+				glog.Errorf("error decrypting file: %w", err)
+				return err
```
victorges, line 235:
```suggestion
				return fmt.Errorf("decrypt error: %w", err)
```

main.go
```diff
@@ -142,9 +145,15 @@ func main() {
 		glog.Info("Postgres metrics connection string was not set, postgres metrics are disabled.")
 	}
 
+	vodDecryptPrivateKey, err := crypto.LoadPrivateKey(cli.VodDecryptPrivateKey)
```
victorges, line 148: We should probably make this logic optional so the startup doesn't crash if these new args are not provided. Just safer in general to avoid failed deploys etc. Could you surround all this logic with a check for whether the CLI args were provided?

main.go
```diff
@@ -142,9 +145,15 @@ func main() {
...
+
+	if err != nil {
+		glog.Fatalf("Error loading vod decrypt private key: %v", err)
+	}
+
```
victorges, line 153: WDYT of adding a logic like this to validate that the public and private keys are compatible?
```suggestion
	publicKey, err := x509.ParsePKCS1PublicKey([]byte(cli.VodDecryptPublicKey))
	if err != nil {
		glog.Fatalf("Error parsing vod decrypt public key: %v", err)
	}
	if !publicKey.Equal(vodDecryptPrivateKey.Public()) {
		glog.Fatalf("Public key does not match private key")
	}
```

Or we could have a helper in `crypto` package like `crypto.ValidatePublicKey(...) error` or something.

Or... Maybe we don't need the public key CLI arg, can just generate it from the private key provided.

clients/input_copy.go
```diff
@@ -32,9 +36,24 @@ type InputCopy struct {
 	SourceOutputUrl string
 }
 
+type DecryptionKeys struct {
+	decryptKey   rsa.PrivateKey
```
victorges, line 40: This could probably be a pointer so we're not copying the private key around

victorges (approved):
LGTM! Just some minor suggestions

victorges:
Would be good to have a review from @mjh1 as well if you have a chance

PR #662, opened 2023-05-10
Detect when mediaconvert encounters errors specific to the input file

mjh1:


mjh1:
Closing, we already have these error codes humanized in task-runner

PR #667, opened 2023-05-12
Fail job if sending the final callback fails

leszko:
Send final callbacks (status: `TranscodeStatusError` and `TranscodeStatusCompleted`) in a sync manner and wait until the result. If sending the callback fails, then set the status of the job to `failed`.

It prevents the situation in which:
1. Catalyst successfully completed its job
2. It failed on sending the callback to task-runner
3. In studio we see the job as `failed`
4. In Catalyst metrics DB we see the job as `completed`

Note that from now on the opposite situation will be possible, so recording the job as `failed` in metrics DB and having task `complated` in task-runner.

### DIFF
--- a/clients/callback_client.go
+++ b/clients/callback_client.go
@@ -171,7 +171,10 @@ func (pcc *PeriodicCallbackClient) SendCallbacks() {
 		// Send non-terminal callbacks here in an async manner
 		// Terminal callbacks are sent when the job is finished in the sync manner
 		if !tsm.IsTerminal() {
-			go pcc.sendCallback(tsm)
+			go func(tsm TranscodeStatusMessage) {
+				// Ignore errors during async callback sending
+				_ = pcc.sendCallback(tsm)
+			}(tsm)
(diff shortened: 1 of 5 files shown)


clients/callback_client.go
```diff
@@ -89,24 +89,42 @@ func recoverer(f func()) {
...
-		}
+func (pcc *PeriodicCallbackClient) SendTranscodeStatus(tsm TranscodeStatusMessage) error {
+	if tsm.URL == "" {
+		return nil
+	}
+	pcc.updateTranscodeStatus(tsm)
+
+	// Terminal callbacks are sent here in a sync manner
+	// Non-terminal callbacks are sent periodically, in an async manner
+	if tsm.IsTerminal() {
```
thomshutt, line 100: Nice, so this also has the side-effect that we get the final callback quicker (i.e we don't have to wait for the callback interval to come around again), right?

thomshutt (approved):

PR #670, opened 2023-05-17
balancer: rewrite local node to 127.0.0.1 in MistUtilLoad

iameli:
Originally we were going to not pass the local node to the balancer entirely — we don't want that, because it means Catalyst nodes would never under any circumstances send people to watch a stream on themselves. Single-node clusters wouldn't work at all. Instead, the trick will be to rewrite our own node URL to 127.0.0.1 on the way in and out of the Mist load balancer. MistUtilLoad is already smart enough to ignore source requests coming from the same IP address: https://github.com/DDVTECH/mistserver/blob/6393a5a4cdc86667351f12f93b1e3cd22477a144/src/utils/util_load.cpp#L635-L638 so that will result in source (replication) requests never routing to the local node while playback requests can still do so.

### DIFF
--- a/balancer/balancer.go
+++ b/balancer/balancer.go
@@ -21,6 +21,7 @@ import (
 
 var mistUtilLoadSingleRequestTimeout = 15 * time.Second
 var mistUtilLoadLoopTimeout = 2 * time.Minute
+var mistLocalAddress = "http://127.0.0.1:4242"
 
 type Balancer interface {
 	Start(ctx context.Context) error
@@ -33,6 +34,7 @@ type Config struct {
 	Args                     []string
 	MistUtilLoadPort         uint32
 	MistLoadBalancerTemplate string
+	NodeName                 string
 }
 
 type BalancerImpl struct {
@@ -133,7 +135,13 @@ func (b *BalancerImpl) UpdateMembers(ctx context.Context, members []cluster.Memb
 func (b *BalancerImpl) changeLoadBalancerServers(ctx context.Context, server, action string) ([]byte, error) {
 	ctx, cancel := context.WithTimeout(ctx, mistUtilLoadSingleRequestTimeout)
 	defer cancel()
-	serverTmpl := fmt.Sprintf(b.config.MistLoadBalancerTemplate, server)
+	var serverTmpl string
+	if server == b.config.NodeName {
+		// Special case — make sure the balancer is aware this one is localhost
+		serverTmpl = mistLocalAddress
+	} else {
+		serverTmpl = b.formatNodeAddress(server)
+	}
 	actionURL := b.endpoint + "?" + action + "server=" + url.QueryEscape(serverTmpl)
 	req, err := http.NewRequest("POST", actionURL, nil)
 	req = req.WithContext(ctx)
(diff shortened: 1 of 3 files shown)


balancer/balancer.go
```diff
@@ -21,6 +21,7 @@ import (
 
 var mistUtilLoadSingleRequestTimeout = 15 * time.Second
 var mistUtilLoadLoopTimeout = 2 * time.Minute
+var mistLocalAddress = "http://127.0.0.1:4242"
```
iameli, line 24: I guess this could not be true; technically MistController could be running on a different port and we could make it configurable.

iameli: Oh wait, we totally already do. Duh. I'll update to use that.

pwilczynskiclearcode: https://github.com/livepeer/catalyst-api/blob/main/main.go#LL57C1-L58C1

balancer/balancer.go
```diff
@@ -196,14 +204,32 @@ func (b *BalancerImpl) getMistLoadBalancerServers(ctx context.Context) (map[stri
...
-	var mistResponse map[string]interface{}
+	var mistResponse map[string]any
 
 	err = json.Unmarshal([]byte(string(bytes)), &mistResponse)
 	if err != nil {
 		return nil, err
 	}
 
-	return mistResponse, nil
+	output := make(map[string]bool, len(mistResponse))
```
pwilczynskiclearcode, line 214: not sure why do we need a map? Will it ever contain `false` as a value?

balancer/balancer.go
iameli: TY!

balancer/balancer.go
```diff
@@ -133,7 +135,13 @@ func (b *BalancerImpl) UpdateMembers(ctx context.Context, members []cluster.Memb
 func (b *BalancerImpl) changeLoadBalancerServers(ctx context.Context, server, action string) ([]byte, error) {
 	ctx, cancel := context.WithTimeout(ctx, mistUtilLoadSingleRequestTimeout)
 	defer cancel()
-	serverTmpl := fmt.Sprintf(b.config.MistLoadBalancerTemplate, server)
+	var serverTmpl string
```
pwilczynskiclearcode, line 138: I'd call `serverURL` or sth... it's not a template any more, right?

balancer/balancer.go
iameli: Nope! But Go doesn't have a set, so this is the idiomatic way to do it. Actually, doing a quick Google, it looks like `map[string]struct{}` [takes zero bytes per entry instead of one byte per entry](https://stackoverflow.com/questions/34018908/golang-why-dont-we-have-a-set-datastructure#comment89711306_34020023), so I'll do that instead.

balancer/balancer.go
```diff
@@ -196,14 +204,32 @@ func (b *BalancerImpl) getMistLoadBalancerServers(ctx context.Context) (map[stri
...
 
 	err = json.Unmarshal([]byte(string(bytes)), &mistResponse)
 	if err != nil {
 		return nil, err
 	}
 
-	return mistResponse, nil
+	output := make(map[string]bool, len(mistResponse))
+
+	for k, _ := range mistResponse {
```
pwilczynskiclearcode, line 216: simply:
```go
	for k := range mistResponse {
```

pwilczynskiclearcode (approved):

balancer/balancer.go
pwilczynskiclearcode: I wasn't sure if we need a set when trusting mist not returning duplicates

PR #679, opened 2023-05-19
Return 200s for push rewrite trigger if stream is deleted or not found

mjh1:
https://linear.app/livepeer/issue/VID-49/catalyst-api-return-200s-for-push-rewrite-triggers


thomshutt (approved):
I think this looks good, probably best to wait for Eli's review before merging though

PR #684, opened 2023-05-23
re-enable mist cleanup script

emranemran:
We had issues in prod when it was first introduced a few weeks ago. The script has now been fixed so we're re-introducing it. Will let it soak in staging prior to launching in prod.

### DIFF
--- a/config/cli.go
+++ b/config/cli.go
@@ -30,6 +30,7 @@ type Cli struct {
 	MistBaseStreamName        string
 	MistLoadBalancerPort      int
 	MistLoadBalancerTemplate  string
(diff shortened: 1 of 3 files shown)


emranemran:
Asked Balder/Jaron to help me re-test locally on their Leiden setup to confirm no adverse effects with this revised version. But test plan is to merge in staging:
- setup long running live streams 
- monitor dashboard
- let it soak with 24-48hrs to ensure the script runs multiple times every 2hrs

thomshutt (approved):

mjh1 (approved):

PR #687, opened 2023-05-26
Ignore non-existing PPS probe error

mjh1:
We've been seeing this probing error from some recordings and have found that the videos still process successfully so we should ignore it.


video/probe.go
```diff
@@ -20,7 +20,22 @@ type Prober interface {
...
-func (p Probe) ProbeFile(url string, ffProbeOptions ...string) (iv InputVideo, err error) {
+func (p Probe) ProbeFile(url string, ffProbeOptions ...string) (InputVideo, error) {
+	iv, err := p.runProbe(url, ffProbeOptions...)
+	if err == nil {
+		return iv, nil
+	}
+
+	// ignore these probing errors if found and re-run with fatal loglevel to obtain the probe data
+	errMsg := strings.ToLower(err.Error())
+	if strings.Contains(errMsg, "parametric stereo signaled to be not-present but was found in the bitstream") ||
```
leszko, line 31: How about creating a const with all error messages to ignore? Like this:

```golang
var IGNORE_ERR_MESSAGES = []string{
	"parametric stereo signaled to be not-present but was found in the bitstream",
	"non-existing pps 0 referenced",
}

...

	for _, ignErr := range IGNORE_ERR_MESSAGES {
		if strings.Contains(errMsg, ignErr) {
                        return p.runProbe(url, "-loglevel", "fatal")
		}
	}
```

leszko (approved):
Added one comment. Other than that, does this PR fix this issue: https://linear.app/livepeer/issue/VID-197/be-less-strict-about-vod-probing?

video/probe.go
thomshutt: Looks good, I'll pick this up since Max is out until next week

thomshutt: Done

thomshutt (approved):

leszko:
> Added one comment. Other than that, does this PR fix this issue: https://linear.app/livepeer/issue/VID-197/be-less-strict-about-vod-probing?

@thomshutt Do you maybe know if it's related?

thomshutt:
> > Added one comment. Other than that, does this PR fix this issue: https://linear.app/livepeer/issue/VID-197/be-less-strict-about-vod-probing?
> 
> @thomshutt Do you maybe know if it's related?

I think just partially. This one is focussed on a particular error we get from Recording sources, but should also remove some false positive rejections from VOD sources too.

PR #699, opened 2023-06-05
encryption: adding spki key to public_key handler to support native js public key import

gioelecerati:


handlers/accesscontrol/encryption.go
```diff
@@ -25,9 +26,17 @@ func (ec *EncryptionHandlersCollection) PublicKeyHandler() httprouter.Handle {
 		w.Header().Set("Content-Type", "application/json")
 		w.WriteHeader(http.StatusOK)
 
+		spkiPublicKey, err := crypto.ConvertToSpki(ec.publicKey)
```
victorges, line 29: Can you call this on the constructor instead? The inputs are constant so we don't need to be recomputing this conversion on every request.

crypto/decryption.go
```diff
@@ -170,3 +171,46 @@ func decryptReaderTo(readerRaw io.Reader, writer io.Writer, decrypter cipher.Blo
...
+
+	pubPEM := pem.EncodeToMemory(&pem.Block{
+		Type:  "PUBLIC KEY",
+		Bytes: pubBytes,
+	})
+
+	pubPEMStr := string(pubPEM)
+
+	// Remove header and footer
+	spkiPubKeyBase64 := strings.Join(strings.Split(pubPEMStr, "\n")[1:len(strings.Split(pubPEMStr, "\n"))-2], "")
```
victorges, line 210: A couple things:
 - You are splitting by `\n` and then joining with an empty string. Is this correct? If so, can you update the comment to explain that as well?

Cosmetic:
 - I was confused by this variable name given what the comment says, maybe just update `pubPEMStr` here?
 - Could you break up this transformation in a couple of lines? It's a bit hard to understand what's happening here and there are duplicate function calls. Sth like a helper `lines := strings.Join(strings.Split(pubPEMStr, "\n")` would already help a lot!

victorges (approved):
Wow I didn't imagine we would have to implement so much formatting ourselves 😩

crypto/decryption.go
gioelecerati: I split it into two lines and made an explainatory comment! Let me know if it's better!

handlers/accesscontrol/encryption.go
gioelecerati: Moved into the handler constructor

PR #704, opened 2023-06-08
transcode: drop last segment in HLS manifest a video track is missing

emranemran:
The last segment in an HLS manifest may contain an audio-only track - this is common during a livestream recording where the video stream can end sooner with a trailing audio stream which results in a segment at the end that just contains audio. This segment should *not* be submitted to the T.

TODO: 
* need to add a unit test for this
* verify recordings without this last-segment-is-audio-only-track issue still works as expected.

### DIFF
--- a/transcode/transcode.go
+++ b/transcode/transcode.go
@@ -94,6 +94,34 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
 		return outputs, segmentsCount, fmt.Errorf("error generating source segment URLs: %s", err)
 	}
 
+	// The last segment in an HLS manifest may contain an audio-only track - this is common
+	// during a livestream recording where the video stream can end sooner with a trailing audio stream
+	// which results in a segment at the end that just contains audio. This segment should *not* be
+	// submitted to the T.
+	lastSegment := sourceSegmentURLs[len(sourceSegmentURLs)-1]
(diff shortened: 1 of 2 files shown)


transcode/transcode.go
```diff
@@ -94,6 +94,34 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
...
+	// GetTrack will return an err if TrackTypeVideo was not found
+	_, err = lastSegmentProbe.GetTrack(video.TrackTypeVideo)
+	if err != nil {
+		var lastSegmentIdx int
+		for i, entry := range sourceManifest.Segments {
+			if entry == nil {
+				lastSegmentIdx = i - 1
+				break
+			}
+		}
```
emranemran, line 118: Thanks for leaving this comment @thomshutt here - a simple change to drop an element from this `Segments` slice was giving me a headache and had me questioning my existence and why this slice length was pre-allocated 1024 bytes: 
https://github.com/livepeer/catalyst-api/blob/main/clients/manifest.go#L69

mjh1: There is a helper func btw: `GetAllSegments()` so you don't have to remember.

mjh1 (commented):
lgtm pending the unit test

mjh1 (approved):

PR #709, opened 2023-06-12
Add field to notfiy task runner about source playback

mjh1:
I'm adding this field on its own initially so that I can update task-runner with it too before using it, thus avoiding any json parse issues


mjh1:
@victorges This is the correct thing to do right, to avoid the chance of any prod breakage if catalyst was rolled out before task-runner?

victorges (approved):

PR #714, opened 2023-06-14
Source playback

mjh1:
To improve our time to ready we are implementing source playback. We write a master playlist pointing at the segmented source playlist and notify task-runner this is available (which in turn notifies studio). When transcoding is complete this master playlist is overwritten with the final transcoded version.

### DIFF
--- a/clients/callback_client.go
+++ b/clients/callback_client.go
@@ -97,7 +97,7 @@ func (pcc *PeriodicCallbackClient) SendTranscodeStatus(tsm TranscodeStatusMessag
 
 	// Terminal callbacks are sent here in a sync manner
 	// Non-terminal callbacks are sent periodically, in an async manner
-	if tsm.IsTerminal() {
+	if tsm.IsTerminal() || tsm.SourcePlayback != nil {
 		return pcc.sendCallback(tsm)
 	}
 	return nil
--- a/pipeline/ffmpeg.go
+++ b/pipeline/ffmpeg.go
@@ -174,6 +174,7 @@ func sendSourcePlayback(job *JobInfo) {
 		return
 	}
 
+	// source playback won't currently work for token gating so we're excluding the private buckets here
 	for _, blocked := range sourcePlaybackBucketBlocklist {
 		if segmentingBucket == blocked {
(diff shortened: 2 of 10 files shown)


clients/callback_client.go
```diff
@@ -97,7 +97,7 @@ func (pcc *PeriodicCallbackClient) SendTranscodeStatus(tsm TranscodeStatusMessag
 
 	// Terminal callbacks are sent here in a sync manner
 	// Non-terminal callbacks are sent periodically, in an async manner
-	if tsm.IsTerminal() {
+	if tsm.IsTerminal() || tsm.SourcePlayback != nil {
```
thomshutt, line 100: Is this just in here for a speed boost?

thomshutt: Would this not mean that we start spamming callbacks for the entire transcode time (i.e because SourcePlayback continues being not nil)

pipeline/ffmpeg.go
```diff
@@ -150,6 +153,63 @@ func (f *ffmpeg) HandleStartUploadJob(job *JobInfo) (*HandlerOutput, error) {
 		}}, nil
 }
 
+var sourcePlaybackBucketBlocklist = []string{"lp-us-catalyst-vod-pvt-monster", "lp-us-catalyst-vod-pvt-com"}
```
thomshutt, line 156: Could you add a comment explaining this please

clients/callback_client.go
mjh1: It seems to not work at all without this, I believe because of the transcoding status being sent without the source playback field (also I am still sending a `TranscodeStatusPreparingCompleted` update for the case where source playback isn't done. So this seems like the safest/best option.. there would also be no harm in some how persisting the source playback field for all subsequent statuses but not sure if it's necessary, a bit of extra complexity when I can only see an issue when we the status fails to send here, which hopefully is quite rare..?

mjh1: Maybe best to keep it simple like this for now and make an improvement if we see a lot of failures to send the status.

thomshutt: 👍  sounds good

thomshutt (approved):

PR #717, opened 2023-06-15
mapic: use Transfer-Encoding: chunked when rejecting streams

iameli:
The right way to reject a stream from PUSH_REWRITE:
1. Respond with an empty string
2. Use "Transfer-Encoding: Chunked" to work around a Mist bug

This adds a helper reject() function in the handler that makes it very easy to do so.

### DIFF
--- a/mapic/mistapiconnector_app.go
+++ b/mapic/mistapiconnector_app.go
@@ -381,11 +381,16 @@ func (mc *mac) triggerDefaultStream(w http.ResponseWriter, r *http.Request, line
 }
 
 func (mc *mac) triggerPushRewrite(w http.ResponseWriter, r *http.Request, lines []string, rawRequest string) bool {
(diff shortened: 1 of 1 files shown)


thomshutt (approved):

thomshutt:
@iameli I don't see `Transfer-Encoding: Chunked` being added anywhere

thomshutt:
Huh yep, even though this suggests it gets set if the content-length is <= 0:
image

I tested without flushing:
```
< HTTP/1.1 200 OK
< Date: Fri, 16 Jun 2023 06:23:14 GMT
< Content-Length: 0
< 
* Connection #0 to host localhost left intact
```
and with:
```
< HTTP/1.1 200 OK
< Date: Fri, 16 Jun 2023 06:23:49 GMT
< Transfer-Encoding: chunked
< 
* Connection #0 to host localhost left intact
```

PR #718, opened 2023-06-16
Fix jwt access control playback

mjh1:
We were only taking into account accessKey in the manifest url substitution logic, now we're handling jwt too.


handlers/playback.go
```diff
@@ -23,12 +23,20 @@ func PlaybackHandler() httprouter.Handle {
...
 		playbackReq := playback.Request{
-			RequestID:  requestID,
-			PlaybackID: params.ByName("playbackID"),
-			File:       params.ByName("file"),
-			AccessKey:  req.URL.Query().Get(playback.KeyParam),
-			Range:      req.Header.Get("range"),
+			RequestID:       requestID,
+			PlaybackID:      params.ByName("playbackID"),
+			File:            params.ByName("file"),
+			GatingParam:     gatingParam,
```
victorges, line 37: Interesting! So the current logic already detects if it's a jwt and handles as that? I thought it got the "type" somewhere, but it makes sense that it doesn't

victorges (approved):
LGTM!

PR #721, opened 2023-06-20
Add timestamps for processing stages to analytics DB

mjh1:


thomshutt (approved):
LGTM, do we need @ecmulli to update the schemas before we merge?

PR #725, opened 2023-06-21
Fix recordings

mjh1:
We want to pass the plain HTTP input URL through for recordings because this is the simplest way to allow mediaconvert to read the input (in case of livepeer pipeline failures). Some parts of the code were expecting an OS URL though so this is what is fixed in this PR, to allow plain HTTP URLs to work too.


clients/input_copy.go
```diff
@@ -36,9 +36,8 @@ type InputCopy struct {
 
 // CopyInputToS3 copies the input video to our S3 transfer bucket and probes the file.
 func (s *InputCopy) CopyInputToS3(requestID string, inputFile *url.URL, decryptor *crypto.DecryptionKeys) (inputVideoProbe video.InputVideo, signedURL string, osTransferURL *url.URL, err error) {
-
 	if isDirectUpload(inputFile) && decryptor == nil {
-		log.Log(requestID, "Direct upload detected", "source", inputFile.String())
+		log.Log(requestID, "Direct upload detected")
```
mjh1, line 40: source is already part of the log context here

thomshutt (approved):

PR #733, opened 2023-06-26
Fix recordings source playback

mjh1:
We need to go back to copying the recording manifest and segments to the playback bucket to allow source playback to work, the playback urls are pointed at this bucket not the recordings one.

To get around the issue we had with mediaconvert reading a signed hls input url (appending the signing queryparams to the segment urls) I am now only generating a signed url when the unsigned version is not accessible.

### DIFF
--- a/clients/input_copy.go
+++ b/clients/input_copy.go
@@ -95,6 +95,9 @@ func (s *InputCopy) CopyInputToS3(requestID string, inputFile *url.URL, decrypto
 
 func getSignedURL(osTransferURL *url.URL) (string, error) {
 	// check if plain https is accessible, if not then the bucket must be private and we need to generate a signed url
+	// in most cases signed urls work fine as input but in the edge case where we have to fall back to mediaconvert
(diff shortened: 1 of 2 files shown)


clients/input_copy.go
```diff
@@ -93,6 +93,21 @@ func (s *InputCopy) CopyInputToS3(requestID string, inputFile *url.URL, decrypto
 	return
 }
 
+func getSignedURL(osTransferURL *url.URL) (string, error) {
+	// check if plain https is accessible, if not then the bucket must be private and we need to generate a signed url
```
mjh1, line 97: @victorges @thomshutt is this too hacky, what do you think?

thomshutt: I quite like this, maybe just add to the comment that signed URLs don't work with Mediaconvert for whoever comes along in a year and thinks "Why don't we just sign everything?"

thomshutt (approved):

PR #734, opened 2023-06-26
Fix ingests with unusual display aspect ratios

mjh1:
Display aspect ratios were seemingly being ignored in the livepeer pipeline so for now we're fixing that by sending jobs to mediaconvert when the display aspect ratio doesn't match the resolution of the video


pipeline/coordinator.go
```diff
@@ -329,6 +329,18 @@ func checkLivepeerCompatible(requestID string, strategy Strategy, iv video.Input
...
+		if track.Type == video.TrackTypeVideo && track.DisplayAspectRatio != "" {
+			aspectRatioParts := strings.Split(track.DisplayAspectRatio, ":")
+			if len(aspectRatioParts) == 2 {
+				w, err := strconv.ParseFloat(aspectRatioParts[0], 64)
+				h, err2 := strconv.ParseFloat(aspectRatioParts[1], 64)
+				if err == nil && err2 == nil &&
+					w/h != float64(track.Width)/float64(track.Height) {
+					log.Log(requestID, "display aspect ratio doesn't match resolution, not supported by Livepeer pipeline", "display_aspect_ratio", track.DisplayAspectRatio, "width", track.Width, "height", track.Height)
+					return livepeerNotSupported(strategy)
+				}
```
mjh1, line 341: I think i'll extract this to a func and check for division by zero

leszko (approved):

pipeline/coordinator.go
thomshutt: I'm a little nervous about us getting false positives here because of the floating point maths, should we add some "fuzziness" factor?

mjh1: Yeah.. you might be right. I did run this over the most recent 1000 videos yesterday and it only found 5 or so, but still feels a little unsafe. I'll check those videos it found and see if there was any valid issue and what the difference is in the calculation.

mjh1: Ok yeah most of them are tiny differences, looks like rounding difference in the aspect ratio and resolution values. e.g. one video had a 476:831 aspect ratio and  476x830 for the resolution. Will add some fuzziness.

mjh1: Actually there were only two that i found and largest difference between the two ratios was 7%, I'm thinking allow up to 20% before sending to mediaconvert?

mjh1:
@thomshutt ready for another review 🙏

pipeline/coordinator.go
thomshutt: Yeah sound good, we can adjust as needed

thomshutt (approved):

PR #737, opened 2023-06-28
[VID-258][VID-4] remove CONN_CLOSE

iameli:
Adds a `Broker` interface responsible for dispatching Mist events to other parts of the code. If some part (like mapic) needs to handle an event, it can be passed the Broker on initialization and use it something like [this](https://github.com/livepeer/catalyst-api/pull/737/files#diff-07d76f8f8ce6678348918ef729dbe84cf0bea01df42fc4c84e0c2aabf8dd8ce4R142).

Also a bonus commit to once again fix the `reject()` logic for webhooks! I swear I had tested and had it working but apparently not.

### DIFF
--- a/api/http_internal.go
+++ b/api/http_internal.go
@@ -26,8 +26,8 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
-func ListenAndServeInternal(ctx context.Context, cli config.Cli, vodEngine *pipeline.Coordinator, mapic mistapiconnector.IMac, bal balancer.Balancer, c cluster.Cluster) error {
-	router := NewCatalystAPIRouterInternal(cli, vodEngine, mapic, bal, c)
+func ListenAndServeInternal(ctx context.Context, cli config.Cli, vodEngine *pipeline.Coordinator, mapic mistapiconnector.IMac, bal balancer.Balancer, c cluster.Cluster, broker misttriggers.Broker) error {
+	router := NewCatalystAPIRouterInternal(ctx, cli, vodEngine, mapic, bal, c, broker)
 	server := http.Server{Addr: cli.HTTPInternalAddress, Handler: router}
 	ctx, cancel := context.WithCancel(ctx)
 
@@ -53,7 +53,7 @@ func ListenAndServeInternal(ctx context.Context, cli config.Cli, vodEngine *pipe
 	return server.Shutdown(ctx)
 }
 
-func NewCatalystAPIRouterInternal(cli config.Cli, vodEngine *pipeline.Coordinator, mapic mistapiconnector.IMac, bal balancer.Balancer, c cluster.Cluster) *httprouter.Router {
+func NewCatalystAPIRouterInternal(ctx context.Context, cli config.Cli, vodEngine *pipeline.Coordinator, mapic mistapiconnector.IMac, bal balancer.Balancer, c cluster.Cluster, broker misttriggers.Broker) *httprouter.Router {
 	router := httprouter.New()
 	withLogging := middleware.LogRequest()
 	withAuth := middleware.IsAuthorized
@@ -72,7 +72,7 @@ func NewCatalystAPIRouterInternal(cli config.Cli, vodEngine *pipeline.Coordinato
 	accessControlHandlers := accesscontrol.NewAccessControlHandlersCollection(cli)
 	encryptionHandlers := accesscontrol.NewEncryptionHandlersCollection(cli, spkiPublicKey)
 	adminHandlers := &admin.AdminHandlersCollection{Cluster: c}
-	mistCallbackHandlers := misttriggers.NewMistCallbackHandlersCollection(cli)
+	mistCallbackHandlers := misttriggers.NewMistCallbackHandlersCollection(cli, broker)
 
 	// Simple endpoint for healthchecks
 	router.GET("/ok", withLogging(catalystApiHandlers.Ok()))
@@ -110,7 +110,7 @@ func NewCatalystAPIRouterInternal(cli config.Cli, vodEngine *pipeline.Coordinato
 	router.GET("/api/pubkey", withLogging(encryptionHandlers.PublicKeyHandler()))
 
 	// Endpoint to receive "Triggers" (callbacks) from Mist
-	router.POST("/api/mist/trigger", withLogging(mistCallbackHandlers.Trigger()))
+	router.POST("/api/mist/trigger", withLogging(mistCallbackHandlers.Trigger(ctx)))
 
 	// Endpoint to receive segments and manifests that ffmpeg produces
 	router.PUT("/api/ffmpeg/:id/:filename", withLogging(ffmpegSegmentingHandlers.NewFile()))
--- a/api/http_internal.go
+++ b/api/http_internal.go
@@ -26,7 +26,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
-func ListenAndServeInternal(ctx context.Context, cli config.Cli, vodEngine *pipeline.Coordinator, mapic mistapiconnector.IMac, bal balancer.Balancer, c cluster.Cluster, broker misttriggers.Broker) error {
+func ListenAndServeInternal(ctx context.Context, cli config.Cli, vodEngine *pipeline.Coordinator, mapic mistapiconnector.IMac, bal balancer.Balancer, c cluster.Cluster, broker misttriggers.TriggerBroker) error {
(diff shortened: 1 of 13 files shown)


middleware/logging.go
```diff
@@ -34,6 +34,11 @@ func (rw *responseWriter) WriteHeader(code int) {
 	rw.wroteHeader = true
 }
 
+func (rw *responseWriter) Flush() {
```
victorges, line 37: Is this used implicitly somewhere?

victorges (approved):
LGTM

middleware/logging.go
iameli: Yeah https://github.com/livepeer/catalyst-api/blob/6b158fe32d216fb5278756f770bceaac1d3f85d0/mapic/mistapiconnector_app.go#L368-L371

handlers/misttriggers/broker.go
```diff
@@ -0,0 +1,80 @@
...
+// Broker provides an interface for allowing multiple components of the app
+// to respond to Mist trigger events without putting all of the application
+// logic into the trigger handling code itself.
+
+// There are three different cases to account for on Mist triggers; and
+// accordingly the TriggerXXX functions have three different signatures:
+// 1. Purely informative, like STREAM_BUFFER. This can't be rejected back
+//    to Mist, so no error signature necessary. All of the callbacks can
+//    be fired in parallel.
+// 2. Allow/deny, like USER_NEW. We either let the new viewer come in
```
leszko, line 20: I see that you didn't use `USER_NEW` (or `PUSH_REWRITE`) in the `Broker` right now. Is the intention to put them here later?

If yes, then I suggest creating some `// TODO: VID-xxx` to put them here later. Otherwise, it's a little confusing why you have it here in the comment, but not in the code.

handlers/misttriggers/broker.go
```diff
@@ -0,0 +1,80 @@
...
+	g.mutex.Lock()
+	defer g.mutex.Unlock()
+	g.funcs = append(g.funcs, cb)
+}
+
+func (g *funcGroup[T]) Trigger(ctx context.Context, payload *T) error {
+	g.mutex.RLock()
+	defer g.mutex.RUnlock()
+	group, ctx := errgroup.WithContext(ctx)
+	// ...yuck. Is there a better way?
```
leszko, line 71:
```suggestion
```

I think it's ok what you do here. So, you can remove the comment 🙃

handlers/misttriggers/broker.go
```diff
@@ -0,0 +1,80 @@
...
+	OnStreamBuffer(func(context.Context, *StreamBufferPayload) error)
+
+	TriggerStreamBuffer(context.Context, *StreamBufferPayload)
+}
+
+func NewBroker() Broker {
+	return &broker{}
+}
+
+type broker struct {
```
leszko, line 39: Please add a unit test.

handlers/misttriggers/stream_buffer.go
```diff
@@ -50,6 +51,7 @@ func (d *MistCallbackHandlersCollection) TriggerStreamBuffer(w http.ResponseWrit
 	}
 
 	rawBody, _ := json.Marshal(body)
+	go d.broker.TriggerStreamBuffer(ctx, body)
```
leszko, line 54: If you run this in an async manner, I wonder what is the value of waiting in the broker itself?

I think that if you need an async workflow here, then probably there is no use in `group.Wait()` in the broker's trigger implementation.

handlers/misttriggers/stream_buffer.go
```diff
@@ -126,6 +128,18 @@ type StreamBufferPayload struct {
...
+}
+
+func (s *StreamBufferPayload) IsFull() bool {
+	return s.State == "FULL"
+}
+
+func (s *StreamBufferPayload) IsRecover() bool {
+	return s.State == "RECOVER"
+}
+
```
leszko, line 142: Nice! I like that we have now separate functions for each check 🏅

handlers/misttriggers/triggers.go
```diff
@@ -18,18 +19,23 @@ const (
...
 
 type MistCallbackHandlersCollection struct {
-	cli *config.Cli
+	cli    *config.Cli
+	broker Broker
 }
 
-func NewMistCallbackHandlersCollection(cli config.Cli) *MistCallbackHandlersCollection {
-	return &MistCallbackHandlersCollection{cli: &cli}
+type TriggerPayload interface {
```
leszko, line 26: Nit: don't you want to move it into `broker.go`?

handlers/misttriggers/broker.go
```diff
@@ -0,0 +1,80 @@
...
+// 2. Allow/deny, like USER_NEW. We either let the new viewer come in
+//    or we kick them out, but there's no return value other than that.
+//    These handlers can be called in parallel too, but any one of them
+//    returning an error will cause an (immediate) trigger rejection.
+// 3. Triggers with response values, like PUSH_REWRITE. These functions need
+//    to return both an error (for rejections) and a string (for responses).
+//    They can't be called in parallel; there really should only be one
+//    handler for these sorts of triggers.
+
+type Broker interface {
```
leszko, line 29: nit: the name `Broker` seems very generic, don't we want to call it something like `TriggerBroker` or `MistTriggerBroker`. I know it's longer, but I imagine myself in a few months thinking "what's this broker? 😄 "

leszko (approved):
Added some comments. Other than that LGTM.

About your question:
>  I'd love some eyes on the interface for this, the pattern kinda feels more like JavaScript than Go

I like this pattern here. In general, Go standards are "be pragmatic". I think this broker pattern is pragmatic here and simplifies a lot. Though, you know, I spend half of my life programming in Java... 🤣

api/http_internal.go
```diff
@@ -53,7 +53,7 @@ func ListenAndServeInternal(ctx context.Context, cli config.Cli, vodEngine *pipe
 	return server.Shutdown(ctx)
 }
 
-func NewCatalystAPIRouterInternal(cli config.Cli, vodEngine *pipeline.Coordinator, mapic mistapiconnector.IMac, bal balancer.Balancer, c cluster.Cluster) *httprouter.Router {
+func NewCatalystAPIRouterInternal(ctx context.Context, cli config.Cli, vodEngine *pipeline.Coordinator, mapic mistapiconnector.IMac, bal balancer.Balancer, c cluster.Cluster, broker misttriggers.Broker) *httprouter.Router {
```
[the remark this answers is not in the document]
thomshutt: This is the only bit of the PR I don't love - the approach I'd probably take is just to YOLO the server shutdown case, since it's not like any of the handlers are going to be long-running processes. 

Another approach would be to maybe store the global context in some top-level variable and then derive a local context off that at the point we're about to loop through instantiating all the handler functions.

handlers/misttriggers/broker.go
```diff
@@ -0,0 +1,60 @@
...
+	defer g.mutex.Unlock()
+	g.funcs = append(g.funcs, cb)
+}
+
+func (g *funcGroup[T]) Trigger(ctx context.Context, payload *T) error {
+	g.mutex.RLock()
+	defer g.mutex.RUnlock()
+	group, ctx := errgroup.WithContext(ctx)
+	// ...yuck. Is there a better way?
+	for _, cb := range g.funcs {
```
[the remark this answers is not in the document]
thomshutt: The normal pattern I've seen here is to do a `cb := cb` inside the other loop, which is clunky but I think a bit more readable than wrapping the function

handlers/misttriggers/broker.go
thomshutt: Yeah, I'd also vote for not making this PR bigger but for prioritising a follow-up to convert the others while this is still fresh in everyone's minds

handlers/misttriggers/broker.go
thomshutt: Definitely agree with this - I read "Broker Pattern" and thought it was going to be some overly-abstracted monstrosity but it actually makes a lot of sense, so I think a more approachable name is all we need

handlers/misttriggers/broker.go
```diff
@@ -0,0 +1,80 @@
...
+
+	"github.com/golang/glog"
+	"golang.org/x/sync/errgroup"
+)
+
+// Broker provides an interface for allowing multiple components of the app
+// to respond to Mist trigger events without putting all of the application
+// logic into the trigger handling code itself.
+
+// There are three different cases to account for on Mist triggers; and
```
thomshutt, line 15: I love this breakdown of the trigger types, really helped to clarify them in my mind and having a separate interface entry per trigger type is a nice way of handling the variances

PR #738, opened 2023-06-29
Clean up unused code

leszko:
I believe we don't use this record event callbacks, but I might be wrong, so please double check @thomshutt @iameli


thomshutt (approved):
That's right, good catch!

PR #740, opened 2023-06-29
Single download pipeline - implement request param to make a source file copy

mjh1:
- Moved a few lines of code that decide the copy destination out of `clients/input_copy.go` and into `pipeline/coordinator.go` to be able to reference the `SourceCopy` flag on the job object. If the flag is set we will now copy the source file to the `video` location which task-runner and studio expect.
- Set some missing probe fields so that we can then remove the probing from task-runner


thomshutt (approved):
Looks good, but should we also add a check that the source file was copied in the Cucumber tests?

leszko (approved):
LGTM 👍

PR #744, opened 2023-06-30
consolidate stream triggers in handlers/misttriggers & broker

iameli-streams:
Big refactor PR! Here are the highlights:

1. ALL handling of Mist triggers has been consolidated in `handlers/misttriggers`. All of that parsing and handling code has 100% test coverage. Should you need to listen to a new Mist stream trigger, you can take a look at the code in `handlers/misttriggers/triggers.go`; there are `OnTriggerName` functions that you can use to register a new handler.
2. Previously we had two Mist clients, one in the `clients` directory and one in the `mapic` directory. These have been consolidated in the `clients` directory, with a few things ported over from the mapic client. This has made mapic much smaller, hooray!
3. Related to the above, Mist is now assumed to be local to the Catalyst node and therefore we can use unauthenticated HTTP requests to the local API. `-mist-user` and `-mist-pass` are no longer used and deprecated.
4. catalyst-api now talks to Mist on startup and makes sure all of its stream triggers are in place with the appropriate URL. One consequence of this is that catalyst-api now crashes if it can't talk to Mist on startup. If you're booting catalyst-api without the rest of the node for development or CI purposes, you can use the new `-no-mist` flag.


clients/mist_client.go
```diff
@@ -17,19 +19,26 @@ type MistAPIClient interface {
 	AddStream(streamName, sourceUrl string) error
 	PushStart(streamName, targetURL string) error
 	DeleteStream(streamName string) error
-	AddTrigger(streamName, triggerName string) error
-	DeleteTrigger(streamName, triggerName string) error
+	AddTrigger(streamName []string, triggerName string, sync bool) error
```
leszko, line 22: 1. Why do we need to change this `string` into `[]string`?
2. Do we still need all these functions: `AddStream`, `PushStart`, etc. I don't see their usage anymore. Maybe we can clean them up?

clients/mist_client.go
```diff
@@ -81,6 +90,48 @@ type MistStreamInfo struct {
...
+}
+
+func (p *MistPush) UnmarshalJSON(data []byte) error {
+	// this field is undocumented and shows up as null everytime.
+	var unknown json.RawMessage
+	var tmp MistPush
+	if err := UnmarshalJSONArray(data, &tmp.ID, &tmp.Stream, &tmp.OriginalURI, &tmp.EffectiveURI, &unknown, &tmp.Stats); err != nil {
+		return err
+	}
+	*p = tmp
```
leszko, line 108: Thing looks weird, why do you assign `tmp` to Receiver? I think the normal flow would be to just return `(MistPush, error)`.

clients/mist_client.go
```diff
@@ -81,6 +90,48 @@ type MistStreamInfo struct {
...
 
+type MistPush struct {
+	ID           int64
+	Stream       string
+	OriginalURI  string
+	EffectiveURI string
+	Stats        *MistPushStats
+}
+
+func (p *MistPush) UnmarshalJSON(data []byte) error {
```
leszko, line 101: Could you add unit test for this function?

clients/mist_client.go
```diff
@@ -81,6 +90,48 @@ type MistStreamInfo struct {
...
+	var unknown json.RawMessage
+	var tmp MistPush
+	if err := UnmarshalJSONArray(data, &tmp.ID, &tmp.Stream, &tmp.OriginalURI, &tmp.EffectiveURI, &unknown, &tmp.Stats); err != nil {
+		return err
+	}
+	*p = tmp
+	return nil
+}
+
+func UnmarshalJSONArray(data []byte, values ...interface{}) error {
```
leszko, line 112: Could you add unit test?

clients/mist_client.go
```diff
@@ -81,6 +90,48 @@ type MistStreamInfo struct {
...
+	if err := json.Unmarshal(data, &valuesData); err != nil {
+		return err
+	}
+	if len(valuesData) > len(values) {
+		valuesData = valuesData[:len(values)]
+	}
+	for i, vd := range valuesData {
+		if err := json.Unmarshal(vd, values[i]); err != nil {
+			return err
+		}
```
leszko, line 123:
```suggestion
	for i := 0; i < len(values); i++ {
		if err := json.Unmarshal(valuesData[i], values[i]); err != nil {
			return err
		}
	}
```

clients/mist_client.go
```diff
@@ -310,49 +361,50 @@ type Config struct {
...
-		}
-		if !f {
+		if !sameStringSlice(streamNames, t.Streams) {
 			res = append(res, t)
 		}
 	}
 	return res
 }
 
+func sameStringSlice(s1, s2 []string) bool {
```
leszko, line 400: The same, I think you could add a unit test for this function.

config/cli.go
```diff
@@ -79,6 +80,11 @@ func (cli *Cli) ShouldMapic() bool {
 	return cli.APIServer != ""
 }
 
+// Should we attempt communication with Mist?
+func (cli *Cli) ShouldMist() bool {
```
leszko, line 84: What happens with Catalyst if not? Does running catalyst without Mist makes any sense?

leszko (commented):
I like the consolidation of Mist triggers. Good work!

Added some comments. Other than that, my two main concerns are:
1. Do we still need support for all these triggers (like some code in `MistClient` seems unused)?
2. If yes, then do we maybe have a diagram with these interactions? I'm getting confused of how it all works

clients/mist_client.go
iameli-streams: 1. We want our triggers to apply to all streams on the node, which is represented in Mist as an empty array; thus the signature change.
2. Some of them aren't used right now, such as `PushStart`, but they _will_ be used as part of fixing multistreaming. So I'd prefer not to delete them now and bring them back later.

More broadly, there are two Mist clients right now -- one in the `mapic` directory and one in the `clients` directory. These need to get consolidated, I'll look into that today.

clients/mist_client.go
iameli-streams: Good question. I copied it from the mapic Mist client, but I'll revisit.

config/cli.go
iameli-streams: I think lots of folks hacking on catalyst-api boot it up without Mist, but the default case should definitely be to crash if there's no Mist there, that indicates a problem with the node. @thomshutt's suggestion was to just have a `-no-mist` parameter for that case, so I'll set that up.

clients/mist_client.go
iameli-streams: Oh, I realized why eventually -- this pattern is to implement [the `json.Unmarshaler` interface](https://pkg.go.dev/encoding/json#Unmarshaler), so this function gets called with its custom behavior when we do a normal `json.Unmarshal`. I guess the `tmp` pattern is so that we don't make any changes to the current struct while unmarshalling? But if there's an error our struct doesn't matter anymore anyway so maybe it can be made cleaner, I'll play with it.

clients/mist_client.go
leszko: Yeah, but I also wonder if we actually use all these functions anywhere? Like `PushStart()`?

iameli-streams:
We use PushStart now that I deleted the other Mist client in mapic and
migrated it to use this one

clients/mist_client.go
iameli: Oops, put the other comment in the wrong place. We use PushStart now that I consolidated the two Mist clients - mapic uses it to start multistreaming

handlers/misttriggers/push_out_start_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+func TestItCanRejectPushOutStartRequests(t *testing.T) {
+	rr := doPushOutStartRequest(t, pushOutStartPayload, func(ctx context.Context, p *PushOutStartPayload) (string, error) {
+		// Proper way to reject Mist
+		return "", nil
+	})
+	require.Equal(t, rr.Result().StatusCode, 200)
+	require.Equal(t, rr.Body.String(), "")
+	// doesn't work in the test environment, but does get set in production:
+	// require.Equal(t, rr.Result().Header.Get("Transfer-Encoding"), "chunked")
```
leszko, line 64:
```suggestion
```

leszko (approved):
I like the refactoring. TBH I didn't review each single line, because AFAIU most of the changes are just moving code from one part to another. 

@iameli If there is any part of the code that you'd like to get reviewed in detail, let me know.

PR #749, opened 2023-07-05
test: avoid port collisions

iameli-streams:
The integration tests were making use of ports 3000, 7979, 8989, and 8935. This obviously needs to port conflicts if you try and run `make integration-test` while running a node locally, which is a shame.

I thought it would be a small change, but it turns out making the local broadcaster URL configurable is a little bit larger - needed to update the flow of how that URL and clients are passed down. Not very familiar with those parts of the code, let me know if there's a better way to do it

### DIFF
--- a/main.go
+++ b/main.go
@@ -56,6 +56,7 @@ func main() {
 	config.URLSliceVarFlag(fs, &cli.ImportIPFSGatewayURLs, "import-ipfs-gateway-urls", "https://vod-import-gtw.mypinata.cloud/ipfs/?pinataGatewayToken={{secrets.LP_PINATA_GATEWAY_TOKEN}},https://w3s.link/ipfs/,https://ipfs.io/ipfs/,https://cloudflare-ipfs.com/ipfs/", "Comma delimited ordered list of IPFS gateways (includes /ipfs/ suffix) to import assets from")
 	config.URLSliceVarFlag(fs, &cli.ImportArweaveGatewayURLs, "import-arweave-gateway-urls", "https://arweave.net/", "Comma delimited ordered list of arweave gateways")
 	fs.BoolVar(&cli.MistCleanup, "run-mist-cleanup", true, "Run mist cleanup script")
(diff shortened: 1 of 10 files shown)


config/config.go
```diff
@@ -11,11 +10,6 @@ var Version string
 // Used so that we can generate fixed timestamps in tests
 var Clock TimestampGenerator = RealTimestampGenerator{}
 
-// Port that the local Broadcaster runs on
-const DefaultBroadcasterPort = 8935
-
-var DefaultBroadcasterURL = fmt.Sprintf("http://127.0.0.1:%d", DefaultBroadcasterPort)
```
thomshutt, line 17: What you've done works well and is the more textbook way to do this - over time I've been leaning more towards the "if it never changes at runtime (or in this case after CLI parsing), just stick it in a global variable and avoid having to pass clients down through the full stack of method signatures" school of thought but I don't feel too strongly about that

thomshutt (approved):

main.go
```diff
@@ -56,6 +56,7 @@ func main() {
 	config.URLSliceVarFlag(fs, &cli.ImportIPFSGatewayURLs, "import-ipfs-gateway-urls", "https://vod-import-gtw.mypinata.cloud/ipfs/?pinataGatewayToken={{secrets.LP_PINATA_GATEWAY_TOKEN}},https://w3s.link/ipfs/,https://ipfs.io/ipfs/,https://cloudflare-ipfs.com/ipfs/", "Comma delimited ordered list of IPFS gateways (includes /ipfs/ suffix) to import assets from")
 	config.URLSliceVarFlag(fs, &cli.ImportArweaveGatewayURLs, "import-arweave-gateway-urls", "https://arweave.net/", "Comma delimited ordered list of arweave gateways")
 	fs.BoolVar(&cli.MistCleanup, "run-mist-cleanup", true, "Run mist cleanup script")
+	fs.StringVar(&cli.BroadcasterURL, "broadcaster-url", "http://127.0.0.1:8935", "URL of local broadcaster")
```
thomshutt, line 59: Could we keep the default as a const in the Config file though?

PR #752, opened 2023-07-07
Fix display aspect ratio check

mjh1:
This fixes the check to account for negative differences. I should have added the customer example into the unit tests the first time around

### DIFF
--- a/pipeline/coordinator.go
+++ b/pipeline/coordinator.go
@@ -385,7 +385,7 @@ func checkDisplayAspectRatio(track video.InputTrack, requestID string) bool {
 	resRatio := float64(track.Width) / float64(track.Height)
 
 	// calculate the difference between the aspect ratio and the real ratio of the resolution, allow up to 20%
-	diff := dar - resRatio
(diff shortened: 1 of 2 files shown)


leszko (approved):

pipeline/coordinator.go
```diff
@@ -385,7 +385,7 @@ func checkDisplayAspectRatio(track video.InputTrack, requestID string) bool {
 	resRatio := float64(track.Width) / float64(track.Height)
 
 	// calculate the difference between the aspect ratio and the real ratio of the resolution, allow up to 20%
-	diff := dar - resRatio
+	diff := math.Abs(dar - resRatio)
```
emranemran, line 388: is 20% difference what ffmpeg allows here or arbitrary?

mjh1: Arbitrary from me looking at 1000 recent uploads: https://github.com/livepeer/catalyst-api/pull/734#discussion_r1243472648 look ok to you?

PR #753, opened 2023-07-07
Probe more segments after segmenting

mjh1:
- We found another case of bad (non-existing PPS 0 referenced) segments so we need to expand the checks to more segments https://linear.app/livepeer/issue/VID-270
- I found that we were ignoring the pps 0 error so I've refactored that to only apply to input videos, since the intention there was to ignore mist recording probe errors
- Included the fix for the other segmenting issue that has been found (audio corruption) https://linear.app/livepeer/issue/VID-287

### DIFF
--- a/pipeline/ffmpeg.go
+++ b/pipeline/ffmpeg.go
@@ -26,6 +26,7 @@ type ffmpeg struct {
 	SourceOutputUrl string
 	// Broadcaster for local transcoding
 	Broadcaster clients.BroadcasterClient
+	probe       video.Prober
 }
 
 func init() {
@@ -88,7 +89,7 @@ func (f *ffmpeg) HandleStartUploadJob(job *JobInfo) (*HandlerOutput, error) {
 	inputInfo := video.InputVideo{
 		Format:    job.InputFileInfo.Format,
 		Duration:  job.InputFileInfo.Duration,
-		SizeBytes: int64(job.sourceBytes),
+		SizeBytes: job.sourceBytes,
 		Tracks: []video.InputTrack{
 			// Video Track
 			{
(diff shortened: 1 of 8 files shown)


mjh1:
@thomshutt should we just check a handful of random segments too?

pipeline/ffmpeg.go
```diff
@@ -129,15 +129,13 @@ func (f *ffmpeg) HandleStartUploadJob(job *JobInfo) (*HandlerOutput, error) {
 	sourceSegments := sourceManifest.GetAllSegments()
 	job.sourceSegments = len(sourceSegments)
 
-	if job.sourceSegments > 0 && job.InputFileInfo.Format != "hls" {
-		firstSeg := sourceSegments[0]
-		lastSeg := sourceSegments[job.sourceSegments-1]
+	if job.sourceSegments > 1 && job.InputFileInfo.Format != "hls" {
+		segmentsToCheck := []int{0, 1, job.sourceSegments - 2, job.sourceSegments - 1}
```
emranemran, line 133: hmm this is still a workaround right? Did we ever narrow down the root cause? Was it a bug in ffmpeg segmenting?

thomshutt, line 133: Do we need range checks to catch e.g a single segment video?

mjh1: Yeah right now it just wouldn't enter the if-statement, will change that

mjh1: Seems to be an ffmpeg but yeah, will be raising a ticket.

emranemran: what's the bug specifically? what types of sources result in this bug?

pipeline/ffmpeg.go
```diff
@@ -129,15 +129,20 @@ func (f *ffmpeg) HandleStartUploadJob(job *JobInfo) (*HandlerOutput, error) {
...
-		firstSeg := sourceSegments[0]
-		lastSeg := sourceSegments[job.sourceSegments-1]
-
-		if err := probeSourceSegment(transcodeRequest.RequestID, firstSeg, transcodeRequest.SourceManifestURL); err != nil {
-			return nil, err
-		}
-		if err := probeSourceSegment(transcodeRequest.RequestID, lastSeg, transcodeRequest.SourceManifestURL); err != nil {
-			return nil, err
+	if job.InputFileInfo.Format != "hls" {
+		if job.sourceSegments < 4 {
```
mjh1, line 133: @thomshutt this feels so ugly but can't think of a nicer way

thomshutt: Seems fine to me!

thomshutt (approved):

pipeline/ffmpeg.go
emranemran: OK i'll add this to my backlog of video bugs to investigate. Seems alright for now.

PR #756, opened 2023-07-10
Set transcode bitrates as a proportion of input bitrate

mjh1:
As a way of addressing https://linear.app/livepeer/issue/VID-228/streameth-recording-uploaded-assets-returns-bad-quality
WDYT about something like this? Will need some careful investigation to check we don't negatively affect existing behaviour, perhaps by searching through previous ingests and testing with their bitrate values.


thomshutt:
Yeah, I think this seems conceptually solid but we should probably put together a batch of test cases

video/profiles.go
```diff
@@ -89,22 +89,23 @@ func GetPlaybackProfiles(iv InputVideo) ([]EncodedProfile, error) {
...
+		videoBitrate = MaxVideoBitrate
+	}
 	profiles := make([]EncodedProfile, 0, len(DefaultTranscodeProfiles)+1)
 	for _, profile := range DefaultTranscodeProfiles {
 		// transcoding job will adjust the width to match aspect ratio. no need to
 		// check it here.
-		lowerQualityThanSrc := profile.Height < video.Height && profile.Bitrate < video.Bitrate
+		lowerQualityThanSrc := profile.Height < video.Height
 		if lowerQualityThanSrc {
+			profile.Bitrate = (profile.Width * profile.Height * videoBitrate) / (video.Width * video.Height)
```
victorges, line 102: nit: I wonder if we need to make this more overflow-safe. There are still ~12 bits of buffer on our current max values (4k @ 288mbps), but these consts can change and this calculation become error prone.

Maybe multiply `videoBitrate` by `dstWidth/srcWidth` then `dstHeight/srcHeight` as floats then come back to an `int64`?

mjh1: good point 👍

mjh1:
I wrote the `TestCheckUpdatedAlgo` test which processes input bitrates and resolutions from a sample of 5000 jobs from prod, it shows that the only differences are lower bitrates like we're trying to achieve and the proportions look sensible. I also checked some examples visually and there wasn't a noticeable reduction in quality for the lower renditions. The top rendition will still match the input bitrate as we do now.

thomshutt (approved):

PR #765, opened 2023-07-19
Don't generate MP4s for long recordings until we can handle them better

thomshutt:


pipeline/coordinator.go
```diff
@@ -303,6 +298,21 @@ func (c *Coordinator) StartUploadJob(p UploadJobPayload) {
 		return nil, nil
 	})
 }
+
+func ShouldGenerateMP4(sourceURL, mp4TargetUrl *url.URL, mp4OnlyShort bool, durationSecs float64) bool {
```
thomshutt, line 302: Split this out into its own method to make testing easier

transcode/transcode_jobs.go
```diff
@@ -41,8 +40,6 @@ func (t *ParallelTranscoding) Start() {
 	t.completed.Add(config.TranscodingParallelJobs)
 	for index := 0; index < config.TranscodingParallelJobs; index++ {
 		go t.workerRoutine()
-		// Add some desync interval to avoid load spikes on segment-encode-end
-		time.Sleep(config.TranscodingParallelSleep)
```
thomshutt, line 45: Unrelated but I spotted this - I think this was us being concerned about a problem that doesn't exist and is safe to remove

pipeline/coordinator.go
```diff
@@ -303,6 +298,21 @@ func (c *Coordinator) StartUploadJob(p UploadJobPayload) {
...
+	// We're currently memory-bound for generating MP4s above a certain file size
+	// This has been hitting us for long recordings, so do a crude "is it longer than 3 hours?" check and skip the MP4 if it is
+	const maxRecordingMP4Duration = 3 * time.Hour
```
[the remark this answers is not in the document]
thomshutt: My thinking was that for Eli's super HD streams though, 3hr is still quite a lot. I didn't want to get into the weeds of trying to factor in source resolution too, but maybe we should do that - WDYT?

PR #766, opened 2023-07-19
Live Multistream Target Switching

leszko:
Adds support for Live Multistream Target Switching. Related [Technical Design Doc](https://www.notion.so/livepeer/Live-Multistream-Target-Switching-Simplified-Design-Doc-575e0150a0084fceb32d7d630dcb69db#4e6c63683a674e68947131f5c892f4d1).

--------------------------
**Changes:**
- Add endpoint `/api/events` to receive events that should be distributed to all Catalyst nodes
- Handle Serf user events to call `RefreshMutlistreamIfNeeded` on mistapiconnector
- Change Mist Client to use `PUSH_AUTO_ADD` (and `PUSH_AUTO_LIST`, `PUSH_AUTO_REMOVE`) instead of `PUSH_START`
- Add `reconcileMultistreamLoop()` to mistapiconnector (which is responsible for keeping Mist autopushes in sync with the cached `streamInfo`
- Remove `mist_client_mock.go` (use `gomock` instead)

-------------------------
**Flow:**
1. When multistream target is updated in Studio DB, then Studio calls Catalyst API endpoint `/api/events`
2. Catalyst API distributes the received event to all Catalyst nodes using serf
3. Each Catalyst API node receives the serf event and checks if it has a stream with the given `playbackId`
4. A Catalyst node that has a stream with the given `playbackId` refreshes its `streamInfo` cache by making calls to Studio API
5. Catalyst triggers the `reconcileMultistream` (which is called anyway every `1 minute` to monitor if Mist didn't remove any `PUSH_AUTO`)
6. Reconile Multistream updates the Mist state, so it has the same `PUSH_AUTO` as the current state of `streamInfo`


cluster/cluster.go
```diff
@@ -23,12 +23,15 @@ type Cluster interface {
...
+	EventChan() <-chan serf.UserEvent
+	BroadcastEvent(serf.UserEvent) error
 }
 
 type ClusterImpl struct {
 	config   *config.Cli
 	serf     *serf.Serf
-	eventCh  chan serf.Event
+	serfCh   chan serf.Event
+	eventCh  chan serf.UserEvent
```
thomshutt, line 34: While you're in here, could you add some comments around what the channels are used for and the flow etc. please?

mapic/mistapiconnector_app.go
```diff
@@ -423,38 +445,107 @@ func (mc *mac) shouldEnableAudio(stream *api.Stream) bool {
...
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+		case <-mc.multistreamUpdated:
+		}
+		mc.reconcileMultistream()
+	}
+}
+
+func (mc *mac) reconcileMultistream() {
```
thomshutt, line 462: Similarly here, the PR description is great but it would be good to get some commentary into the code about how we're reconciling the streams vs. Mist's records of them, why we need to call this on a loop etc.

thomshutt (approved):

cluster/cluster.go
leszko: Updated in 351b6385c825aceb2cf33c6897e8e88a3ee84076

mapic/mistapiconnector_app.go
leszko: Updated in 351b6385c825aceb2cf33c6897e8e88a3ee84076

handlers/events.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/xeipuuv/gojsonschema"
+	"io"
+	"net/http"
+)
+
+type EventsHandlersCollection struct {
+	Cluster cluster.Cluster
+}
+
+func (d *EventsHandlersCollection) Events() httprouter.Handle {
```
victorges, line 19: Is there any authentication on this API? Not a critical issue but might wanna consider adding something, if nothing yet (like the shared secret from task-runner <-> catalyst)

handlers/events_test.go
```diff
@@ -0,0 +1,70 @@
...
+			requestBody:  "",
+			wantHttpCode: 400,
+		},
+		{
+			requestBody:  "invalid payload",
+			wantHttpCode: 400,
+		},
+		{
+			requestBody: `{
+				"resource": "stream"
```
victorges, line 39: I know it will fail but might wanna add a test for different `resource` values as well

main.go
```diff
@@ -295,6 +301,40 @@ func reconcileBalancer(ctx context.Context, bal balancer.Balancer, c cluster.Clu
...
+	type EventPayload struct {
+		Resource   string `json:"resource"`
+		PlaybackID string `json:"playback_id"`
+	}
+
+	go func() {
+		var eventPayload EventPayload
+		err := json.Unmarshal(e.Payload, &eventPayload)
+		if err != nil {
+			glog.Errorf("cannot unmarshall received serf event: %v", e)
```
victorges, line 326: Should it `return` here?

mapic/mistapiconnector_app.go
```diff
@@ -423,38 +445,113 @@ func (mc *mac) shouldEnableAudio(stream *api.Stream) bool {
...
+	mistPushAutoList, err := mc.mist.PushAutoList()
+	if err != nil {
+		glog.Errorf("error executing PUSH_AUTO_LIST on Mist, cannot reconcile multistream err=%v", err)
+		return
+	}
+	var filteredMistPushAutoList []clients.MistPushAuto
+	mistMap := map[key]bool{}
+	for _, e := range mistPushAutoList {
+		k := toKey(e.Stream, e.Target)
+		if isMultistream(k) {
```
victorges, line 491: What other things does Mist return here which are not multistream? Is it the recordings bucket?

mapic/mistapiconnector_app.go
```diff
@@ -423,38 +445,113 @@ func (mc *mac) shouldEnableAudio(stream *api.Stream) bool {
...
+			filteredMistPushAutoList = append(filteredMistPushAutoList, e)
+			mistMap[toKey(e.Stream, e.Target)] = true
+		}
+	}
+
+	// Get the existing PUSH from Mist
+	stats, err := mc.mist.GetStats()
```
victorges, line 498: I think this is the function we had that used some global lock, from which we disabled the stats_collector from mapic! If we're bringing it back here, we might as well bring back the multistream metrics as well without much risk 👀 

OTOH, if we think the `stream_stats` command is causing issues, we might want to create a separate version of `GetStats` which only calls the `PUSH_LIST` but not the `STREAM_STATS` (iirc it calls both today)

mapic/mistapiconnector_app.go
```diff
@@ -423,38 +445,113 @@ func (mc *mac) shouldEnableAudio(stream *api.Stream) bool {
...
+		mc.reconcileMultistream()
+	}
+}
+
+// reconcileMultistream makes sure that Mist contains the multistream pushes exactly as specified in streamInfo cache.
+// There may be multiple reasons why Mist is not in sync with streamInfo cache:
+// - streamInfo cache has changed (multistream target was turned on/off or multistream target was added/removed)
+// - Mist removed its push for some reason
+// Note that we use Mist AUTO_PUSH (which in turn makes sure that the PUSH is always available).
+func (mc *mac) reconcileMultistream() {
```
victorges, line 468: Love this func!

mapic/mistapiconnector_app.go
```diff
@@ -509,43 +606,62 @@ func (mc *mac) getStreamInfo(playbackID string) (*streamInfo, error) {
...
+			metrics:          &data.MultistreamMetrics{},
+		}
+		if infoExists {
+			info.mu.Lock()
+			push, pushExists := info.pushStatus[pushURL]
+			info.mu.Unlock()
+			if pushExists {
+				newPushes[pushURL].metrics = push.metrics
+				newPushes[pushURL].pushStopped = push.pushStopped
+				newPushes[pushURL].pushStartEmitted = push.pushStartEmitted
```
victorges, line 647: Shouldn't we emit new push-start events if a push was disabled and then re-enabled? Or is this logic for some other case?

mapic/mistapiconnector_app_test.go
victorges, line 1: Nice!

mapic/stats_collector.go
```diff
@@ -77,6 +77,9 @@ func (c *metricsCollector) collectMetrics(ctx context.Context) error {
 		if err := ctx.Err(); err != nil {
 			return err
 		}
+		if streamID == "" {
```
victorges, line 80: When might this happen?

victorges (commented):
Late review 😀

cluster/cluster.go
```diff
@@ -234,21 +245,42 @@ func ResolveNodeURL(c Cluster, streamURL string) (string, error) {
...
+			case e := <-c.serfCh:
+				switch evt := e.(type) {
+				case serf.UserEvent:
+					select {
+					case <-ctx.Done():
+						return
+					case c.eventCh <- evt:
+						// Event moved to eventCh
+					default:
+						// Overflow event gets dropped
```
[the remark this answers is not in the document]
leszko: Added logs in aa0323d9ae4ae2a2a6f0313d60cccbf26467f784

handlers/events_test.go
leszko: Added in c5331d91362ad527de44f026d0406dd7f16b62f5

handlers/events.go
leszko: For now not. But I think we don't need it right now. YAGNI..hehe. We don't event know if this endpoint will need an authentication. If we move back to the idea of EIP-712 signed events, then this endpoint won't even require any auth.

main.go
leszko: Good point. Added in 6a732de290b2faa93d747fe43e75914c362fbafd

mapic/mistapiconnector_app.go
leszko: Currently, apart from multistream, auto push is used for recordings.

mapic/mistapiconnector_app.go
leszko: Ahh, I didn't know that calling stream_stats form Mist causes a global lock 😱 

Ok, then I think it's safer to not re-use it here. I'll create a separate `GetPushList()` function.

mapic/stats_collector.go
leszko: Good question. TBH Not sure right now. I think I added it because (for some reason) Mist returned `""` stream ID which caused panic.

mapic/mistapiconnector_app.go
leszko: I just realized that we're anyway calling this `GetStats` currently! So, I think it cannot cause so many issue. Then, probably let's leave it as it is.

PR #767, opened 2023-07-19
TEMP BRANCH: use application/x-www-form-urlencoded

iameli:
This ensures a smooth rollout of https://github.com/livepeer/catalyst-api/pull/764


iameli:
Never mind, let's upstream this for a while, we're not shipping the new Mist version for a bit

PR #768, opened 2023-07-19
cluster: allow subpaths in protocol urls

iameli-streams:
Currently we discard the subpath in a protocol redirect. This changes this, allowing us to redirect to, for example, a Mist hosted on a subdirectory like `/mist/hls/stream/index.m3u8`.


thomshutt:
@iameli what's the status of this? Can we close it?

iameli:
@thomshutt It was merged with something else.

PR #770, opened 2023-07-20
Local probe on timeout

mjh1:
This is a work in progress.
When we get a timeout on probing we should probe from local disk instead. After speaking to @thomshutt though we decided we might as well always copy to disk since it would only be mediaconvert jobs that wouldn't be copied to disk for segmenting.

Removing the file now also happens in the `Coordinator.finishJob()` function, It would be better/safer if we could do the local copy in the top level function in `Coordinator` so that we can still remove the file in a `defer()`, however I looked into this and there are goroutines spawned after the input probing happens so the `defer()` option wouldn't work as the function exits early.

Had to do some refactoring to simplify things which unfortunately pollutes the diff. I basically moved the main input_copy function into the pipeline package, this made it possible to pass in the `UploadJobPayload` struct and call the `CopySourceToDisk()` on it. It also means we can mutate the `UploadJobPayload` rather than having multiple return values (we can also remove some of the function params and read them from the struct).

### DIFF
--- a/clients/file_copy.go
+++ b/clients/file_copy.go
@@ -1,0 +1,128 @@
+package clients
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"path"
+	"strings"
+	"time"
+
+	"github.com/cenkalti/backoff/v4"
+	"github.com/hashicorp/go-retryablehttp"
+	"github.com/livepeer/catalyst-api/crypto"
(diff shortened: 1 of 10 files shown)


thomshutt (approved):
LGTM - once it's on Staging we should play around with different failure modes for files and make sure they're cleaned up, then also make sure we have disk usage monitoring in place to check it's not leaking once it's live

thomshutt:
Should we ship or close this?

PR #777, opened 2023-07-26
wip: clipping support

emranemran:


video/clip.go
```diff
@@ -0,0 +1,79 @@
...
+	for _, s := range relSegs {
+		if s.SeqId == seg.SeqId {
+			return true
+		}
+	}
+	return false
+}
+
+// Finds the segment in an HLS manifest that contains the timestamp (aka playhead) specified.
+func getRelevantSegment(allSegments []*m3u8.MediaSegment, relevantSegments *[]*m3u8.MediaSegment, playHeadTime float64, startingIdx uint64) (segNum uint64) {
```
thomshutt, line 36: Why is relevantSegments an input and not an output?

thomshutt: I think you can do it all with indexes at the point, so something like `getTimestampSegmentIndex(allSegments []*m3u8.MediaSegment, playHeadTime float64) (int, error)`

video/clip.go
```diff
@@ -0,0 +1,88 @@
+package video
+
+import (
+	"fmt"
+	"github.com/grafov/m3u8"
+	"github.com/livepeer/catalyst-api/log"
+)
+
+// Function to calculate the total duration of the HLS manifest
+/*func calculateTotalDuration(manifest *m3u8.MediaPlaylist) float64 {
```
thomshutt, line 10: Can we get rid of this?

emranemran: Done

video/clip.go
```diff
@@ -0,0 +1,72 @@
...
+		// Check if the playhead is within the current segment and skip to
+		// the next segment if it's not. Also update the play head by referencing
+		// the starting time of the next segment.
+		playHeadDiff = playHeadTime - segment.Duration
+		if playHeadDiff > 0.0 {
+			playHeadTime = playHeadDiff
+			continue
+		}
+		// Skip adding to slice of relevantSegments if the segment already exists
+		// (e.g. start and end times fall within same segment index)
```
mjh1, line 42: if this is the only case where this is possible maybe we could just increment `startingIdx` by one and exit if we've gone past the `endTime`. Or check that the current index is equal to `startingIdx`. Then we could avoid looping through `relevantSegments`

video/clip.go
mjh1: But maybe it would make more sense if `getRelevantSegment` just returned an index, then in `ClipManifest` it can form the `relevantSegments` slice when it knows both the start and end index.

emranemran: Alright, updated code to use just indices instead of passing pointers around.

video/clip.go
emranemran: Updated the logic to simplify all of this further.

emranemran:
**Updates**: 
* `ClipManifest` now returns a slice of `m3u8.MediaSegment` that contains **all** segments from the specified start time to end time
* `getRelevantSegment` now returns just indices and doesn't use any pointers (even though my brain sometimes prefers passing around pointers). 
* Added back `getTotalDurationAndSegments` function that's useful to simplify the range loop
* Updated tests to match the new logic

video/clip.go
```diff
@@ -0,0 +1,90 @@
...
+func getTotalDurationAndSegments(manifest *m3u8.MediaPlaylist) (float64, uint64) {
+	if manifest == nil {
+		return 0.0, 0
+	}
+	var totalDuration float64
+	var segmentCount uint64
+	for _, segment := range manifest.Segments {
+		if segment == nil {
+			break
+		}
```
mjh1, line 18: nit: just use `manifest.GetAllSegments()` then you can avoid this if statement and just take the length of that slice for segmentCount rather than needed to calculate yourself.

video/clip.go
```diff
@@ -0,0 +1,90 @@
...
+		endSegIdx = manifestSegments - 1
+	} else {
+		endSegIdx, err = getRelevantSegment(manifest.Segments, endTime, startSegIdx)
+		if err != nil {
+			return nil, fmt.Errorf("error clipping: failed to get an ending index")
+		}
+	}
+
+	// Generate a slice of all segments that overlap with startTime to endTime
+	relevantSegments := manifest.Segments[startSegIdx : endSegIdx+1]
```
mjh1, line 81: it looks like this `endSegIdx+1` will be out of range if we hit the line above `endSegIdx = manifestSegments - 1` is there a unit test covering it?

video/clip.go
emranemran: Ah good point, I seem to have missed that method already exists in the m3u8 library. Will update.

emranemran: Updated

mjh1 (approved):

PR #778, opened 2023-07-27
[SRE-160] CDN redirect

pwilczynskiclearcode:
https://linear.app/livepeer/issue/SRE-160/cdn-big-red-button
catalyst-api to support new `-cdn-redirect-playback-ids` and `-cdn-redirect-prefix` parameters.

This implementation requires changes to `catalyst-api` helm values and `catalyst` deployment.,
In future the list of CDN-redirected playbacks could be pulled from studio...

### DIFF
--- a/config/cli.go
+++ b/config/cli.go
@@ -59,6 +59,8 @@ type Cli struct {
 	GateURL                   string
 	StreamHealthHookURL       string
 	BroadcasterURL            string
+	CdnRedirectPrefix         *url.URL
+	CdnRedirectPlaybackIDs    []string
 }
 
 // Return our own URL for callback trigger purposes
--- a/handlers/geolocation/gelocation_test.go
+++ b/handlers/geolocation/gelocation_test.go
@@ -6,14 +6,17 @@ import (
 	"math/rand"
 	"net/http"
 	"net/http/httptest"
+	"net/url"
 	"testing"
 
 	"github.com/golang/mock/gomock"
 	"github.com/julienschmidt/httprouter"
 	"github.com/livepeer/catalyst-api/cluster"
 	"github.com/livepeer/catalyst-api/config"
+	"github.com/livepeer/catalyst-api/metrics"
 	mockbalancer "github.com/livepeer/catalyst-api/mocks/balancer"
 	mockcluster "github.com/livepeer/catalyst-api/mocks/cluster"
+	"github.com/prometheus/client_golang/prometheus/testutil"
 	"github.com/stretchr/testify/require"
 )
 
@@ -325,6 +328,44 @@ func TestNodeHostPortRedirect(t *testing.T) {
 		result(n).
 		hasStatus(http.StatusTemporaryRedirect).
 		hasHeader("Location", "https://right-host/any/path")
+}
+
+func TestCdnRedirect(t *testing.T) {
+	n := mockHandlers(t)
+	CdnRedirectedPlaybackId := "def_ZXY-999"
+	n.Config.NodeHost = "someurl.com"
+	n.Config.CdnRedirectPrefix, _ = url.Parse("https://external-cdn.com/mist")
+	n.Config.CdnRedirectPlaybackIDs = []string{CdnRedirectedPlaybackId}
+
+	// this one to be redirected to the closest node
+	requireReq(t, fmt.Sprintf("/hls/%s/index.m3u8", playbackID)).
+		result(n).
+		hasStatus(http.StatusTemporaryRedirect).
+		hasHeader("Location", fmt.Sprintf("http://%s/hls/%s/index.m3u8", closestNodeAddr, playbackID))
+
+	require.Equal(t, testutil.CollectAndCount(metrics.Metrics.CDNRedirectCount), 0)
+
+	// this playbackID is configured to be redirected to CDN
+	requireReq(t, fmt.Sprintf("/hls/%s/index.m3u8", CdnRedirectedPlaybackId)).
+		result(n).
+		hasStatus(http.StatusTemporaryRedirect).
+		hasHeader("Location", fmt.Sprintf("http://external-cdn.com/mist/hls/%s/index.m3u8", CdnRedirectedPlaybackId))
+
+	require.Equal(t, testutil.CollectAndCount(metrics.Metrics.CDNRedirectCount), 1)
+	require.Equal(t, testutil.ToFloat64(metrics.Metrics.CDNRedirectCount.WithLabelValues("unknown")), float64(0))
+	require.Equal(t, testutil.ToFloat64(metrics.Metrics.CDNRedirectCount.WithLabelValues(CdnRedirectedPlaybackId)), float64(1))
+
+}
+
+func TestCdnRedirectIncorrectPlaybackID(t *testing.T) {
+	n := mockHandlers(t)
+	n.Config.CdnRedirectPrefix, _ = url.Parse("https://external-cdn.com/mist")
+
+	// incorrect playbackID
+	requireReq(t, "/hls/index.m3u8/").
+		result(n).
+		hasStatus(http.StatusNotFound)
+
 }
 
 type httpReq struct {
--- a/handlers/geolocation/geolocation.go
+++ b/handlers/geolocation/geolocation.go
@@ -14,6 +14,7 @@ import (
 	"github.com/livepeer/catalyst-api/cluster"
 	"github.com/livepeer/catalyst-api/config"
 	"github.com/livepeer/catalyst-api/handlers/misttriggers"
+	"github.com/livepeer/catalyst-api/metrics"
 )
 
 type GeolocationHandlersCollection struct {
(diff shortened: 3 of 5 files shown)


handlers/geolocation/geolocation.go
```diff
@@ -26,29 +27,72 @@ type GeolocationHandlersCollection struct {
...
+			if hostNoPort == c.Config.CdnRedirectHost {
+				glog.V(6).Info("Request comes from CDN. Returning playback directly")
+				newURL, err := url.Parse(r.URL.String())
+				if err != nil {
+					glog.Errorf("failed to parse incoming url for redirect url=%s err=%s", r.URL.String(), err)
+					w.WriteHeader(http.StatusInternalServerError)
+					return
+				}
+				newURL.Scheme = protocol(r)
+				newURL.Host = c.Config.NodeHost
```
pwilczynskiclearcode, line 55: Not sure where should I redirect to avoid calling `Balancer.GetBestNode`?
`c.Config.NodeHost` makes sense but not sure if it was introduced for this purpose?
@iameli

pwilczynskiclearcode: Screenshot 2023-08-14 at 17 12 12

handlers/geolocation/geolocation.go
```diff
@@ -26,29 +27,51 @@ type GeolocationHandlersCollection struct {
...
-			host := r.Host
-			if host != nodeHost {
-				newURL, err := url.Parse(r.URL.String())
-				if err != nil {
-					glog.Errorf("failed to parse incoming url for redirect url=%s err=%s", r.URL.String(), err)
-					w.WriteHeader(http.StatusInternalServerError)
-					return
-				}
+			doCdnRedirect := contains(playbackID, c.Config.CdnRedirectPlaybackIDs)
+			if doCdnRedirect { // playbackID found in CdnRedirectPlaybackIDs
```
leszko, line 43: super nit: I think this comment does not explain a anything, it basically says the same what the code in the line above.
```suggestion
			if doCdnRedirect {
```

handlers/geolocation/geolocation.go
```diff
@@ -26,29 +27,51 @@ type GeolocationHandlersCollection struct {
...
+				newURL, _ := url.Parse(r.URL.String())
```
leszko, line 44: I think you should handle err coming from `url.Parse()` both here and two lines below.

handlers/geolocation/geolocation.go
```diff
@@ -26,29 +27,51 @@ type GeolocationHandlersCollection struct {
...
 				newURL.Scheme = protocol(r)
-				newURL.Host = nodeHost
+				cdnPrefix, _ := url.Parse(c.Config.CdnRedirectPrefix)
```
leszko, line 46: I think you can consider parsing it in `cli.go` then don't need to do it here on every coming request. The same way it's done, e.g., here: https://github.com/livepeer/catalyst-api/blob/a4ce9988472c249b333ae4117ca173185108b952/main.go#L56

handlers/geolocation/geolocation.go
```diff
@@ -26,29 +27,51 @@ type GeolocationHandlersCollection struct {
...
+				newURL.Host = cdnPrefix.Host
+				newURL.Path = cdnPrefix.Path + newURL.Path
```
leszko, line 48:
```suggestion
				newURL.Path = path.Join(cdnPrefix.Path,newURL.Path)
```

config/cli.go
```diff
@@ -59,6 +59,8 @@ type Cli struct {
 	GateURL                   string
 	StreamHealthHookURL       string
 	BroadcasterURL            string
+	CdnRedirectPrefix         string
```
leszko, line 62: As in some other comment below, I think you could have here the `*url.URL`, like this:

```go
CdnRedirectURL *url.URL
```

handlers/geolocation/gelocation_test.go
```diff
@@ -327,6 +329,44 @@ func TestNodeHostPortRedirect(t *testing.T) {
...
 }
 
+func TestCdnRedirect(t *testing.T) {
+	n := mockHandlers(t)
+	CdnRedirectedPlaybackId := "def_ZXY-999"
+	n.Config.NodeHost = "someurl.com"
+	n.Config.CdnRedirectPrefix = "https://external-cdn.com/mist"
+	n.Config.CdnRedirectPlaybackIDs = []string{CdnRedirectedPlaybackId}
+
+	// this one to be redirected to closest node
```
leszko, line 339:
```suggestion
	// this one to be redirected to the closest node
```

handlers/geolocation/gelocation_test.go
```diff
@@ -327,6 +329,44 @@ func TestNodeHostPortRedirect(t *testing.T) {
 		hasHeader("Location", "https://right-host/any/path")
 }
 
+func TestCdnRedirect(t *testing.T) {
```
leszko, line 332: Nice test 😀

leszko (approved):
Added some comments, mostly minor. 

Also, do I understand correctly that we the implementation is slightly different than what you described in the design doc [CDN “Big Red Button”](https://www.notion.so/livepeer/CDN-Big-Red-Button-8e88b7f422994176a3cc8ba9773ceb16), because you actually don't check anywhere the if the host header is coming from CDN. Is it because, the URL coming from CDN is always going directly to Mist (and not this updated `RedirectHandler()`)?

handlers/geolocation/geolocation.go
```diff
@@ -26,29 +27,51 @@ type GeolocationHandlersCollection struct {
...
-	return func(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
-		nodeHost := c.Config.NodeHost
-		redirectPrefixes := c.Config.RedirectPrefixes
+	return func(w http.ResponseWriter, r *http.Request, params httprouter.Params) {
+		host := r.Host
+		prefix, playbackID, pathTmpl, isValid := parsePlaybackID(r.URL.Path)
+
+		if c.Config.CdnRedirectPrefix != "" {
+
+			if !isValid {
```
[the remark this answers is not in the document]
pwilczynskiclearcode: I'm not sure what's the intention of [these tests](https://github.com/livepeer/catalyst-api/blob/main/handlers/geolocation/gelocation_test.go#L280-L300) they test redirects without playbackID in the URL path.
I don't understand if we want to make sure that redirect works for whatever path or should we always expect playbackID.

handlers/geolocation/geolocation.go
pwilczynskiclearcode: I used `url.JoinPath()` instead

pwilczynskiclearcode:
@leszko Yes. We don't check header as initial design said. We do the routing on the nginx level. I updated the diagram

main.go
```diff
@@ -85,6 +85,8 @@ func main() {
 	fs.StringVar(&cli.NodeName, "node", hostname, "Name of this node within the cluster")
 	config.SpaceSliceFlag(fs, &cli.BalancerArgs, "balancer-args", []string{}, "arguments passed to MistUtilLoad")
 	fs.StringVar(&cli.NodeHost, "node-host", "", "Hostname this node should handle requests for. Requests on any other domain will trigger a redirect. Useful as a 404 handler to send users to another node.")
+	config.CommaSliceFlag(fs, &cli.CdnRedirectPlaybackIDs, "cdn-redirect-playback-ids", []string{}, "PlaybackIDs to be redirected to CDN.")
+	config.URLVarFlag(fs, &cli.CdnRedirectPrefix, "cdn-redirect-prefix", "", "CDN URL where streams selected by -cdn-redirect-playback-ids are redirected. E.g. https://externalcdn.livepeer.com/mist/")
```
pwilczynskiclearcode, line 89: [infra related changes](https://github.com/livepeer/livepeer-infra/pull/1541/files)

hjpotter92 (approved):
don't have enough context in golang to have any insightful review. the redirect handler changes seem to be flowing as we'd expect in real usage. and the test cases do cover it!

handlers/geolocation/geolocation.go
```diff
@@ -26,29 +27,51 @@ type GeolocationHandlersCollection struct {
 
 // redirect an incoming user to a node for playback or 404 handling
```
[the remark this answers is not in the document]
pwilczynskiclearcode: I tried my best to explain but I don't fully understand everything from this function

handlers/geolocation/geolocation.go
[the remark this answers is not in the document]
iameli-streams: @pwilczynskiclearcode The `nodeHost` stuff is the 404 handler! The thinking is this: `catalyst-node-1` crashes, but nginx is still online so the request gets handled by a different catalyst in the same cluster. So then this happens:

```
> GET https://catalyst-node-1/hls/video+foo/index.m3u8
(catalyst-node-2 thinks: uhhh, that's not me, I only handle requests for nodeHost=playback.livepeer.studio, let's send them there)
< 302 Location: https://playback.livepeer.studio/hls/video+foo/index.m3u8

> GET https://playback.livepeer.studio/hls/video+foo/index.m3u8
(catalyst-node-2 thinks: that's better! Let's send them to the regular geolocation loop... looks like catalyst-node-5 isn't handling much traffic right now.)
< 302 Location: https://catalyst-node-5/hls/video+foo/index.m3u8

> GET https://catalyst-node-5/hls/video+foo/index.m3u8
< 200 OK
```

I coded that process to work for any arbitrary URL just in case. If it needs to work with just a playback id, that's okay too.

handlers/geolocation/geolocation.go
```diff
@@ -24,38 +25,59 @@ type GeolocationHandlersCollection struct {
...
+			if doCdnRedirect {
+				newURL, _ := url.Parse(r.URL.String())
 				newURL.Scheme = protocol(r)
-				newURL.Host = nodeHost
+				newURL.Host = c.Config.CdnRedirectPrefix.Host
+				newURL.Path, _ = url.JoinPath(c.Config.CdnRedirectPrefix.Path, newURL.Path)
 				http.Redirect(w, r, newURL.String(), http.StatusTemporaryRedirect)
-				glog.V(6).Infof("NodeHost redirect host=%s nodeHost=%s from=%s to=%s", host, nodeHost, r.URL, newURL)
+				glog.V(6).Infof("CDN redirect host=%s from=%s to=%s", r.Host, r.URL, newURL)
+				metrics.Metrics.CDNRedirectCount.WithLabelValues(playbackID).Inc()
```
iameli-streams, line 50: Metrics wooooo

handlers/geolocation/geolocation.go
```diff
@@ -179,3 +201,12 @@ func protocol(r *http.Request) string {
 	}
 	return "http"
 }
+
+func contains[T comparable](v T, list []T) bool {
```
iameli-streams, line 205: Generics! Nice.

PR #781, opened 2023-07-28
clients: update manifest ordering logic to handle 4K resolutions

emranemran:
HLS players begin playing a stream by typically loading the first rendition listed in a master playlist. The player will attempt downloading a few segments from the first rendition before playback can begin. For 4K resolutions, the segments can be fairly large in size (~20MB in some cases) and this results in a very slow TTFF. In order to optimize quick playback, we prioritize a rendition lower than 4K and then let the HLS player adjust quality as playback continues -- this ensures TTFF are short and the viewer sees playback begin quickly.

Fix VID-222

### DIFF
--- a/clients/manifest.go
+++ b/clients/manifest.go
@@ -108,6 +108,13 @@ func GenerateAndUploadManifests(sourceManifest m3u8.MediaPlaylist, targetOSURL s
 		}
 	})
 
+	// If the first rendition is 4k or greater resolution, then swap with the second rendition. HLS players
+	// typically load the first rendition in a master playlist and this can result in long downloads (and
+	// hence long TTFF) for high-res video segments.
+	if len(transcodedStats) >= 2 && (transcodedStats[0].Width >= 3840 || transcodedStats[0].Height >= 3840) {
+		transcodedStats[0], transcodedStats[1] = transcodedStats[1], transcodedStats[0]
+	}
+
 	for i, profile := range transcodedStats {
 		// For each profile, add a new entry to the master manifest
 		masterPlaylist.Append(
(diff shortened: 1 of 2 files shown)


thomshutt (approved):
Looks great, very neat! I think we might want to even bring the threshold down a bit further, but let's see how this goes first

emranemran:
> Looks great, very neat! I think we might want to even bring the threshold down a bit further, but let's see how this goes first

Yeah, agreed. There's just way too many resolution combos to handle but the common ones are here for video which I used as a reference: 
https://www.vdocipher.com/blog/video-resolution/

<h3 style="box-sizing: inherit; font-family: AvenirNext-Regular, Arial, sans-serif; font-weight: 600; line-height: 1.2; margin: 40px 0px 20px; font-size: 24px; color: rgb(20, 20, 20); font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><b style="box-sizing: inherit; font-weight: 700;">Video Resolution Chart</b></h3>

Resolution | Name | Aspect Ratio | Pixel Size
-- | -- | -- | --
SD (Standard Definition) | 480p | 4:3 | 640 x 480
HD (High Definition) | 720p | 16:9 | 1280 x 720
Full HD (FHD) | 1080p | 16:9 | 1920 x 1080
2K video (Quad HD) | 2k or 1440 | 16:9 | 2560 x 1440
4K video or Ultra HD (UHD) | 4K or 2160p | 1:1.9 | 3840 x 2160
8K video or Full Ultra HD | 8K or 4320p | 16∶9 | 7680 x 4320

PR #783, opened 2023-07-31
eli: misc bugfixes for livepeer-in-a-box

iameli:


balancer/balancer.go
```diff
@@ -46,7 +47,8 @@ type BalancerImpl struct {
 	cmd      *exec.Cmd
 	endpoint string
 	// Blocks until initial startup
-	startupOnce sync.Once
+	startupOnce  sync.Once
+	startupError error
```
leszko, line 51: Technically, if you use this field by multiple goroutines, it should be protected with a mutex.

leszko (approved):
Added one comment, other than that, LGTM 👍

balancer/balancer.go
iameli-streams: I think it is protected, in practice? If I understand correctly. The field is exclusively written to inside `b.startupOnce.Do`, so we can be confident there will only ever be one writer. And then it's only ever read _after_ the Do https://github.com/livepeer/catalyst-api/blob/66e8c0a02ead2bb77360c32e2eb666c7f2ea4a9c/balancer/balancer.go#L103, meaning that we'll have the same effect as if we used a RWMutex? And after that it will never be modified again.

PR #787, opened 2023-08-02
clipping: clip a segment

emranemran:
Part 2 in series of Clipping related PRs. 

This change adds logic to take any segment and clip it to the start/end time specified and also re-encodes it with some x264 args to re-encode and place I-frames at the right intervals. 

TODO: an integration test will be added in a separate PR.

### DIFF
--- a/video/clip.go
+++ b/video/clip.go
@@ -4,7 +4,16 @@ import (
 	"fmt"
 	"github.com/grafov/m3u8"
 	"github.com/livepeer/catalyst-api/log"
+	ffmpeg "github.com/u2takey/ffmpeg-go"
+	"time"
 )
+
+// format time in secs to be copatible with ffmpeg's expected time syntax
+func formatTime(seconds float64) string {
+	duration := time.Duration(seconds * float64(time.Second))
+	timeObj := time.Date(0, 0, 0, 0, 0, 0, 0, time.UTC).Add(duration)
+	return timeObj.Format("15:04:05")
+}
 
 func getTotalDurationAndSegments(manifest *m3u8.MediaPlaylist) (float64, uint64) {
 	if manifest == nil {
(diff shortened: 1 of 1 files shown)


video/clip.go
```diff
@@ -86,3 +98,24 @@ func ClipManifest(requestID string, manifest *m3u8.MediaPlaylist, startTime, end
...
+	err := ffmpeg.Input(tsInputFile).
+		Output(tsOutputFile, ffmpeg.KwArgs{"ss": start, "to": end,
+			"c:v":          "libx264",
+			"r":            "24",
+			"g":            "48",
+			"keyint_min":   "48",
+			"sc_threshold": 50,
+			"bf":           "0",
+			"c:a":          "copy"}).
+		OverWriteOutput().ErrorToStdOut().Run()
```
emranemran, line 115: Note: these settings will likely be revisited to be optimized before we release clipping. Using these reasonable defaults for now to finish integration.

video/clip.go
```diff
@@ -4,8 +4,20 @@ import (
 	"fmt"
 	"github.com/grafov/m3u8"
 	"github.com/livepeer/catalyst-api/log"
+	ffmpeg "github.com/u2takey/ffmpeg-go"
+	"time"
 )
 
+// format time in secs to be copatible with ffmpeg's expected time syntax
+func formatTime(seconds float64) string {
```
thomshutt, line 12: This works, but Go does have pretty good builtin time formatting

video/clip.go
```diff
@@ -86,3 +98,24 @@ func ClipManifest(requestID string, manifest *m3u8.MediaPlaylist, startTime, end
...
 }
+
+// Clips a segment by re-encoding so that I-frames are placed at the right intervals
+func ClipSegment(tsInputFile, tsOutputFile string, startTime, endTime float64) error {
+	start := formatTime(startTime)
+	end := formatTime(endTime)
+	err := ffmpeg.Input(tsInputFile).
+		Output(tsOutputFile, ffmpeg.KwArgs{"ss": start, "to": end,
+			"c:v":          "libx264",
+			"r":            "24",
```
thomshutt, line 109: Could we have some comments explaining all these values please?

emranemran: Done

video/clip.go
emranemran: Updated to use the builtin methods.

thomshutt (approved):

PR #796, opened 2023-08-07
Stop falling back based on audio codec

thomshutt:
We're initially going to check if the transcode pipeline already handles doing the audio transcode to AAC and if not, start doing that transcode inside Catalyst API


iameli (approved):
LGTM though I am not an expert on the VOD pipeline code

PR #806, opened 2023-08-14
Add support new output format (fragmented-MP4 packaged with HLS)

emranemran:
This PR adds the option to generate fragmented-mp4s packaged with HLS. 

**TODOs:**
~~* Add option to generate different mp4 types in vod request handler~~
~~* Fix location of output dir for fmp4 in object-store~~
~~* Fix mp4 output response to Studio~~
~~-- Fix probing issue post-transcode~~
~~-- Fix fmp4 output to generate separate a/v tracks~~ --> will be a followup PR
~~-- Fix/add unit tests~~ (separate PR will be used for integration tests)
~~-- Some more testing in staging including regression testing~~

/Fix VID-341

### DIFF
--- a/handlers/upload.go
+++ b/handlers/upload.go
@@ -23,9 +23,10 @@ import (
 )
 
 type UploadVODRequestOutputLocationOutputs struct {
-	HLS       string `json:"hls"`
-	MP4       string `json:"mp4"`
-	SourceMp4 bool   `json:"source_mp4"`
+	HLS           string `json:"hls"`
+	MP4           string `json:"mp4"`
+	FragmentedMP4 string `json:"fragmented_mp4"`
+	SourceMp4     bool   `json:"source_mp4"`
 }
 
 type UploadVODRequestOutputLocation struct {
@@ -91,6 +92,15 @@ func (r UploadVODRequest) getTargetMp4Output() (UploadVODRequestOutputLocation, 
 		}
 	}
 	return UploadVODRequestOutputLocation{}, false
+}
+
+func (r UploadVODRequest) getTargetFragMp4Output() UploadVODRequestOutputLocation {
+	for _, o := range r.OutputLocations {
+		if o.Outputs.FragmentedMP4 == "enabled" {
+			return o
+		}
+	}
+	return UploadVODRequestOutputLocation{}
 }
 
 func (r UploadVODRequest) getSourceCopyEnabled() bool {
@@ -165,9 +175,14 @@ func (d *CatalystAPIHandlersCollection) handleUploadVOD(w http.ResponseWriter, r
 	if err != nil {
 		return false, errors.WriteHTTPBadRequest(w, "Invalid request payload", err)
 	}
+	fragMp4TargetOutput := uploadVODRequest.getTargetFragMp4Output()
+	fragMp4TargetURL, err := toTargetURL(fragMp4TargetOutput, requestID)
+	if err != nil {
+		return false, errors.WriteHTTPBadRequest(w, "Invalid request payload", err)
+	}
 
-	if hlsTargetURL == nil && mp4TargetURL == nil {
-		return false, errors.WriteHTTPBadRequest(w, "Invalid request payload", errors2.New("none of output enabled: hls or mp4"))
+	if hlsTargetURL == nil && mp4TargetURL == nil && fragMp4TargetURL == nil {
+		return false, errors.WriteHTTPBadRequest(w, "Invalid request payload", errors2.New("none of output enabled: hls or mp4 or f-mp4"))
 	}
 
 	if strat := uploadVODRequest.PipelineStrategy; strat != "" && !strat.IsValid() {
@@ -184,6 +199,7 @@ func (d *CatalystAPIHandlersCollection) handleUploadVOD(w http.ResponseWriter, r
 		CallbackURL:           uploadVODRequest.CallbackUrl,
 		HlsTargetURL:          hlsTargetURL,
 		Mp4TargetURL:          mp4TargetURL,
+		FragMp4TargetURL:      fragMp4TargetURL,
 		Mp4OnlyShort:          mp4OnlyShort,
 		AccessToken:           uploadVODRequest.AccessToken,
 		TranscodeAPIUrl:       uploadVODRequest.TranscodeAPIUrl,
--- a/transcode/transcode.go
+++ b/transcode/transcode.go
@@ -120,7 +120,7 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
 	transcodedStats := statsFromProfiles(transcodeProfiles)
 
 	renditionList := video.TRenditionList{RenditionSegmentTable: make(map[string]*video.TSegmentList)}
-        // Only populate video.TRenditionList map if MP4/FragmentedMP4 is enabled or short-form video detection.
+	// Only populate video.TRenditionList map if MP4/FragmentedMP4 is enabled or short-form video detection.
 	// And if the original input file was an HLS video, then only generate an MP4 for the highest bitrate profile.
 	var maxBitrate int64
 	var maxProfile video.EncodedProfile
@@ -175,10 +175,19 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
 	var mp4OutputsPre []video.OutputVideoFile
 	// Transmux received segments from T into a single mp4
 	if transcodeRequest.GenerateMP4 {
-		mp4TargetUrlBase, err := url.Parse(transcodeRequest.Mp4TargetUrl)
+		// Check if we should generate a standard MP4, fragmented MP4, or both.
+		mp4TargetUrlBase, enableStandardMp4, err := getMp4OutputType(transcodeRequest.Mp4TargetUrl)
 		if err != nil {
 			return outputs, segmentsCount, err
 		}
+		fragMp4TargetUrlBase, enableFragMp4, err := getMp4OutputType(transcodeRequest.FragMp4TargetUrl)
+		if err != nil {
+			return outputs, segmentsCount, err
+		}
+		if !(enableStandardMp4 || enableFragMp4) {
+			return outputs, segmentsCount, fmt.Errorf("a valid mp4 or fragmented-mp4 URL must be provided since MP4 output was requested")
+		}
+
 		for rendition, segments := range renditionList.RenditionSegmentTable {
 			// a. create folder to hold transmux-ed files in local storage temporarily
 			err := os.MkdirAll(TRANSMUX_STORAGE_DIR, 0700)
@@ -299,6 +308,21 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
 	outputs = []video.OutputVideo{output}
 	// Return outputs for .dtsh file creation
 	return outputs, segmentsCount, nil
+}
+
+// getMp4OutputType checks the target url of the MP4 or Fragmented-MP4
+// output location and returns an *url.URL along with a boolean to
+// indicate that specific output type (mp4 or f-mp4) has been enabled.
+func getMp4OutputType(targetUrl string) (*url.URL, bool, error) {
+	if len(targetUrl) != 0 {
+		targetUrlBase, err := url.Parse(targetUrl)
+		if err != nil {
+			return nil, false, err
+		} else {
+			return targetUrlBase, true, nil
+		}
+	}
+	return nil, false, fmt.Errorf("mp4 output url is not set")
 }
 
 // getHlsTargetURL extracts URL for storing rendition HLS segments.
--- a/transcode/transcode.go
+++ b/transcode/transcode.go
@@ -29,6 +29,7 @@ type TranscodeSegmentRequest struct {
 	SourceOutputURL   string                 `json:"source_output_url"`
 	HlsTargetURL      string                 `json:"target_url"`
 	Mp4TargetUrl      string                 `json:"mp4_target_url"`
+	FragMp4TargetUrl  string                 `json:"fragmented_mp4_target_url"`
 	StreamKey         string                 `json:"streamKey"`
 	AccessToken       string                 `json:"accessToken"`
 	TranscodeAPIUrl   string                 `json:"transcodeAPIUrl"`
@@ -119,7 +120,7 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
 	transcodedStats := statsFromProfiles(transcodeProfiles)
 
 	renditionList := video.TRenditionList{RenditionSegmentTable: make(map[string]*video.TSegmentList)}
-	// Only populate video.TRenditionList map if MP4 is enabled via override or short-form video detection.
+        // Only populate video.TRenditionList map if MP4/FragmentedMP4 is enabled or short-form video detection.
 	// And if the original input file was an HLS video, then only generate an MP4 for the highest bitrate profile.
 	var maxBitrate int64
 	var maxProfile video.EncodedProfile
--- a/transcode/transcode.go
+++ b/transcode/transcode.go
@@ -222,36 +222,41 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
 				break
 			}
 
-			// d. Transmux the single .ts file into an .mp4 file
-			mp4OutputFileName := concatTsFileName[:len(concatTsFileName)-len(filepath.Ext(concatTsFileName))] + ".mp4"
-			err = video.MuxTStoMP4(concatTsFileName, mp4OutputFileName)
-			if err != nil {
-				log.Log(transcodeRequest.RequestID, "error transmuxing", "err", err)
-				continue
+			// d. Mux the .ts file to generate either a regular MP4 (w/ faststart) or fMP4 packaged with HLS/DASH
+			if enableStandardMp4 {
+				// Transmux the single .ts file into an mp4 file
+				mp4OutputFileName := concatTsFileName[:len(concatTsFileName)-len(filepath.Ext(concatTsFileName))] + ".mp4"
+				standardMp4OutputFiles, err := video.MuxTStoMP4(concatTsFileName, mp4OutputFileName)
+				if err != nil {
+					log.Log(transcodeRequest.RequestID, "error transmuxing to regular mp4", "file", mp4OutputFileName, "err", err)
+					continue
+				}
+				// Upload the mp4 file
+				mp4Out, err := uploadMp4Files(transcodeRequest.RequestID, mp4TargetUrlBase, standardMp4OutputFiles, rendition)
+				if err != nil {
+					return outputs, segmentsCount, fmt.Errorf("error uploading transmuxed standard mp4 file: %s", err)
+				}
+				mp4OutputsPre = append(mp4OutputsPre, mp4Out...)
 			}
 
-			// e. Upload mp4 output file
-			mp4OutputFile, err := os.Open(mp4OutputFileName)
-			defer os.Remove(mp4OutputFileName)
-			if err != nil {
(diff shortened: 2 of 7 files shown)


handlers/upload.go
```diff
@@ -82,15 +82,17 @@ func (r UploadVODRequest) getTargetHlsOutput() UploadVODRequestOutputLocation {
 	return UploadVODRequestOutputLocation{}
 }
 
-func (r UploadVODRequest) getTargetMp4Output() (UploadVODRequestOutputLocation, bool) {
+func (r UploadVODRequest) getTargetMp4Output() (UploadVODRequestOutputLocation, bool, bool) {
```
emranemran, line 85: @leszko need your input on commit `3f9f5eb4013b2174ac76f088bba9e3788e6408eb`.  The `MP4` field in  `UploadVODRequestOutputLocation` is already set by studio when we need to generate an mp4 file. The change here adds an option to generate a regular old mp4 (MP4 set to `enabled`) or a fragmented mp4 (MP4 set to `fragmented`). In other words, we should only generate either a regular mp4 or fragmented-mp4. Do you see any issues or dependencies from Studio perspective?

thomshutt: I don't love this - that parameter is for controlling _whether_ an MP4 is generated, rather than the type. Could we add the same parameter but for fMP4? That also opens the door to generating both, if that's ever something customers want

transcode/transcode.go
```diff
@@ -212,36 +213,59 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
...
-			// e. Upload mp4 output file
-			mp4OutputFile, err := os.Open(mp4OutputFileName)
-			defer os.Remove(mp4OutputFileName)
-			if err != nil {
-				log.Log(transcodeRequest.RequestID, "error opening mp4", "file", mp4OutputFileName, "err", err)
-				break
+			// d. Mux the .ts file to generate either a regular MP4 (w/ faststart) or fMP4 packaged with HLS/DASH
+			var mp4OutputFiles []string
+			if !transcodeRequest.Mp4Fragmented {
+				fmt.Println("XXX regular MP4", transcodeRequest.RequestID)
```
thomshutt, line 219: Drop before merge

transcode/transcode.go
```diff
@@ -212,36 +213,59 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
...
+				// Transmux the single .ts file into an mp4 file
+				mp4OutputFileName := concatTsFileName[:len(concatTsFileName)-len(filepath.Ext(concatTsFileName))] + ".mp4"
+				standardMp4OutputFiles, err := video.MuxTStoMP4(concatTsFileName, mp4OutputFileName)
+				if err != nil {
+					log.Log(transcodeRequest.RequestID, "error transmuxing to regular mp4", "file", mp4OutputFileName, "err", err)
+					continue
+				}
+				mp4OutputFiles = append(mp4OutputFiles, standardMp4OutputFiles...)
+			} else {
+				fmt.Println("XXX frag MP4", transcodeRequest.RequestID)
```
thomshutt, line 229: Drop

transcode/transcode.go
```diff
@@ -212,36 +213,59 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
...
-
-			// e. Upload mp4 output file
-			mp4OutputFile, err := os.Open(mp4OutputFileName)
-			defer os.Remove(mp4OutputFileName)
-			if err != nil {
-				log.Log(transcodeRequest.RequestID, "error opening mp4", "file", mp4OutputFileName, "err", err)
-				break
+			// d. Mux the .ts file to generate either a regular MP4 (w/ faststart) or fMP4 packaged with HLS/DASH
+			var mp4OutputFiles []string
+			if !transcodeRequest.Mp4Fragmented {
```
thomshutt, line 218: Could we flip this so that we're not checking for a negative? Or like in my other comment, maybe support both?

thomshutt:
Conceptually looks good, I like the splitting out of the two different MP4 generation methods rather than trying to combine them. Tag me once the tests are passing and it's ready for a final review.

handlers/upload.go
emranemran: Will update.

transcode/transcode.go
emranemran: Yeah, forgot to take out debug prints.

transcode/transcode.go
emranemran: Yeah, will update this.

handlers/upload.go
emranemran: Added new parameter for fMP4 which required a refactor - see latest commits (still wip but commit `f467915c364a0032847dec5f1445d2316570d29f` should be good for review).

emranemran:
**Updates (still WIP)**
* refactored code to allow both regular MP4 and fragmented MP4 generation
* some debug prints need to be cleaned up still and tests need to be fixed/added. 
* if you are reviewing, then going commit by commit will be best as I tried to split them up logically for easier review.

leszko (commented):
Added some early comments. I can review it again when you add all TODO point, but in general it looks good.

transcode/transcode.go
```diff
@@ -174,10 +175,19 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
...
+		// Check if we should generate a standard MP4, fragmented MP4, or both.
+		mp4TargetUrlBase, enableStandardMp4, err := getMp4OutputType(transcodeRequest.Mp4TargetUrl)
 		if err != nil {
 			return outputs, segmentsCount, err
 		}
+		fragMp4TargetUrlBase, enableFragMp4, err := getMp4OutputType(transcodeRequest.FragMp4TargetUrl)
+		if err != nil {
+			return outputs, segmentsCount, err
+		}
+		if !(enableStandardMp4 || enableFragMp4) {
```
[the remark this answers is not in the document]
emranemran: In case both output URLs for mp4 and fmp4 is malformed somehow in the incoming request. There are additional layers of checks elsewhere in catalyst-api and probably Studio as well. I can remove if you think it's redundant but thought we keep an extra layer of security here.

transcode/transcode.go
```diff
@@ -253,8 +275,14 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
 	var mp4Outputs []video.OutputVideoFile
 	if transcodeRequest.GenerateMP4 {
 		for _, mp4Out := range mp4OutputsPre {
+			fmt.Println("XXX: ", transcodeRequest.RequestID, "looking at", mp4Out)
 			mp4Out.Location = strings.ReplaceAll(mp4Out.Location, transcodeRequest.Mp4TargetUrl, mp4PlaybackBaseURL)
 
+			// TODO: Ignore f-mp4 output files for now. A follow-on PR will be used to address f-mp4 files in response to Studio 
+			if filepath.Ext(mp4Out.Location) != ".mp4" {
```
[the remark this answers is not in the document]
emranemran: Not sure I understand. If mp4 output is detected, we will skip this ` if` block. When fmp4 output is requested, the files are `m4a` and ` m3u8` so I want to skip over those temporarily until the Studio integration is done.

transcode/transcode.go
```diff
@@ -300,6 +328,53 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
...
+		}
+		mp4OutputsPre = append(mp4OutputsPre, mp4Out)
+	}
+	return mp4OutputsPre, nil
+}
+
+// getMp4OutputType checks the target url of the MP4 or Fragmented-MP4
+// output location and returns an *url.URL along with a boolean to
+// indicate that specific output type (mp4 or f-mp4) has been enabled.
+func getMp4OutputType(targetUrl string) (*url.URL, bool, error) {
```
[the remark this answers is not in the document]
emranemran: `bool` here means that the given target url is valid so enable that type of output. I thought about using an enum here but I think that makes the logic a bit more convoluted here given the usage further down the code where we actually use these `bool`s.

emranemran:
@leszko @thomshutt @mjh1 ready for one more round of review. 

Before merge: 
* run regression test suite

Post merge: 
* add integration tests for the transmux methods
* staging tests to check for regressions

leszko (commented):
Added a few more comments, but in general it looks good. I played around with this and was able to generate the fmp4 files 🎉 

One general question, I had the following path: `/cakeout/hls/` set in my Storj bucket. Then, I've send the following request:
```
            "outputs": {
                "hls": "enabled",
                "mp4": "enabled",
                "fragmented_mp4": "enabled"
            }
```

Now, I was a little surprised by where the output files are located. Is that correct?
- HLS files are in `/cakeout/hls/index.m3u8`, `cakeout/hls/360p0/*`, `cakeout/hls/720p0/*`, `cakeout/hls/1080p0/*`=> ALL GOOD
- MP4 files are in `/cakeout/hls/360p0.mp4`, `/cakeout/hls/720p0.mp4`, `/cakeout/hls/1080p0.mp4` => ALL GOOD
- fMP4 files are in `/cakeout/hls/360p0/fmp4/360p0.m3u8`, `/cakeout/hls/360p0/fmp4/360p0.m4s` (and the corresponding dirs for `720p0` and `1080p0` => Is that correct? Don't we need some manifest files in the root dir like we have for HLS?

transcode/transcode.go
```diff
@@ -212,38 +222,50 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
...
-
-			// e. Upload mp4 output file
-			mp4OutputFile, err := os.Open(mp4OutputFileName)
-			defer os.Remove(mp4OutputFileName)
-			if err != nil {
-				log.Log(transcodeRequest.RequestID, "error opening mp4", "file", mp4OutputFileName, "err", err)
-				break
+			// d. Mux the .ts file to generate either a regular MP4 (w/ faststart) or fMP4 packaged with HLS/DASH
+			if enableStandardMp4 {
+				fmt.Println("XXX regular MP4", transcodeRequest.RequestID)
```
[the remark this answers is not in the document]
emranemran: Oops, forgot about all these.

transcode/transcode.go
```diff
@@ -212,38 +222,50 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
...
-				continue
-			}
-
-			// e. Upload mp4 output file
-			mp4OutputFile, err := os.Open(mp4OutputFileName)
-			defer os.Remove(mp4OutputFileName)
-			if err != nil {
-				log.Log(transcodeRequest.RequestID, "error opening mp4", "file", mp4OutputFileName, "err", err)
-				break
+			// d. Mux the .ts file to generate either a regular MP4 (w/ faststart) or fMP4 packaged with HLS/DASH
```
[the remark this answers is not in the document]
emranemran: I'll clean this up. It's actually documenting the steps in MP4 generation if you scroll up a few comments.

handlers/upload.go
```diff
@@ -23,9 +23,10 @@ import (
 )
 
 type UploadVODRequestOutputLocationOutputs struct {
-	HLS       string `json:"hls"`
-	MP4       string `json:"mp4"`
-	SourceMp4 bool   `json:"source_mp4"`
+	HLS           string `json:"hls"`
+	MP4           string `json:"mp4"`
+	FragmentedMP4 string `json:"fragmented_mp4"`
```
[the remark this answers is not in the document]
emranemran: I found it useful to break it up here as some of the code becomes hard to distinguish between MP4 and FMP when you're reading through it.

video/transmux.go
```diff
@@ -4,17 +4,71 @@ import (
...
+	_, err = os.Stat(mp4OutputFile)
+	if err != nil {
+		return nil, fmt.Errorf("transmux error: failed to stat MP4 media file: %s", err)
+	} else {
+		transmuxOutputFiles = append(transmuxOutputFiles, mp4OutputFile)
+	}
+	return transmuxOutputFiles, nil
+}
+
+func MuxTStoFMP4WithHLS(tsInputFile, fmp4ManifestOutputFile string) ([]string, error) {
```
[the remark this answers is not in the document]
emranemran: Yeah, i'll add this function as a test here as a follow-up PR after some staging tests: 
https://github.com/livepeer/catalyst-api/blob/cc8e4d396060494f1b7b6fee0a416be43f08b56f/test/features/vod.feature#L51

emranemran:
> Added a few more comments, but in general it looks good. I played around with this and was able to generate the fmp4 files 🎉
> 
> One general question, I had the following path: `/cakeout/hls/` set in my Storj bucket. Then, I've send the following request:
> 
> ```
>             "outputs": {
>                 "hls": "enabled",
>                 "mp4": "enabled",
>                 "fragmented_mp4": "enabled"
>             }
> ```
> 
> Now, I was a little surprised by where the output files are located. Is that correct?
> 
> * HLS files are in `/cakeout/hls/index.m3u8`, `cakeout/hls/360p0/*`, `cakeout/hls/720p0/*`, `cakeout/hls/1080p0/*`=> ALL GOOD
> * MP4 files are in `/cakeout/hls/360p0.mp4`, `/cakeout/hls/720p0.mp4`, `/cakeout/hls/1080p0.mp4` => ALL GOOD
> * fMP4 files are in `/cakeout/hls/360p0/fmp4/360p0.m3u8`, `/cakeout/hls/360p0/fmp4/360p0.m4s` (and the corresponding dirs for `720p0` and `1080p0` => Is that correct? Don't we need some manifest files in the root dir like we have for HLS?

Yep, will fix this based on offline discussion with Rafal/Thom/Max

leszko:
The current directory structure is like this:
image

So all the fmp4 files are embedded into the `fmp4` directory. I think it's ok, but may be a little confusing for the Transcode API users. Imagine you send the following request:
```
{
    "input": {
        "url": "some-video-url"
    },
    "storage": {
        "type": "s3",
        "endpoint": "https://gateway.storjshare.io",
        "credentials": {
            "accessKeyId": "access-key",
            "secretAccessKey": "secret-key"
        },
        "bucket": "my-bucket"
    },
    "outputs": {
        "hls": {
            "path": "/hls/"
        },
        "mp4": {
            "path": "/mp4/"
        },
        "fragmented_mp4": {
           "path": "/fmp4"
        }
    }
}
```

Now, you could probably expect that your fmp4 files will end it in the `/fmp4` location, but actually they'll end up in `/fmp4/fmp4` location.

I leave it up to you, but thinking about it again, I'd probably flatten the `fmp4` structure and then for our upload task, we would just add `/fmp4` in the request between task-runner<>catalyst.

leszko (approved):
LGTM 👍 

Added one comment again about the directory structure.

Plus, do we support (want to support) fmp4 in the mediaconvert pipeline?

transcode/transcode.go
```diff
@@ -300,6 +320,52 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
...
+		mp4OutputsPre = append(mp4OutputsPre, mp4Out)
+	}
+	return mp4OutputsPre, nil
+}
+
+// getMp4OutputType checks the target url of the MP4 or Fragmented-MP4
+// output location and returns an *url.URL along with a boolean to
+// indicate that specific output type (mp4 or f-mp4) has been enabled.
+func getMp4OutputType(targetUrl string) (*url.URL, bool, error) {
+	if len(targetUrl) != 0 {
```
mjh1, line 358: nit: Could you invert the if statement and return early to avoid the extra indentation?

transcode/transcode.go
```diff
@@ -300,6 +320,52 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
 	return outputs, segmentsCount, nil
 }
 
+func uploadMp4Files(requestID string, basePath *url.URL, mp4OutputFiles []string, rendition string) ([]video.OutputVideoFile, error) {
+	var mp4OutputsPre []video.OutputVideoFile
+	// e. Upload all mp4 related output files
+	for _, o := range mp4OutputFiles {
+		var filename string
+		mp4OutputFile, err := os.Open(o)
+		defer os.Remove(mp4OutputFile.Name())
```
mjh1, line 329: this needs to be moved down below the `err != nil` check

mjh1:
> The current directory structure is like this: image
> 
> So all the fmp4 files are embedded into the `fmp4` directory. I think it's ok, but may be a little confusing for the Transcode API users. Imagine you send the following request:
> 
> ```
> {
>     "input": {
>         "url": "some-video-url"
>     },
>     "storage": {
>         "type": "s3",
>         "endpoint": "https://gateway.storjshare.io",
>         "credentials": {
>             "accessKeyId": "access-key",
>             "secretAccessKey": "secret-key"
>         },
>         "bucket": "my-bucket"
>     },
>     "outputs": {
>         "hls": {
>             "path": "/hls/"
>         },
>         "mp4": {
>             "path": "/mp4/"
>         },
>         "fragmented_mp4": {
>            "path": "/fmp4"
>         }
>     }
> }
> ```
> 
> Now, you could probably expect that your fmp4 files will end it in the `/fmp4` location, but actually they'll end up in `/fmp4/fmp4` location.
> 
> I leave it up to you, but thinking about it again, I'd probably flatten the `fmp4` structure and then for our upload task, we would just add `/fmp4` in the request between task-runner<>catalyst.

Well spotted 👍 annoyingly I think we'll have to add a check to add the prefix `/fmp4` only if the prefixes for hls and fmp4 aren't the same, as I don't think we'd want the flat structure of regular hls and fmp4 flat in the same folder.

leszko:
> > Now, you could probably expect that your fmp4 files will end it in the `/fmp4` location, but actually they'll end up in `/fmp4/fmp4` location.
> > I leave it up to you, but thinking about it again, I'd probably flatten the `fmp4` structure and then for our upload task, we would just add `/fmp4` in the request between task-runner<>catalyst.
> 
> @leszko Well spotted 👍 annoyingly I think we'll have to add a check to add the prefix `/fmp4`if the prefixes for hls and fmp4 are the same, as I don't think we'd want the flat structure of regular hls and fmp4 in the same folder as they share a similar structure and might be confusing if they're mixed together.

Maybe, but for Transcode API we can leave it "up to the user" and for Upload Task just add the `/hls` prefix in task-runner.

mjh1:
> Maybe, but for Transcode API we can leave it "up to the user" and for Upload Task just add the `/hls` prefix in task-runner.

Ah i see what you're saying yeah that sounds good 👍

emranemran:
> The current directory structure is like this: image
> 
> So all the fmp4 files are embedded into the `fmp4` directory. I think it's ok, but may be a little confusing for the Transcode API users. Imagine you send the following request:
> 
> ```
> {
>     "input": {
>         "url": "some-video-url"
>     },
>     "storage": {
>         "type": "s3",
>         "endpoint": "https://gateway.storjshare.io",
>         "credentials": {
>             "accessKeyId": "access-key",
>             "secretAccessKey": "secret-key"
>         },
>         "bucket": "my-bucket"
>     },
>     "outputs": {
>         "hls": {
>             "path": "/hls/"
>         },
>         "mp4": {
>             "path": "/mp4/"
>         },
>         "fragmented_mp4": {
>            "path": "/fmp4"
>         }
>     }
> }
> ```
> 
> Now, you could probably expect that your fmp4 files will end it in the `/fmp4` location, but actually they'll end up in `/fmp4/fmp4` location.
> 
> I leave it up to you, but thinking about it again, I'd probably flatten the `fmp4` structure and then for our upload task, we would just add `/fmp4` in the request between task-runner<>catalyst.

Got it, i'll open a ticket for us to make this change. For now only a single customer will likely be using fmp4 output.

emranemran:
> LGTM 👍
> 
> Added one comment again about the directory structure.
> 
> Plus, do we support (want to support) fmp4 in the mediaconvert pipeline?

I think we probably want to in the future to keep both pipelines consistent. For now just livepeer pipeline should be sufficient. @thomshutt wdyt?

PR #809, opened 2023-08-17
Ignore empty files when copying multiple files

mjh1:
This is to guard against an issue we had with recordings where we saw an empty segment at the end of a manifest.

Verified this on staging with a problematic input: https://lp-us-catalyst-recordings-com.storage.googleapis.com/hls/b7caqhzl06ma2iqn/e0e9a30f-769c-4e50-9b69-2cf0ac97e68c/output.m3u8


mjh1:
> Will this result in an unfetchable file in the manifest though?

When I tested it was excluded from the final manifest after it had been sent through transcoding

thomshutt (approved):

PR #813, opened 2023-08-21
Generate master playlists for fragmented-mp4 media playlists

emranemran:
transcode/manifest: generate master fmp4 playlists and upload to object-store

### DIFF
--- a/clients/manifest.go
+++ b/clients/manifest.go
@@ -17,6 +17,7 @@ import (
 const (
 	MASTER_MANIFEST_FILENAME = "index.m3u8"
 	MANIFEST_UPLOAD_TIMEOUT  = 5 * time.Minute
+	FMP4_POSTFIX_DIR         = "fmp4"
 )
 
 func DownloadRetryBackoffLong() backoff.BackOff {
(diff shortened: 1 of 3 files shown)


transcode/transcode.go
```diff
@@ -385,6 +406,17 @@ func getHlsTargetURL(tsr TranscodeSegmentRequest) (*url.URL, error) {
 	}
 }
 
+func getFmp4TargetURL(tsr TranscodeSegmentRequest) (*url.URL, error) {
+	if tsr.FragMp4TargetUrl != "" {
```
mjh1, line 410: nit: reverse the logic and do `return nil, nil` here

clients/manifest.go
```diff
@@ -183,6 +183,55 @@ func GenerateAndUploadManifests(sourceManifest m3u8.MediaPlaylist, targetOSURL s
...
+		} else {
+			var resolutionA = transcodedStats[a].Width * transcodedStats[a].Height
+			var resolutionB = transcodedStats[b].Width * transcodedStats[b].Height
+
+			return resolutionA > resolutionB
+		}
+	})
+
+	// For each profile, add a new entry to the master manifest
+	for i, profile := range transcodedStats {
```
mjh1, line 206: Do we already have this logic for standard HLS that we could reuse?

mjh1 (approved):

transcode/transcode.go
emranemran: Will change it.

clients/manifest.go
emranemran: Yeah i just copied for now. i'll turn this into a function.

emranemran:
@mjh1 I'm handling the edits in a follow-up PR i'm doing - just wanted to merge this as the new PR depends on this and it's easier for me to make the changes in the new one.

PR #815, opened 2023-08-21
Add track durations to match task-runner

mjh1:
I'm correcting some differences I saw to task-runners probing logic so that we can use catalyst's output and remove the logic from task-runner

### DIFF
--- a/video/probe.go
+++ b/video/probe.go
@@ -13,7 +13,10 @@ import (
 	"gopkg.in/vansante/go-ffprobe.v2"
 )
 
-var unsupportedVideoCodecList = []string{"mjpeg", "jpeg", "png"}
+var (
+	unsupportedVideoCodecList = []string{"mjpeg", "jpeg", "png"}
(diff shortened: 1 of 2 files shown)


video/probe.go
```diff
@@ -231,3 +241,22 @@ func parseFps(framerate string) (float64, error) {
...
 }
+
+func findFormat(format string) string {
+	actualFormats := strings.Split(format, ",")
+	for _, f := range supportedFormats {
+		if containsStr(actualFormats, f) {
+			return f
+		}
+	}
+	return actualFormats[0]
```
leszko, line 252: What if `actualFormats` is empty? Or it won't ever happen?

leszko (approved):

video/probe.go
mjh1: Good question, I just checked and we should be ok as strings.Split will return `[""]` if `format` is empty, which will give us the same behaviour as before setting the format to `""`.

PR #818, opened 2023-08-22
Skip source copy for HLS input

mjh1:
This is to drastically speed up processing for recordings. A new config param was needed to allow source playback still to work, this is so that app knows what hostname to use in the source playback manifest (to reach the recordings bucket rather than the normal asset bucket). E.g. a source playback manifest for a recording will look something like:

```
#EXTM3U
#EXT-X-VERSION:3
#EXT-X-STREAM-INF:PROGRAM-ID=0,BANDWIDTH=4000000,RESOLUTION=1280x720,NAME="720p"
//recordings-cdn.lp-playback.studio/hls/b7caqhzl06ma2iqn/e0e9a30f-769c-4e50-9b69-2cf0ac97e68c/output.m3u8
```

### DIFF
--- a/clients/input_copy.go
+++ b/clients/input_copy.go
@@ -46,15 +46,20 @@ func NewInputCopy() *InputCopy {
 
 // CopyInputToS3 copies the input video to our S3 transfer bucket and probes the file.
 func (s *InputCopy) CopyInputToS3(requestID string, inputFile, osTransferURL *url.URL, decryptor *crypto.DecryptionKeys) (inputVideoProbe video.InputVideo, signedURL string, err error) {
-	err = CopyAllInputFiles(requestID, inputFile, osTransferURL, decryptor)
-	if err != nil {
-		err = fmt.Errorf("failed to copy file(s): %w", err)
-		return
-	}
+	if IsHLSInput(inputFile) {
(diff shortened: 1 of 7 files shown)


config/config.go
```diff
@@ -54,3 +54,8 @@ var ImportArweaveGatewayURLs []*url.URL
 var PrivateBucketURL *url.URL
 
 var HTTPInternalAddress string
+
+var DefaultSourcePlaybackHosts = map[string]string{
+	"lp-us-catalyst-recordings-com.storage.googleapis.com":     "//recordings-cdn.lp-playback.studio/hls",
```
thomshutt, line 59: Even though it's not strictly secret, I don't love having these exposed. Could we just fallback to some default behaviour if the CLI param isn't present?

thomshutt (approved):

config/config.go
mjh1: I think we'd have to just disable source playback in that case otherwise it'd require some tricky logic to then ensure the source copy does get done. I think that would be ok as we'd still have a big speed improvement from dropping the copy, but in theory we should be able to roll the cli param out at the same time and this won't be a problem.

PR #819, opened 2023-08-22
fix: active idle status & restore metrics collection

gioelecerati:


mapic/stats_collector.go
```diff
@@ -30,45 +31,27 @@ type metricsCollector struct {
 	infoProvider
 }
 
-func startMetricsCollector(ctx context.Context, period time.Duration, nodeID, ownRegion string, mapi clients.MistAPIClient, lapi *api.Client, producer event.AMQPProducer, amqpExchange string, infop infoProvider) {
+func createMetricsCollector(ctx context.Context, period time.Duration, nodeID, ownRegion string, mapi clients.MistAPIClient, lapi *api.Client, producer event.AMQPProducer, amqpExchange string, infop infoProvider) *metricsCollector {
```
victorges, line 34: We don't need this `ctx` argument anymore, that one was only used in the `mainLoop`. `period` is also unused. Make sure to remove any others.

mapic/stats_collector.go
```diff
@@ -87,11 +70,11 @@ func (c *metricsCollector) collectMetrics(ctx context.Context) error {
...
-			// avoid spamming metrics for playback-only catalyst instances. This means
-			// that if mapic restarts we will stop sending metrics from previous
-			// streams as well, but that's a minor issue (curr stream health is dying).
-			glog.Infof("Skipping metrics for lazily created stream info. streamId=%q metrics=%+v", streamID, metrics)
+
+		stream := c.infoProvider.wildcardPlaybackID(info.stream)
+		isIngest := isIngestStream(stream, info, mistState)
+
+		if !isIngest {
 			continue
```
victorges, line 78: WDYT of adding a verbose log here ([`glog.V(8)...`](https://github.com/livepeer/livepeer-infra/blob/ae887391834954e4809a5086f816e69a6ccd0d87/values/staging/catalyst.values.yaml#L21)) when a stream is skipped, so you can explicitly test that on staging? i.e.: playback a stream from a different node and make sure that node doesn't send metrics, only the ingest node does.

mapic/mistapiconnector_app.go
```diff
@@ -150,12 +151,12 @@ func (mc *mac) Start(ctx context.Context) error {
 		glog.Infof("AMQP url is empty!")
 	}
 	if producer != nil && mc.config.MistScrapeMetrics {
-		startMetricsCollector(ctx, statsCollectionPeriod, mc.nodeID, mc.ownRegion, mc.mist, lapi, producer, ownExchangeName, mc)
+		mc.metricsCollector = createMetricsCollector(ctx, statsCollectionPeriod, mc.nodeID, mc.ownRegion, mc.mist, lapi, producer, ownExchangeName, mc)
```
victorges, line 154: We used to run the stats collector loop every `statsCollectionPeriod` (10s) while the `reconcileLoop` runs every 60s. I think 60s is too much for updating the `isActive` property, as it is very close to the limit at which the Studio API will reset its value assuming the stream is unused (I think it's 70s there).

To avoid making the `isActive` field going back-and-forth, wdyt of going on a middle ground here, maybe `30s`? Also make sure to delete any unused consts like this `statsCollectionPeriod`, or update its name and value to the new abstractions.

@leszko WDYT? You think reducing the reconcile loop to 30s would be too short?

victorges (commented):
Will leave for @leszko to give the final approval.

mapic/mistapiconnector_app.go
gioelecerati: Changed it to 30s to test it in staging, will leave this conversation open for @leszko feedback

PR #825, opened 2023-08-25
Create empty USER_END trigger handler

thomshutt:
Writing the contents of the USER_END trigger to a database for analytics.


api/http_internal.go
```diff
@@ -113,6 +116,9 @@ func NewCatalystAPIRouterInternal(cli config.Cli, vodEngine *pipeline.Coordinato
 	// Handler for USER_NEW triggers
 	broker.OnUserNew(accessControlHandlers.HandleUserNew)
 
+	// Handler for USER_END triggers. Only do this if we have a database
```
iameli, line 119: I assume this is implemented in the handler, not really pertinent information
```suggestion
	// Handler for USER_END triggers.
```

thomshutt: Good catch, it _was_ here and then I realised it made more sense inside

handlers/analytics/user_end.go
```diff
@@ -0,0 +1,73 @@
...
+		"stream_id_duration_s",
+		"protocol",
+		"protocol_duration_s",
+		"ip_address",
+		"ip_address_duration_s",
+		"tags"
+		) values($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13)`
+	_, err := a.db.Exec(
+		insertDynStmt,
+		time.Now().Unix(),               // timestamp_ms
```
iameli, line 54: Liar!
```suggestion
		time.Now().UnixMilli(),               // timestamp_ms
```

handlers/analytics/user_end.go
```diff
@@ -0,0 +1,73 @@
...
+	return AnalyticsHandler{db: db}
+}
+
+func (a *AnalyticsHandler) HandleUserEnd(ctx context.Context, payload *misttriggers.UserEndPayload) error {
+	// If there's nowhere to write to, this handler is a no-op
+	if a.db == nil {
+		return nil
+	}
+
+	streamNames := strings.Split(payload.StreamNames, ",")
```
iameli, line 29: Probably this kind of string-parsing stuff should be handled in the trigger-parsing code, so then the `UserEndPayload` can just contain clean slices of data

iameli (approved):
LGTM, left one comment that we should probably do the comma string parsing prior to the trigger-handling code but this is entirely shippable

handlers/analytics/user_end.go
thomshutt: Yeah great call - in the initial implementation I hadn't realised we weren't just passing everything through as strings to the DB. I'll update now.

PR #829, opened 2023-08-29
[SRE-160] CDN redirect only for /hls

pwilczynskiclearcode:
This reverts commit b5a530d514d23c06b5ad64abfc5a5f29dd637fd0.

### DIFF
--- a/balancer/balancer.go
+++ b/balancer/balancer.go
@@ -335,10 +335,12 @@ func (b *BalancerImpl) GetBestNode(ctx context.Context, redirectPrefixes []strin
 	return fallbackNode(fallbackAddr, fallbackPrefix, playbackID, redirectPrefixes[0], err)
 }
 
+// `playbackID`s matching the pattern `....-....-.....-....`
+var regexpStreamKey = regexp.MustCompile(`^(?:\w{4}-){3}\w{4}$`)
+
 func fallbackNode(fallbackAddr, fallbackPrefix, playbackID, defaultPrefix string, err error) (string, string, error) {
 	// Check for `playbackID`s matching the pattern `....-....-.....-....`
-	r := regexp.MustCompile(`^(?:\w{4}-){3}\w{4}$`)
-	if r.MatchString(playbackID) {
+	if regexpStreamKey.MatchString(playbackID) {
 		return fallbackAddr, playbackID, nil
 	}
 
--- a/handlers/geolocation/geolocation.go
+++ b/handlers/geolocation/geolocation.go
@@ -14,6 +14,7 @@ import (
 	"github.com/livepeer/catalyst-api/cluster"
 	"github.com/livepeer/catalyst-api/config"
 	"github.com/livepeer/catalyst-api/handlers/misttriggers"
+	"github.com/livepeer/catalyst-api/metrics"
 )
 
 type GeolocationHandlersCollection struct {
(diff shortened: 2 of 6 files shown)


handlers/geolocation/geolocation.go
```diff
@@ -117,14 +140,15 @@ func parsePlus(plusString string) (string, string) {
 	return prefix, playbackID
 }
 
+var regexpHLSPath = regexp.MustCompile(`^/hls/([\w+-]+)/(.*index.m3u8.*)$`)
```
pwilczynskiclearcode, line 143: Compile regexp only once, during initialization.

https://pkg.go.dev/regexp#MustCompile
> MustCompile is like Compile but panics if the expression cannot be parsed. It simplifies safe initialization of global variables holding compiled regular expressions.

handlers/geolocation/geolocation.go
```diff
@@ -117,60 +141,63 @@ func parsePlus(plusString string) (string, string) {
...
```
pwilczynskiclearcode, line 144: Compile regexp only once, during initialization.

https://pkg.go.dev/regexp#MustCompile
> MustCompile is like Compile but panics if the expression cannot be parsed. It simplifies safe initialization of global variables holding compiled regular expressions.

handlers/geolocation/geolocation.go
```diff
@@ -24,38 +25,61 @@ type GeolocationHandlersCollection struct {
...
+		host := r.Host
+		pathType, prefix, playbackID, pathTmpl := parsePlaybackID(r.URL.Path)
+
+		if c.Config.CdnRedirectPrefix != nil && (pathType == "hls" || pathType == "webrtc") {
+			doCdnRedirect := contains(playbackID, c.Config.CdnRedirectPlaybackIDs)
+			if doCdnRedirect {
+				if pathType == "webrtc" {
+					w.WriteHeader(http.StatusNotAcceptable) // 406 error
+					metrics.Metrics.CDNRedirectWebRTC406.WithLabelValues(playbackID).Inc()
+					glog.V(6).Infof("%s not supported for CDN-redirected %s", r.URL.Path, playbackID)
```
pwilczynskiclearcode, line 42: After deploying on staging it will be tested in the player to make sure that 406 falls back into `/hls/` endpoint.

balancer/balancer.go
```diff
@@ -335,10 +335,12 @@ func (b *BalancerImpl) GetBestNode(ctx context.Context, redirectPrefixes []strin
 	return fallbackNode(fallbackAddr, fallbackPrefix, playbackID, redirectPrefixes[0], err)
 }
 
+// `playbackID`s matching the pattern `....-....-.....-....`
+var regexpPlaybackID = regexp.MustCompile(`^(?:\w{4}-){3}\w{4}$`)
```
leszko, line 339: Why do we need playback id to match this patter? 🤔

main.go
```diff
@@ -87,6 +87,8 @@ func main() {
 	fs.StringVar(&cli.NodeName, "node", hostname, "Name of this node within the cluster")
 	config.SpaceSliceFlag(fs, &cli.BalancerArgs, "balancer-args", []string{}, "arguments passed to MistUtilLoad")
 	fs.StringVar(&cli.NodeHost, "node-host", "", "Hostname this node should handle requests for. Requests on any other domain will trigger a redirect. Useful as a 404 handler to send users to another node.")
+	config.CommaSliceFlag(fs, &cli.CdnRedirectPlaybackIDs, "cdn-redirect-playback-ids", []string{}, "PlaybackIDs to be redirected to CDN.")
```
pwilczynskiclearcode, line 90: We might need to accept sth like `-cdn-redirect-playback-ids=all` to be easily press big red button for entire traffic

balancer/balancer.go
iameli: It's not actually a playback id if it's in this format! It's a stream key, and we match this to send them directly to a Mist endpoint instead of doing playback load balancing because it means they're streaming in 
```suggestion
var regexpStreamKey = regexp.MustCompile(`^(?:\w{4}-){3}\w{4}$`)
```

balancer/balancer.go
```diff
@@ -335,10 +335,12 @@ func (b *BalancerImpl) GetBestNode(ctx context.Context, redirectPrefixes []strin
...
+
 func fallbackNode(fallbackAddr, fallbackPrefix, playbackID, defaultPrefix string, err error) (string, string, error) {
 	// Check for `playbackID`s matching the pattern `....-....-.....-....`
-	r := regexp.MustCompile(`^(?:\w{4}-){3}\w{4}$`)
-	if r.MatchString(playbackID) {
+	if regexpPlaybackID.MatchString(playbackID) {
```
iameli, line 343:
```suggestion
	if regexpStreamKey.MatchString(playbackID) {
```

handlers/geolocation/geolocation.go
```diff
@@ -24,38 +25,61 @@ type GeolocationHandlersCollection struct {
...
-					glog.Errorf("failed to parse incoming url for redirect url=%s err=%s", r.URL.String(), err)
-					w.WriteHeader(http.StatusInternalServerError)
+	return func(w http.ResponseWriter, r *http.Request, params httprouter.Params) {
+		host := r.Host
+		pathType, prefix, playbackID, pathTmpl := parsePlaybackID(r.URL.Path)
+
+		if c.Config.CdnRedirectPrefix != nil && (pathType == "hls" || pathType == "webrtc") {
+			doCdnRedirect := contains(playbackID, c.Config.CdnRedirectPlaybackIDs)
+			if doCdnRedirect {
+				if pathType == "webrtc" {
```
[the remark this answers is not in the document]
leszko: Ok, makes sense. Please document it, because it's not obvious 🙃

leszko (approved):
Looks good from my end, but please address the comments and wait for the final review from @iameli before merging.

balancer/balancer.go
pwilczynskiclearcode: That's still confusing because be pass `playbackID` into `fallbackNode()` function from `RedirectHandler()`... so in part of the code it's playbackID and now it would be considered stream-key?

handlers/geolocation/geolocation.go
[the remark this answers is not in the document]
pwilczynskiclearcode: Documented in the code and on the diagram https://www.notion.so/livepeer/CDN-Big-Red-Button-8e88b7f422994176a3cc8ba9773ceb16

main.go
iameli: Maybe `*` instead of `all` just so it can't collide?

iameli (approved):
LGTM!

PR #830, opened 2023-08-29
Bring in-flight HTTP requests into capacity calculation

thomshutt:
This is to avoid the situation when we get flooded with jobs in quick succession and hit a race condition where they haven't got far enough into processing for the "jobs in flight" number to be updated yet.


mjh1:
Ah the unit tests have caught an issue, because i'm trying to create the job earlier it doesn't work when we're doing the background external strategy as we're using the same job object for two concurrent jobs.

thomshutt:
@mjh1 let's just drop that strategy, I don't think we'll need it again

mjh1:
@thomshutt look ok to you? I definitely need to do a bit more testing but it's looking good so far

thomshutt:
@mjh1 yeah, lgtm!

mjh1 (approved):

PR #832, opened 2023-08-29
Improve Transcode Quality

leszko:
**Phase 1** as described in the [related Design Doc](https://www.notion.so/livepeer/Transcode-Quality-investigation-improvements-11314327aa484fb6927a48b102bd54bf?pvs=4#b021e1557f0b49a3b88c22ae6e077e92).

Related to https://github.com/livepeer/lpms/pull/371 and https://github.com/livepeer/go-livepeer/pull/2848

Before most Os update the go-livepeer version to support the CRF parameter, I suggest to do the following in order to improve slightly the transcode quality:
- CRF=27
- Bitrate=1.2*Current Bitrate

Rationale from Design Doc:
> In Phase 1, we won’t really use the CRF parameter, so I’d set it to some conservative value that won’t make the transcoding quality worse (27 sounds like a good value) and still use bitrate to limit the max video bitrate. At the same time, I’d slightly increase the bitrate parameter, because we currently set it too low. For the highest resolution, we set it to the average bitrate of the source video, which results in a worse quality than the source, because we effectively set: max bitrate = source average bitrate. After this change, the quality will be slightly better and the segment size will be slightly bigger.


video/profiles.go
```diff
@@ -12,6 +12,7 @@ const (
 	MaxVideoBitrate         = 288_000_000
 	TrackTypeVideo          = "video"
 	TrackTypeAudio          = "audio"
+	defaultCRF              = 27
```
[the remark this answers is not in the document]
leszko: Just from some tests. Please check [this part of design doc](https://www.notion.so/livepeer/Transcode-Quality-investigation-improvements-11314327aa484fb6927a48b102bd54bf?pvs=4#b021e1557f0b49a3b88c22ae6e077e92) for details. 23 will give way bigger segment size.

leszko:
@mjh1 After my change, all the transcode profiles JSON tests are failing. Do you have any way to re-generate them or something? How did they got created?

video/profiles.go
```diff
@@ -134,7 +142,10 @@ func GetDefaultPlaybackProfiles(video InputTrack) ([]EncodedProfile, error) {
 		// check it here.
 		lowerQualityThanSrc := profile.Height < video.Height && profile.Bitrate < video.Bitrate
 		if lowerQualityThanSrc {
-			relativeBitrate := float64(profile.Width*profile.Height) * (float64(videoBitrate) / float64(video.Width*video.Height))
+			// relativeBitrate needs to be slightly higher than the proportional average bitrate of the source video.
+			// Livepeer network uses bitrate to set max bitrate for encoding, so for the video to look good, we multiply
+			// it by a factor of 1.2.
+			relativeBitrate := 1.2 * float64(profile.Width*profile.Height) * (float64(videoBitrate) / float64(video.Width*video.Height))
```
emranemran, line 148: how did we determine the 1.2 multiplier?

video/profiles.go
```diff
@@ -103,22 +106,27 @@ func SetTranscodeProfiles(inputVideoStats InputVideo, reqTranscodeProfiles []Enc
...
 		}
 	}
 	return reqTranscodeProfiles, nil
 }
 
-func GenerateSingleProfileWithTargetBitrate(videoTrack InputTrack, videoBitrate int64) []EncodedProfile {
+func GenerateSingleProfileWithTargetParams(videoTrack InputTrack, videoProfile EncodedProfile) []EncodedProfile {
 	profiles := make([]EncodedProfile, 0)
+	var CRF uint = defaultCRF
+	if videoProfile.CRF != 0 {
```
emranemran, line 119: check should be `> 0` since we can't have negative CRF values?

video/profiles.go
```diff
@@ -103,22 +106,27 @@ func SetTranscodeProfiles(inputVideoStats InputVideo, reqTranscodeProfiles []Enc
...
+		CRF = videoProfile.CRF
```
emranemran, line 120: CRF = 0 is still a valid value (meaning lossless). Are we explicitly saying crf = 0 should never be allowed or should this check for values between 0 and 51?

video/profiles.go
leszko: I didn't really. I just know that what we currently have is too low, because we're setting: `max bitrate = source avg bitrate` which gives poor results.

As described in the doc, I plan to not set it at all, when the Os start to support the `quality` param (Phase 2).

video/profiles.go
leszko: This is a good point. The problem with `0` is that everywhere in go and protobuf, it's the default => so, the assumption is that `0` is a not set value. 

Anyway, since as you know, there is no direct correspondence of the the `quality` value and the `crf` / `cq`, let's maybe not think about this corner case too much.

leszko:
@emranemran I also renamed the `crf` to `quality` everywhere as we discussed. PTAL.

video/profiles.go
```diff
@@ -134,7 +142,10 @@ func GetDefaultPlaybackProfiles(video InputTrack) ([]EncodedProfile, error) {
 		// check it here.
 		lowerQualityThanSrc := profile.Height < video.Height && profile.Bitrate < video.Bitrate
 		if lowerQualityThanSrc {
```
thomshutt, line 144: Don't we also want this for the rendition that's the same resolution as the source?

leszko: Ohh wait, maybe you're right that we should also increase the bitrate in line 157. Will update it there.

leszko: Updated in 13a6bf2ed7c79f1f2e5781d5f46123dd4c16dced

thomshutt (approved):

emranemran (approved):

PR #835, opened 2023-08-30
Implement Dash and HLS Fmp4

mjh1:
Implementing Dash and HLS Fmp4 with a single ffmpeg command. I've moved the fmp4 ffmpeg call to after the rendition loop in `transcode.go` so that we can pass all of the concatenated ts files to ffmpeg as multiple inputs. FFmpeg then generates all of the required manifests and the separate rendition media files:

Screenshot 2023-08-30 at 10 37 47

I've started working on a cucumber test for this but will do it in a separate PR as I had to refactor some of the cucumber code a bit.

I also wasn't able to use the `github.com/u2takey/ffmpeg-go` library, I couldn't work out how to pass in multiple inputs. So I'm just doing basic `os/exec` instead.


video/transmux.go
```diff
@@ -20,75 +24,86 @@ func MuxTStoMP4(tsInputFile, mp4OutputFile string) ([]string, error) {
...
+		"-bsf:a", "aac_adtstoasc",
+		"-f", "dash",
+		"-dash_segment_type", "mp4",
+		"-single_file", "1",
+		"-hls_playlist", "1",
+		"-hls_time", "10",
+		"-hls_playlist_type", "vod",
+		"-hls_segment_type", "fmp4",
+		"-vtag", "avc1",
+		"-atag", "mp4a",
```
mjh1, line 64: Would like to figure out how to avoid needing these two params. Ffmpeg errors without them (when using ts input rather than mp4)

transcode/transcode.go
```diff
@@ -59,10 +61,6 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
 	if err != nil {
 		return outputs, segmentsCount, err
 	}
-	fmp4TargetURL, err := getFmp4TargetURL(transcodeRequest)
```
emranemran, line 62: I think this should still be here to validate that the url was set correctly and then in line 182, use `fmp4TargetURL` instead of `transcodeRequest.FragMp4TargetUrl`.

emranemran (commented):
Left a few comments but overall the new ffmpeg command LGTM.

video/transmux.go
```diff
@@ -20,75 +24,86 @@ func MuxTStoMP4(tsInputFile, mp4OutputFile string) ([]string, error) {
...
+	)
+	args = append(args, mapArgs...)
+	args = append(args, fmp4ManifestOutputFile)
 
-	fmp4VideoOutputFile := fmp4ManifestOutputFile[:len(fmp4ManifestOutputFile)-len(filepath.Ext(fmp4ManifestOutputFile))] + ".m4s"
+	timeout, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
```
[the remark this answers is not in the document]
emranemran: OK just saw your comment on the description and that makes sense - i think I remember running into a similar issue as well

thomshutt (approved):
Agree with Emran's comment about using the ffmpeg library but otherwise LGTM!

transcode/transcode.go
mjh1: It should be ok I think, we're calling `getMp4OutputType` instead which does the same stuff as `getFmp4TargetURL` did, it checks if the string is empty and if not parses it to a url

emranemran (approved):

PR #840, opened 2023-08-31
Fix running cucumber tests individually

mjh1:
I found if I tried to run one of the vod tests on their own the test would fail because the global variable sourceOutputDir was not set at the time InitializeScenario ran, so would be empty on the step context. So I moved the startup code to the step context to ensure SourceOutputDir gets set properly on the step context.


thomshutt (approved):
Good catch!

PR #851, opened 2023-09-07
[SRE-160] CDN Redirect - add missing prefix

pwilczynskiclearcode:


handlers/geolocation/geolocation.go
```diff
@@ -46,7 +46,13 @@ func (c *GeolocationHandlersCollection) RedirectHandler() httprouter.Handle {
 				newURL, _ := url.Parse(r.URL.String())
 				newURL.Scheme = protocol(r)
 				newURL.Host = c.Config.CdnRedirectPrefix.Host
-				newURL.Path, _ = url.JoinPath(c.Config.CdnRedirectPrefix.Path, newURL.Path)
+				fullPlaybackID := playbackID
+				if prefix == "" {
+					fullPlaybackID = fmt.Sprintf("video+%s", fullPlaybackID)
```
leszko, line 51: What is the case when the `prefix` is an empty string? And why we use `video+` (and not `videorec+`) in that case?

pwilczynskiclearcode: Initially I thought we can avoid `GetBestNode()` node... but querying mist seems to be the only way to extract the correct prefix

iameli (approved):
Cool!

leszko (approved):
LGTM 👍

PR #854, opened 2023-09-08
Pass job info by value to fix metrics for fallback jobs

mjh1:
Since we're re-using the job object now we need to pass by value here to avoid picking up values from the fallback job


thomshutt (approved):

thomshutt:
LGTM but linting isn't happy

thomshutt (approved):

PR #857, opened 2023-09-12
clip: add scaffolding to accept clipping requests

emranemran:
This adds additional parameters to the vod endpoint to accept clipping start/end times. If the fields are set with valid values, the input manifest will be clipped and then a new clipped manifest will be used in the VOD pipeline to generate assets as usual. The logic for clipping the actual manifest will be done as a follow-on PR.

### DIFF
--- a/handlers/upload.go
+++ b/handlers/upload.go
@@ -49,6 +49,9 @@ type UploadVODRequest struct {
 	TargetSegmentSizeSecs int64                  `json:"target_segment_size_secs"`
 	Profiles              []video.EncodedProfile `json:"profiles"`
 	PipelineStrategy      pipeline.Strategy      `json:"pipeline_strategy"`
+
+	// Forwarded to clipping stage:
+	ClipStrategy video.ClipStrategy `json:"clip_strategy"`
 }
 
 type UploadVODResponse struct {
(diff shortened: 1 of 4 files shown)


handlers/upload.go
```diff
@@ -101,6 +104,21 @@ func (r UploadVODRequest) IsProfileValid() bool {
...
+
+	if startTime < 0 || endTime <= 0 || startTime == endTime {
+		return false
+	}
+
+	epoch := time.Date(1970, 1, 1, 0, 0, 0, 0, time.UTC)
+	start := epoch.Add(time.Duration(startTime) * time.Second)
+	end := epoch.Add(time.Duration(endTime) * time.Second)
+
+	return !start.After(end)
```
thomshutt, line 119: Could we just do `startTime < endTime` here?

video/clip.go
```diff
@@ -55,6 +62,13 @@ func getRelevantSegment(allSegments []*m3u8.MediaSegment, playHeadTime float64,
 	return 0, fmt.Errorf("error clipping: did not find a segment that falls within %v seconds", playHeadTime)
 }
 
+// Function that will take a source URL manifest and return a new URL
+// pointing to the clipped manifest
+func ClipInput(requestID string, srcUrl *url.URL, startTime, endTime float64) (*url.URL, error) {
```
mjh1, line 67: Is the plan for this function to call ClipManifest after parsing `srcUrl` into a manifest?

emranemran: Yep, using a separate PR and trying to keep PRs small for easier review.

handlers/upload.go
emranemran: done

mjh1 (approved):

PR #870, opened 2023-09-18
clip: add support to clip and upload clipped segments

emranemran:
Add manifest handling logic to download a source manifest, call clipping methods on it, and then upload the returned segments to OS.

Note 1: The clipped manifest generation and upload step will be done in a follow-on PR.

Note 2: cucumber tests will be added in a future PR

### DIFF
--- a/video/clip.go
+++ b/video/clip.go
@@ -8,10 +8,6 @@ import (
 	"time"
 )
 
-const (
-	ClipStorageDir = "/tmp/clip_stage"
-)
-
 type ClipStrategy struct {
 	Enabled    bool
 	StartTime  float64 `json:"start_time,omitempty"`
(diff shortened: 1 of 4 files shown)


video/clip.go
```diff
@@ -5,10 +5,13 @@ import (
 	"github.com/grafov/m3u8"
 	"github.com/livepeer/catalyst-api/log"
 	ffmpeg "github.com/u2takey/ffmpeg-go"
-	"net/url"
 	"time"
 )
 
+const (
+	ClipStorageDir = "/tmp/clip_stage"
```
mjh1, line 12: I think we should use golang's built in tmp functions e.g. [os.MkdirTemp()](https://pkg.go.dev/os#MkdirTemp)

https://github.com/livepeer/catalyst-api/blob/506e24fe359d53362d154d37c72541d7876baff7/pipeline/ffmpeg.go#L279

emranemran: Sure, done.

mjh1 (approved):

PR #872, opened 2023-09-18
handlers: add playback-id to clipping request

emranemran:
This playback-id will be returned in the clipping response back to studio so that clipped assets can be tracked to the original asset's playback id.

### DIFF
--- a/handlers/schemas/UploadVOD.yaml
+++ b/handlers/schemas/UploadVOD.yaml
@@ -31,6 +31,8 @@ properties:
         type: "integer"
       end_time:
         type: "integer"
+      playback_id:
+        type: "string"
     required:
       -  "start_time"
     additionalProperties: false
--- a/pipeline/coordinator.go
+++ b/pipeline/coordinator.go
@@ -276,7 +276,7 @@ func (c *Coordinator) StartUploadJob(p UploadJobPayload) {
 		if clients.IsHLSInput(sourceURL) {
 			// Currently we only clip an HLS source (e.g recordings or transcoded asset)
 			if p.ClipStrategy.Enabled {
-				log.Log(p.RequestID, "clippity clipping the input")
+				log.Log(p.RequestID, "clippity clipping the input", "Playback-ID", p.ClipStrategy.PlaybackID)
 				// Use new clipped manifest as the source URL
 				sourceURL, err = video.ClipInput(p.RequestID, sourceURL, p.ClipStrategy.StartTime, p.ClipStrategy.EndTime)
 				if err != nil {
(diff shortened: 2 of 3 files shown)


video/clip.go
```diff
@@ -10,9 +10,10 @@ import (
 )
 
 type ClipStrategy struct {
-	Enabled   bool
-	StartTime float64 `json:"start_time,omitempty"`
-	EndTime   float64 `json:"end_time,omitempty"`
+	Enabled    bool
+	StartTime  float64 `json:"start_time,omitempty"`
+	EndTime    float64 `json:"end_time,omitempty"`
+	PlaybackId string  `json:"playback_id,omitempty"` // playback-id of asset to clip
```
thomshutt, line 16: Petty, but `ID` should be capitalised in the variable name to follow Golang guidelines 

```suggestion
	PlaybackID string  `json:"playback_id,omitempty"` // playback-id of asset to clip
```

pipeline/coordinator.go
```diff
@@ -276,7 +276,7 @@ func (c *Coordinator) StartUploadJob(p UploadJobPayload) {
 		if clients.IsHLSInput(sourceURL) {
 			// Currently we only clip an HLS source (e.g recordings or transcoded asset)
 			if p.ClipStrategy.Enabled {
-				log.Log(p.RequestID, "clippity clipping the input")
+				log.Log(p.RequestID, "clippity clipping the input", "Playback-ID", p.ClipStrategy.PlaybackId)
```
thomshutt, line 279:
```suggestion
				log.Log(p.RequestID, "clippity clipping the input", "Playback-ID", p.ClipStrategy.PlaybackID)
```

handlers/schemas/UploadVOD.yaml
```diff
@@ -31,8 +31,11 @@ properties:
         type: "integer"
       end_time:
         type: "integer"
+      playback_id:
+        type: "string"
     required:
       -  "start_time"
+      -  "playback_id"
```
thomshutt, line 38: I don't think we should make this required (if we're aiming for non-Studio Catalyst users in the theoretical future)

mjh1 (commented):
lgtm after thom's comments

handlers/schemas/UploadVOD.yaml
emranemran: Done

pipeline/coordinator.go
emranemran: Fixed

video/clip.go
emranemran: Fixed

mjh1 (approved):

PR #877, opened 2023-09-22
clip: switch start/end time in request to unix timestamps in milliseconds

emranemran:
The player is able to send absolute unix timestamps when setting the start/end times of the clips. So we switch to accepting unix timestamps in milliseconds and then convert it to offsets in seconds.

TODO: need to add a test.

### DIFF
--- a/video/clip.go
+++ b/video/clip.go
@@ -10,9 +10,9 @@ import (
 
 type ClipStrategy struct {
 	Enabled    bool
-	StartTime  float64 `json:"start_time,omitempty"`
-	EndTime    float64 `json:"end_time,omitempty"`
-	PlaybackID string  `json:"playback_id,omitempty"` // playback-id of asset to clip
+	StartTime  int64  `json:"start_time,omitempty"`
+	EndTime    int64  `json:"end_time,omitempty"`
+	PlaybackID string `json:"playback_id,omitempty"` // playback-id of asset to clip
 }
 
 // format time in secs to be copatible with ffmpeg's expected time syntax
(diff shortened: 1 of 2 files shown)


video/clip.go
```diff
@@ -62,6 +62,16 @@ func getRelevantSegment(allSegments []*m3u8.MediaSegment, playHeadTime float64,
 	return 0, fmt.Errorf("error clipping: did not find a segment that falls within %v seconds", playHeadTime)
 }
 
+func ConvertUnixMillisToSeconds(firstSegment *m3u8.MediaSegment, startTimeUnixMillis, endTimeUnixMillis int64) (float64, float64) {
+	firstSegProgramDateTimeUTC := firstSegment.ProgramDateTime
```
mjh1, line 66: Is ProgramDateTime guaranteed to be set at this point or should we error or something if not present?

emranemran: Good point - if the manifest doesn't have PROGRAM-DATE-TIME tags then it should fail out here because we won't be able to calculate offsets. Will update.

emranemran: Added check for empty timestamp.

video/clip.go
```diff
@@ -10,9 +10,9 @@ import (
 
 type ClipStrategy struct {
 	Enabled    bool
-	StartTime  float64 `json:"start_time,omitempty"`
-	EndTime    float64 `json:"end_time,omitempty"`
-	PlaybackID string  `json:"playback_id,omitempty"` // playback-id of asset to clip
+	StartTime  int64  `json:"start_time,omitempty"`
```
gioelecerati, line 13: We need to make this change as well in go-api-client too, in the ClipTaskParams definition - will do

gioelecerati (approved):
Looking good to me

video/clip.go
```diff
@@ -62,6 +62,25 @@ func getRelevantSegment(allSegments []*m3u8.MediaSegment, playHeadTime float64,
...
+	firstSegProgramDateTimeUTC := firstSegment.ProgramDateTime
+	if firstSegProgramDateTimeUTC.IsZero() {
+		return 0.0, 0.0, fmt.Errorf("error clipping: PROGRAM-DATE-TIME of first segment is not set")
+	}
+	firstSegUnixMillis := firstSegProgramDateTimeUTC.UnixNano() / int64(time.Millisecond)
+
+	startTimeSeconds := float64(startTimeUnixMillis-firstSegUnixMillis) / 1000.0
+	endTimeSeconds := float64(endTimeUnixMillis-firstSegUnixMillis) / 1000.0
+
+	log.Log(requestID, "Clipping timestamps",
```
emranemran, line 75: @gioelecerati FYI: I couldn't reproduce the issue we saw with timestamps being affected by timezone. So hopefully this little log lets us easily spot an weirdness after we test on staging.

PR #879, opened 2023-09-24
clip: format start/end times passed to ffmpeg to use milliseconds

emranemran:
The -ss and -to parameters used by ffmpeg was previously only converting to HH:MM:SS and the milliseconds portion was being dropped. Update the time formatting logic so that a format of HH:MM:SS.m... is used. This allows millisecond accuracy for clipping.

### DIFF
--- a/video/clip.go
+++ b/video/clip.go
@@ -16,10 +16,11 @@ type ClipStrategy struct {
 }
 
 // format time in secs to be copatible with ffmpeg's expected time syntax
-func formatTime(seconds float64) string {
-	duration := time.Duration(seconds * float64(time.Second))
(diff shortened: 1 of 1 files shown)


mjh1 (approved):

video/clip.go
```diff
@@ -16,10 +16,11 @@ type ClipStrategy struct {
 }
 
 // format time in secs to be copatible with ffmpeg's expected time syntax
-func formatTime(seconds float64) string {
```
mjh1, line 19: Is this epoch/unix time?

emranemran: It's actually the time in seconds from the *beginning* of the recording stream i.e. delta from t=0 (first segment). I'm doing a follow-on change related in this file and i'll add a comment here to make that clear.

PR #881, opened 2023-09-25
UploadVOD yaml: do not set start_time as required

emranemran:
The clipping request is already validated in catalyst-api so we should not be setting start_time in the vod request as this will prevent recordings/vod from working.


gioelecerati (approved):
LGTM

PR #884, opened 2023-09-26
manifest: set the clipped manifest URL correctly for storj

emranemran:
This is a bit of a workaround (hack) to correctly set storj URLs for the clipped manifest file. A follow-on PR will be used to set the output relative path in the request directly which will will simplify this logic.

### DIFF
--- a/clients/manifest.go
+++ b/clients/manifest.go
@@ -336,23 +336,18 @@ func ClipInputManifest(requestID, sourceURL, clipTargetUrl string, startTimeUnix
 	// to the clipped manifest file using the public source url. This logic should be
 	// simplified by setting an output folder explicitly as a param in the ClippingStrategy
 	// struct as part of the clipping request
-	clipTargetOSURL, err := url.Parse(clipTargetUrl)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


mjh1:
Noticed this issue when I came online this morning and tested, had the same idea to fix in this way ha 👍

mjh1 (approved):

clients/manifest.go
```diff
@@ -336,23 +336,18 @@ func ClipInputManifest(requestID, sourceURL, clipTargetUrl string, startTimeUnix
...
 	if err != nil {
-		return nil, fmt.Errorf("error clipping: failed to parse clipped playlist: %s", err)
+		return nil, fmt.Errorf("error clipping: failed to parse sourceURL: %s", err)
 	}
 	// set the correct path to clip.m3u8 file in the base url that will be used as the
 	// input file to next VOD (transcode) stage.
-	clipTargetPublicURL.Path += clipPlaybackRelPath + "/" + ClipManifestFilename
 
-	return clipTargetPublicURL, nil
+	return source.JoinPath("..", clipPlaybackRelPath, ClipManifestFilename), nil
```
emranemran, line 350: nice

PR #890, opened 2023-09-28
clip: fix segment clipping and manifest generation bugs

emranemran:


clients/manifest.go
```diff
@@ -364,6 +378,18 @@ func CreateClippedPlaylist(origManifest m3u8.MediaPlaylist, segs []*m3u8.MediaSe
...
+		// Remove PROGRAM-DATE-TIME tag from first clipped segment so that player doesn't
+		// run into seek issues or display incorrect times on playhead
+		if i == 0 {
+			s.ProgramDateTime = t
+		}
+		// Add a DISCONTINUITY tag to let hls players know about different encoding between
+		// segments. But don't do this if there's a single segment in the clipped manifest
+		// or contains total two segments in the clipped manifest (i.e. back to back segments
+		// were run through ffmpeg)
+		if i-1 == 0 && totalSegs > 2 {
```
thomshutt, line 390: Do we also need this before the final segment?

video/clip.go
```diff
@@ -89,67 +97,109 @@ func ConvertUnixMillisToSeconds(requestID string, firstSegment *m3u8.MediaSegmen
...
+//	"c:a": "aac": re-encode audio and clip.
 func ClipSegment(tsInputFile, tsOutputFile string, startTime, endTime float64) error {
-	start := formatTime(startTime)
-	end := formatTime(endTime)
-	err := ffmpeg.Input(tsInputFile).
-		Output(tsOutputFile, ffmpeg.KwArgs{"ss": start, "to": end,
+	var start, end string
+	var args ffmpeg.KwArgs
+	if endTime < 0 {
+		// Clip from specified start time to the end of
```
thomshutt, line 167: ❤️  love these comments

video/clip.go
```diff
@@ -89,67 +97,109 @@ func ConvertUnixMillisToSeconds(requestID string, firstSegment *m3u8.MediaSegmen
...
+// Currently using a combination of these settings for the encode step:
 //
 //	"c:v": "libx264": Specifies H.264 video codec.
-//	"r": "24": Sets video frame rate to 24 FPS.
 //	"g": "48": Inserts keyframe every 48 frames (GOP size).
 //	"keyint_min": "48": Minimum keyframe interval.
 //	"sc_threshold": 50: Detects scene changes with threshold 50.
 //	"bf": "0": Disables B-frames for bidirectional prediction.
-//	"c:a": "copy": Copies audio stream without re-encoding.
+//	"c:a": "aac": re-encode audio and clip.
```
thomshutt, line 162: Why did this change?

video/clip.go
```diff
@@ -89,67 +97,109 @@ func ConvertUnixMillisToSeconds(requestID string, firstSegment *m3u8.MediaSegmen
...
+			"c:a":          "aac",
 			"c:v":          "libx264",
-			"r":            "24",
 			"g":            "48",
 			"keyint_min":   "48",
 			"sc_threshold": 50,
-			"bf":           "0",
-			"c:a":          "copy"}).
+			"ss":           start}
+	} else if startTime < 0 {
```
mjh1, line 177: Is it definitely `< 0` and not `<=0` just double checking, is it set to -1 or something somewhere?

clients/manifest.go
emranemran: No, because the final segment is not re-encoded to shares the same params as the segment before it -- so a/v decoders don't need to be re-initialized.

video/clip.go
emranemran: This was a mistake in copy/paste from local cmd line prototype to code. The audio track needs to be re-encoded so that it gets clipped at the same places as video track.

video/clip.go
emranemran: Yup, see this: 
https://github.com/livepeer/catalyst-api/pull/890/files#diff-47d60fc2d61cc45a41ecad1e68397953b10fe4d7d8dcd3e404b5568263fbd8c2R298

I'm using it as an override to the ClipSegment function to clip with different ffmpeg calls. I don't *love* how I did this so will refactor and clean it up in a follow-up. I think having separate functions with different ffmpeg calls would be more readable here.

mjh1 (approved):

thomshutt (approved):

PR #894, opened 2023-09-29
Transcode all H264 audio to AAC, to allow us to process it via the Livepeer pipeline

thomshutt:


mjh1 (approved):
What was the use case we needed this for btw?

PR #897, opened 2023-10-03
clipping fixes

emranemran:
**coordinator: add retries to clipping**
* In several cases, we may attempt clipping segments that have been
partially written to storage. In this case, ffmpeg/ffprobe operations
will fail and we should retry a bit before giving up.

**clip: add map arguments to ffmpeg clip commands**
* The clip segments that are re-encoded will end up with the video track
  first and audio track second wheras the track ordering is reversed for
the source recorded segments. If these segments are transmuxed into a
single ts/mp4 file, the resulting file will stop playing after the first
segment's worth of video. To resolve this, this commit adds the 'map'
args to the ffmpeg command to ensure the audio track is always placed
first and video track second similar to the recorded segments.

* The ffmpeg call is also being switched to a cmd.Run call since the
  go-ffmpeg library doesn't allow us to have multiple '-map' args in the
command line (one instance gets dropped resulting in an incorrect
ffmpeg cmd).

### DIFF
--- a/pipeline/coordinator.go
+++ b/pipeline/coordinator.go
@@ -4,6 +4,7 @@ import (
 	"crypto/rsa"
 	"database/sql"
 	"fmt"
+	"github.com/cenkalti/backoff/v4"
 	"math"
 	"net/url"
 	"os"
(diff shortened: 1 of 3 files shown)


pipeline/coordinator.go
```diff
@@ -145,6 +146,10 @@ func (j *JobInfo) ReportProgress(stage clients.TranscodeStatus, completionRatio
 	_ = j.statusClient.SendTranscodeStatus(tsm)
 }
 
+func ClippingRetryBackoff() backoff.BackOff {
```
thomshutt, line 149: This ties us up for quite a while in terms of job capacity. Is there a more intelligent method we can move to in the future? Maybe having Task Runner handle the retries?

emranemran: True, we can do that. Let me file a ticket and we'll follow up quickly after launch to address this.

mjh1 (approved):
nice tidy up of the ffmpeg args if statement 👍

PR #902, opened 2023-10-06
transcode: add logic to force a session re-init

emranemran:
When a sequence of segments has a discontinuity (e.g. encoded differently), the transcoder at T must be re-initialized so that transcode operations can be completed correctly. The T takes some liberties and assumes a sequence of segments are encoded the same way and thus, ends up not re-initializing certain parts of the pipeline (e.g. demuxer, etc) to save time and improve efficiency. This can result in corrupted segments with two video tracks instead of one audio and one video track.

For clipping, the first segment (index=0) is encoded differently from the rest of the file. As such, a session reinit must be forced at the T using a header set in the request to B.

### DIFF
--- a/transcode/transcode.go
+++ b/transcode/transcode.go
@@ -428,6 +428,12 @@ func transcodeSegment(
 	} else {
 		transcodeConf.ForceSessionReinit = false
 	}
+	// This is a temporary workaround that implements the same logic
+	// as the previous if block -- a new manifestID will force a
+	// T session re-init between segment at index=0 and index=1.
+	if int64(segment.Index) == 0 {
+		manifestID = manifestID + "_clip"
+	}
 
 	var tr clients.TranscodeResult
 	err := backoff.Retry(func() error {
--- a/transcode/transcode.go
+++ b/transcode/transcode.go
@@ -431,7 +431,7 @@ func transcodeSegment(
 	// This is a temporary workaround that implements the same logic
 	// as the previous if block -- a new manifestID will force a
 	// T session re-init between segment at index=0 and index=1.
-	if int64(segment.Index) == 0 {
+	if transcodeRequest.IsClip && int64(segment.Index) == 0 {
 		manifestID = manifestID + "_clip"
 	}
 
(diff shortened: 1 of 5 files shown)


transcode/transcode.go
```diff
@@ -411,6 +412,23 @@ func transcodeSegment(
...
+	// This is required because the segment at idx=0 is a locally
+	// re-encoded segment and the following segment at idx=1 is a
+	// source recorded segment. Without a re-init of the transcoder,
+	// the different encoding between the two segments causes the
+	// transcode operation to incorrectly tag the output segment as
+	// having two video tracks.
+	if transcodeRequest.IsClip && int64(segment.Index) == 1 {
+		transcodeConf.ForceSessionReinit = true
+	} else {
+		transcodeConf.ForceSessionReinit = false
```
leszko, line 429: Nit: I think you can skip this `else` since `false` is the default value of bool.

leszko (approved):
It looks good. However, before the new go-livepeer version adoption we may need to have a hacky workaround and change `manifestID` for the clipped segments.

transcode/transcode.go
mjh1: Or just `transcodeConf.ForceSessionReinit = transcodeRequest.IsClip && int64(segment.Index) == 1`

mjh1: Do we not need a re-init for the final segment too?

mjh1 (approved):

transcode/transcode.go
leszko: > Do we not need a re-init for the final segment too (when that's also clipped)?

That's a good point. I guess that yes!

emranemran: > Do we not need a re-init for the final segment too (when that's also clipped)?
No actually, surprisingly. For some reason, the transcoder session is able to handle going from source recording segment to final segment that's also clipped (i.e re-encoded differently). Need to understand lpms better to understand why that is. Planning to follow up with clipping directly at T layer and hopefully understand this better.

PR #904, opened 2023-10-06
manifest: add DISCONTINUITY tags to clip playlist

emranemran:
A discontinuity tag is required in the clip playlist so that it shows the correct duration.

### DIFF
--- a/clients/manifest.go
+++ b/clients/manifest.go
@@ -98,7 +98,7 @@ func GetSourceSegmentURLs(sourceManifestURL string, manifest m3u8.MediaPlaylist)
 
 // Generate a Master manifest, plus one Rendition manifest for each Profile we're transcoding, then write them to storage
 // Returns the master manifest URL on success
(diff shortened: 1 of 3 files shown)


0xcadams (approved):
LGTM, retro review from Thom/Max

clients/manifest.go
```diff
@@ -146,6 +146,12 @@ func GenerateAndUploadManifests(sourceManifest m3u8.MediaPlaylist, targetOSURL s
 			}
 		}
 
+		if isClip {
+			_, totalSegs := video.GetTotalDurationAndSegments(renditionPlaylist)
+			renditionPlaylist.Segments[1].Discontinuity = true
+			renditionPlaylist.Segments[totalSegs-1].Discontinuity = true
```
thomshutt, line 152: Will we get an Array Out Of Bounds for a single segment clip? Possibly not because of the weird array preallocation stuff of the playlist library, but probably good to guard against anyway

emranemran: Yup exactly what we ran into. Fixed here: https://github.com/livepeer/catalyst-api/commit/dd521620011ac4e9f27840985fd4121c403026c6

PR #907, opened 2023-10-09
clip: fix clipping when start/end segments fall in same segment

emranemran:


0xcadams (approved):
LGTM, retroactive by Thom/Max

mjh1 (approved):

PR #909, opened 2023-10-09
Fix short segment length

mjh1:
This is to fix the case where the source has two keyframes very close together resulting in a tiny segment. The ffmpeg format `hls` that we were using unfortunately has no minimum segment duration option and just splits on every keyframe. So I've switched to `segment` which does allow you to set a minimum segment duration.

### DIFF
--- a/test/steps/ffmpeg.go
+++ b/test/steps/ffmpeg.go
@@ -83,6 +83,9 @@ func (s *StepContext) TheSourceManifestIsWrittenToStorageWithinSeconds(secs, num
 	if mediaManifest.TargetDuration != 11.0 {
 		return fmt.Errorf("expected manifest to have a Target Duration of 11 but got: %f", mediaManifest.TargetDuration)
 	}
+	if mediaManifest.MediaType != m3u8.VOD {
+		return fmt.Errorf("expected manifest to have playlist type VOD but got: %v", mediaManifest.MediaType)
+	}
 
 	return nil
 }
--- a/test/steps/ffmpeg.go
+++ b/test/steps/ffmpeg.go
@@ -80,8 +80,8 @@ func (s *StepContext) TheSourceManifestIsWrittenToStorageWithinSeconds(secs, num
 	if mediaManifest.Version() != 3 {
 		return fmt.Errorf("expected manifest to be HLSv3 but got version: %d", mediaManifest.Version())
 	}
-	if mediaManifest.TargetDuration != 10.0 {
-		return fmt.Errorf("expected manifest to have a Target Duration of 10 but got: %f", mediaManifest.TargetDuration)
+	if mediaManifest.TargetDuration != 11.0 {
+		return fmt.Errorf("expected manifest to have a Target Duration of 11 but got: %f", mediaManifest.TargetDuration)
 	}
 
 	return nil
--- a/video/segment.go
+++ b/video/segment.go
@@ -2,11 +2,12 @@ package video
 
 import (
 	"fmt"
+	"strings"
 
 	ffmpeg "github.com/u2takey/ffmpeg-go"
 )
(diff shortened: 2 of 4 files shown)


video/segment.go
```diff
@@ -15,16 +16,16 @@ func Segment(sourceFilename string, outputManifestURL string, targetSegmentSize
...
-				"hls_playlist_type": "vod",
-				"hls_list_size":     "0",
-				"hls_time":          targetSegmentSize,
-				"method":            "PUT",
+				"f":                 "segment",
+				"segment_list":      outputManifestURL,
+				"segment_list_type": "m3u8",
+				"segment_format":    "mpegts",
+				"segment_time":      targetSegmentSize,
+				"min_seg_duration":  "2",
```
emranemran, line 28: why 2?

emranemran (approved):

test/steps/ffmpeg.go
```diff
@@ -80,8 +80,11 @@ func (s *StepContext) TheSourceManifestIsWrittenToStorageWithinSeconds(secs, num
 	if mediaManifest.Version() != 3 {
 		return fmt.Errorf("expected manifest to be HLSv3 but got version: %d", mediaManifest.Version())
 	}
-	if mediaManifest.TargetDuration != 10.0 {
-		return fmt.Errorf("expected manifest to have a Target Duration of 10 but got: %f", mediaManifest.TargetDuration)
+	if mediaManifest.TargetDuration != 11.0 {
```
leszko, line 83: What is this `11.0`?

leszko (approved):

video/segment.go
mjh1: Yeah i'm not sure, I picked it for now just because I remember never seeing any lower than that in the past really. @thomshutt what do you think, is it ok to pick something pretty arbitrary like this and put it in a constant?

test/steps/ffmpeg.go
mjh1: The different ffmpeg filter seems to have slightly different behaviour with estimating the target duration. We could override it to be our [requested duration](https://github.com/livepeer/catalyst-api/pull/909/files#diff-800ac29ee96767d39950a993dc844eac954d9d93ce5cd9973fbeadf5cc7f0530L14) but I'm not sure that it matters. cc @thomshutt

thomshutt: I think that as long as it's larger than the largest segment it produces then that's fine

thomshutt: Are the segments it produces the same size as before (other than the problematic one)?

mjh1: Mostly the same but small differences. e.g. for this test:
`-f hls`
```
2023/10/11 16:03:46 #EXTM3U
#EXT-X-VERSION:3
#EXT-X-TARGETDURATION:10
#EXT-X-MEDIA-SEQUENCE:0
#EXT-X-PLAYLIST-TYPE:VOD
#EXTINF:10.000000,
index0.ts
#EXTINF:10.000000,
index1.ts
#EXTINF:10.000000,
index2.ts
#EXTINF:0.033333,
index3.ts
#EXT-X-ENDLIST
```
`-f segment`
```
2023/10/11 16:02:58 #EXTM3U
#EXT-X-VERSION:3
#EXT-X-PLAYLIST-TYPE:VOD
#EXT-X-MEDIA-SEQUENCE:0
#EXT-X-TARGETDURATION:11
#EXTINF:10.044,
index0.ts
#EXTINF:10.000,
index1.ts
#EXTINF:10.000,
index2.ts
#EXTINF:0.033,
index3.ts
#EXT-X-ENDLIST

thomshutt (approved):

PR #914, opened 2023-10-11
Limit source playback bitrate

mjh1:
Some uploaded sources have very high bitrates which wouldn't be playable over most internet connections so we should limit source playback to sensible bitrates.


thomshutt (approved):
Unit test for the new method might be nice

PR #915, opened 2023-10-11
Support multiple private buckets

mjh1:


playback/playback.go
```diff
@@ -105,17 +105,32 @@ func appendAccessKey(uri, gatingParam, gatingParamName string) (string, error) {
...
-	if err != nil {
+	if len(config.PrivateBucketURLs) < 1 {
+		return nil, errors.New("playback failed, no private buckets configured")
+	}
+
+	var (
+		err error
+		f   *drivers.FileInfoReader
+	)
+	// try all private buckets until object is found or return error
```
thomshutt, line 116: Do we have any way of hinting this or making it more efficient? I guess if we put Storj first in the list then most "hot" stuff will be there in the future

thomshutt (approved):

playback/playback.go
mjh1: Annoyingly no way of knowing the right bucket afaict, the hostname in the playback urls needs to stay the same to get routed to catalyst, could I set up a cname..? Also don't think it's worth overcomplicating the code with parallelism, it would have to be a weird issue with google/storj to slow things down, normally it's just going to be a quick failure if the file doesn't exist. I think I should at least set some timeouts though, doesn't look like they're being set ha!

victorges: Is this for access controlled playback? If so, one alternative for the future would be having the "access control hook" return the OS URL of the file as well. That one runs in the context of a playback ID and can fetch the corresponding OS ID on the Studio DB, including it in the response.

PR #916, opened 2023-10-12
metric/coordinator: add clip metric to track vod clip jobs

emranemran:


mjh1 (approved):

metrics/metrics.go
```diff
@@ -41,7 +41,7 @@ type CatalystAPIMetrics struct {
 	VODPipelineMetrics VODPipelineMetrics
 }
 
-var vodLabels = []string{"source_codec_video", "source_codec_audio", "pipeline", "catalyst_region", "num_profiles", "stage", "version", "is_fallback_mode", "is_livepeer_supported"}
+var vodLabels = []string{"source_codec_video", "source_codec_audio", "pipeline", "catalyst_region", "num_profiles", "stage", "version", "is_fallback_mode", "is_livepeer_supported", "is_clip"}
```
thomshutt, line 44: Would it make sense to make this a vod_type field? Then we can also potentially differentiate recordings

PR #921, opened 2023-10-16
Generate thumbnails with ffmpeg

mjh1:
Eventually we want to be able to generate thumbnails using our transcoders (alongside segment transcoding), for now though I'm implementing this in catalyst using ffmpeg similar to other cases.


pipeline/thumbnails_test.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/stretchr/testify/require"
+	"gopkg.in/vansante/go-ffprobe.v2"
+)
+
+func TestGenerateThumbs(t *testing.T) {
+	tests := []struct {
+		name  string
+		input string
+	}{
+		{
```
leszko, line 20: nit: do you plan to add more test scenarios here? If not, then you may skip this `tests` part and just a have a single test.

leszko (approved):
👍

pipeline/thumbnails.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	defer os.RemoveAll(tempDir)
+
+	args := []string{
+		"-i", in,
+		"-vf", "select=eq(pict_type\\,I)",
+		"-vsync", "vfr",
+		"-vf", fmt.Sprintf("fps=1/%d", framesEverySecs),
+		path.Join(tempDir, "/keyframes_%03d.jpg"),
+	}
```
emranemran, line 37: I think it would be cleaner to move these ffmpeg into it's own place like `video/thumbs.go` similar to transmux/clip/etc?

pipeline/thumbnails.go
```diff
@@ -0,0 +1,93 @@
...
+		return err
+	}
+	builder := bytes.Buffer{}
+	_, err = builder.WriteString("WEBVTT\n")
+	if err != nil {
+		return err
+	}
+	timestamp := time.Time{}
+	outputLocation := output.JoinPath("thumbnails").String()
+	for _, file := range files {
```
emranemran, line 65: Same here - a function like `GenerateVTT(filename string)` in `video/thumbs.go` could potentially generate and write the file to disk and once it returns, this function can call `clients.UploadToOSURL` to prevent recursive dependency in video package. It also makes the code cleaner and abstracts away all video-y logic away from the pipeline package. wdyt?

pipeline/thumbnails.go
emranemran: Also i think this incantation of ffmpeg is simple enough for us to use go-ffmpeg instead of cmd.Run right?

handlers/upload.go
```diff
@@ -134,33 +126,26 @@ func (r UploadVODRequest) getTargetMp4Output() (UploadVODRequestOutputLocation,
...
 
-func (r UploadVODRequest) getSourceCopyEnabled() bool {
-	for _, o := range r.OutputLocations {
-		if o.Outputs.SourceMp4 {
-			return true
-		}
-	}
-	return false
-}
-
```
emranemran, line 163: nice

emranemran:
Did you happen to check resource utilization on the pods when ffmpeg is running for somewhat long vod file? Do CPUs get pegged?

pipeline/thumbnails.go
emranemran: Also these args could use some comments to explain what they're doing.

emranemran (changes requested):

pipeline/thumbnails.go
mjh1: Ah yep all good points will do 👍

mjh1: Ah.. just remembered there are two `vf` options so can't use the map in go-ffmpeg

mjh1: > I think it would be cleaner to move these ffmpeg into it's own place like `video/thumbs.go` similar to transmux/clip/etc?

Definitely agree it should be separated, for now I've created a new package just for thumbnails though to keep it simple. Then I'll do a follow up PR to rearrange a couple of things as it seems like we might have too much going on in the video package.

mjh1:
> Did you happen to check resource utilization on the pods when ffmpeg is running for somewhat long vod file? Do CPUs get pegged?

Thanks for the reminder 😬 yeah it's not brilliant:
Screenshot 2023-10-18 at 20 50 50

I've done a fair bit of digging today though and I can't see a simple way to improve it, you can get ffmpeg to output keyframes which is obviously really efficient but then we end up with different thumbnail frequency for different videos. So might have to leave it like this for now until we move it to the transcoders, and only turn it on for specific customers that need it?

mjh1:
@emranemran I've now switched to extracting keyframes only, this reduces processing time to under 1 second for a 10 min video and ~8 seconds for a 1hr15 video I tested with. Sound ok?

emranemran (approved):

emranemran:
> @emranemran I've now switched to extracting keyframes only, this reduces processing time to under 1 second for a 10 min video and ~8 seconds for a 1hr15 video I tested with. Sound ok?

Nice, that's a great perf improvement for now. I'll begin working on moving this to T-layer as a separate effort.

PR #922, opened 2023-10-16
mapic: Add event-based nuking on serf

victorges:
Let's fix nuking once and for all. (famous last words)

### DIFF
--- a/mapic/mistapiconnector_app.go
+++ b/mapic/mistapiconnector_app.go
@@ -43,6 +43,7 @@ type (
 		Start(ctx context.Context) error
 		MetricsHandler() http.Handler
 		RefreshMultistreamIfNeeded(playbackID string)
+		NukeStream(playbackID string)
 	}
 
 	pushStatus struct {
@@ -169,6 +170,24 @@ func (mc *mac) MetricsHandler() http.Handler {
 func (mc *mac) RefreshMultistreamIfNeeded(playbackID string) {
 	if mc.streamExists(playbackID) {
 		mc.refreshMultistream(playbackID)
+	}
+}
+
+func (mc *mac) NukeStream(playbackID string) {
+	// ignore the base name from the input, in case someone happens to send one in the playbackId
+	if strings.Contains(playbackID, "+") {
+		playbackID = strings.Split(playbackID, "+")[1]
+	}
+
+	streamNames := []string{
+		mc.wildcardPlaybackID(&api.Stream{PlaybackID: playbackID}),               // not recorded
+		mc.wildcardPlaybackID(&api.Stream{PlaybackID: playbackID, Record: true}), // recorded
+	}
+
+	for _, streamName := range streamNames {
+		if err := mc.mist.DeleteStream(streamName); err != nil {
+			glog.Errorf("Error nuking stream playbackId=%s streamName=%s err=%q", streamName, playbackID, err)
+		}
 	}
 }
 
--- a/main.go
+++ b/main.go
@@ -336,6 +336,9 @@ func processClusterEvent(mapic mistapiconnector.IMac, e serf.UserEvent) {
 		case "stream":
 			mapic.RefreshMultistreamIfNeeded(eventPayload.PlaybackID)
 			return
+		case "nuke":
+			mapic.NukeStream(eventPayload.PlaybackID)
+			return
 		default:
 			glog.Errorf("unsupported serf event: %v", e)
 		}


mapic/mistapiconnector_app.go
```diff
@@ -506,18 +525,66 @@ func (mc *mac) reconcileLoop(ctx context.Context) {
...
 			glog.Errorf("error executing query on Mist, cannot reconcile err=%v", err)
 			continue
 		}
+		mc.reconcileStreams(mistState)
 		mc.reconcileMultistream(mistState)
 		mc.processStats(mistState)
 	}
 }
 
+func (mc *mac) reconcileStreams(mistState clients.MistState) {
```
leszko, line 541: I wonder if it's possible to maybe add unit tests for this? Wdyt?

mapic/mistapiconnector_app.go
```diff
@@ -506,18 +525,66 @@ func (mc *mac) reconcileLoop(ctx context.Context) {
...
 		case <-ticker.C:
-		case <-mc.multistreamUpdated:
+		case <-mc.streamUpdated:
 		}
 		mistState, err := mc.mist.GetState()
 		if err != nil {
 			glog.Errorf("error executing query on Mist, cannot reconcile err=%v", err)
 			continue
 		}
+		mc.reconcileStreams(mistState)
```
leszko, line 535: Nit: Update the doc for the `reconcileLoop(ctx context.Context)`

mapic/mistapiconnector_app.go
```diff
@@ -506,18 +525,66 @@ func (mc *mac) reconcileLoop(ctx context.Context) {
...
 		mc.reconcileMultistream(mistState)
 		mc.processStats(mistState)
 	}
 }
 
+func (mc *mac) reconcileStreams(mistState clients.MistState) {
+	for streamName, _ := range mistState.ActiveStreams {
+		// read map directly to avoid creating streamInfo object even for playback-only (pull) streams
```
leszko, line 543: I'm not sure I understand this comment. Could you maybe rephrase? We don't create any stream info here 🤔

mapic/mistapiconnector_app.go
```diff
@@ -506,18 +525,66 @@ func (mc *mac) reconcileLoop(ctx context.Context) {
...
+				var err error
+				si, err = mc.refreshStreamInfo(streamName)
+				if err != nil {
+					glog.Errorf("error refreshing stream info streamName=%s err=%v", streamName, err)
+					continue
+				}
+			}
+
+			mc.reconcileSingleStream(si)
+		}
```
leszko, line 559:
```suggestion
			if mistState.IsIngestStream(streamName) {
			si, err := mc.getStreamInfo(mistStreamName2playbackID(streamName))
			if err != nil {
				glog.Errorf("error getting stream info streamName=%s err=%v", streamName, err)
				continue
			}
			mc.reconcileSingleStream(si)
		}
```

How about this logic? Should work the same and is simpler.

mapic/mistapiconnector_app.go
victorges: Yeah ideally we would, but I won't have bandwidth for that this week since we're not officially prioritizing this as a task/project. Since this is more of a back-office flow I considered it less of a requirement for now and am willing to merge it as is. Otherwise, if no one is able to pick it up I can come back to this when back from PTO in november.

mapic/mistapiconnector_app.go
victorges: This reads from the map instead of calling `getStreamInfo`, which would fetch the info object even for the playback streams. I can rephrase it

mapic/mistapiconnector_app.go
victorges: According to this comment, it is not the same: https://github.com/livepeer/dms-api/blob/835bb9ed3a912bd7ef9d3007b7ef9dc681101abd/mapic/mistapiconnector_app.go#L839

Do you think that case is not that important? I can change to that logic it's much cleaner 😌

mapic/mistapiconnector_app.go
victorges: Removed the comment altogether with your suggested change.

mapic/mistapiconnector_app.go
victorges: OK I've decided to go forward with it. If the docs I linked are correct, worst case we will wait 30 seconds until the next reconcile loop and it should proceed with the nuke there. We will not be nuking streams that have just started to the point of a "race condition" anywya.

leszko: I think it's correct as I suggested (and as you changed it), because we iterate over Active Streams from Mist (so the scenario mentioned in the comment ☝️ is not important here).

mapic/mistapiconnector_app.go
leszko: I see. I just checked and adding the unit tests was not so difficult, so added them in 3416b6a9b6d3da4451688367a59ee3d46d0dd9e3

leszko (approved):
Looks good 👍  Though, keep in mind, I haven't tried if it works!

mapic/mistapiconnector_app.go
victorges: Awesome, thanks!

victorges:
@leszko It was tested by @iameli-streams yesterday on his daily meta-livestreams 😁 
Thanks again for adding the tests!

mapic/mistapiconnector_app_test.go
```diff
@@ -180,3 +180,84 @@ func TestReconcileMultistream(t *testing.T) {
 	require.ElementsMatch(t, expectedAutoToRemove, recordedAutoRemove)
 	require.ElementsMatch(t, expectedPushToStop, recordedPushStop)
 }
+
+func TestReconcileStreams(t *testing.T) {
```
victorges, line 184: 🎉 ❤️

PR #931, opened 2023-10-18
Add C2PA signing

leszko:
Add C2PA signing to the FFMPEG pipeline.

**How does sample output videos look like?**
- [Output video when the source video is C2PA signed](https://link.storjshare.io/raw/juwmcjzevuqxwg4hwxv3wzuci32q/out/c2pa21/270p0.mp4)
- [Output video when the source video is not C2PA signed](https://link.storjshare.io/raw/juwmcjzevuqxwg4hwxv3wzuci32q/out/c2pa12/360p0.mp4)

You can download these videos and verify in the C2PA online verify tool: https://contentcredentials.org/verify

**Note that this PR**
- Adds signing only for the FFMPEG pipeline, not yet for Mediaconvert pipeline
- Adds `c2pa.published` assertion done by Livepeer Studio, not yet `c2pa.transcoded`
- Adds C2PA signature only to the output mp4 files, not yet to HLS

**Detailed changes**
- Add `c2pa` package which uses `c2patool` tool to sign videos
- Add two flags `c2pa-private-key` and `c2pa-certs` to define paths to PK and Certs used for C2PA signature
- Add `c2pa` bool param to the VOD request to decide if the output mp4 videos should be signed
- Change the logic of `copyFileToLocalTmpAndSegment()` not to delete the local copy of the source file, because it's needed to preserve the source signature


main.go
```diff
@@ -360,3 +363,23 @@ func handleSignals(ctx context.Context) error {
 		}
 	}
 }
+
+func createC2PA(cli *config.Cli) *c2pa.C2PA {
```
thomshutt, line 367: Shall we have this also return an error to be more Go-ish? Then we can log in one place if it fails and unit test the failure cases

thomshutt (approved):

main.go
leszko: Good idea, updated in 098b892bd46f01574387dbcde8e0ca79403012f0

PR #936, opened 2023-10-23
Fix mp4 output file naming for the bitrate only case

mjh1:


mjh1:
> @mjh1 What if we have multiple mp4 outputs? Will they all have the same name? 🤔

@leszko Wouldn't they all have a different path? What kind of request are you thinking of?

leszko:
> > @mjh1 What if we have multiple mp4 outputs? Will they all have the same name? 🤔
> 
> @leszko Wouldn't they all have a different path? What kind of request are you thinking of?

I think that we use the profile name for the path of mp4. Could you try making an API call to studio like this one:
```
curl --location 'http://0.0.0.0:8888/api/transcode' \
--header 'Authorization: Bearer <API_KEY>' \
--data '{
    "input": {
        "url": "https://storage.googleapis.com/thom-vod-testing/mustwork/Pexels%20Videos%203444.mp4"
    },
    "storage": {
        "type": "s3",
        "endpoint": "https://gateway.storjshare.io",
        "credentials": {
            "accessKeyId": "<STORJ_S3_ACCESS_KEY>",
            "secretAccessKey": "<STORJ_S3_SECRET_KEY>"
        },
        "bucket": "<BUCKET>"
    },
    "outputs": {
        "mp4": {
            "path": "/mp4/"
        }
    }
}'
```

It results in having the following files in the bucket:
```
/mp4/360p0.mp4
/mp4/720p0.mp4
/mp4/1080p0.mp4
```

So, I think that after your change they might overwrite each other.

mjh1:
Ah I see, I think we're ok here as we're defining the profiles explicitly, so the name is only applicable to a single profile. the files created look like this:
profile:
```
                {
                    "name": "foo",
                    "bitrate": 400000
                }
```
`/mp4/foo.mp4`

emranemran (approved):

PR #952, opened 2023-11-06
Fix Idle Stream Issue

leszko:
Fix https://linear.app/livepeer/issue/PS-91/switchboard-idlepending-status-inaccurate
Fix https://linear.app/livepeer/issue/VID-456/recording-stuck-in-progress
Fix https://linear.app/livepeer/issue/PS-84/livespace-webhooks-spamming

It fixes the issue that a stream is Idle (`"isActive": true,`) after finishing streaming for 90s.

**Explanation**

1. In https://github.com/livepeer/catalyst-api/pull/842 we started a loop in which every `1 min` we collect metrics from Mist and then call `setActive(true)` on studio.
2. That can lead to the race condition between this `setActive(true)` and `setActive(false)` sent from `STREAM_BUFFER` trigger when the stream is stopped
3. The root cause is that at the time of sending `setActive(true)` from the metrics collector, we're actually not sure that the stream is still active.
4. In this (and two other related PRs: https://github.com/livepeer/go-api-client/pull/52, https://github.com/livepeer/studio/pull/1957) I've added a new studio endpoint `/heartbeat` used to refresh the stream `lastSeen` state without stating if it's active or not

@gioelecerati FYI: There is also a race condition between calling `/setactive` and `/hook/health`, but I've never encountered them having different values. They are both triggered with the same Mist trigger, so I think we can leave it as it is right now.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,7 @@ require (
 	github.com/hashicorp/serf v0.10.1
 	github.com/julienschmidt/httprouter v1.3.0
 	github.com/lib/pq v1.10.9
-	github.com/livepeer/go-api-client v0.4.12-0.20231106122927-b996b8714ada
+	github.com/livepeer/go-api-client v0.4.12
 	github.com/livepeer/go-tools v0.3.3
 	github.com/livepeer/joy4 v0.1.1
 	github.com/livepeer/livepeer-data v0.7.4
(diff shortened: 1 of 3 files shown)


gioelecerati:
> FYI: There is also a race condition between calling /setactive and /hook/health, but I've never encountered them having different values. They are both triggered with the same Mist trigger, so I think we can leave it as it is right now.

Curious, I saw them having different values in my box - maybe that's not happening outside of it. Thanks for helping out, the investigation & the fixes!

gioelecerati (approved):

victorges:
Neat!

PR #956, opened 2023-11-07
Move setActive(true) from the trigger 'PUSH_REWRITE' to 'STREAM_BUFFER'

leszko:
This should fix the idle stream issue in the scenario when the stream is started and suddenly stopped.

### DIFF
--- a/mapic/mistapiconnector_app.go
+++ b/mapic/mistapiconnector_app.go
@@ -196,9 +196,9 @@ func (mc *mac) handleStreamBuffer(ctx context.Context, payload *misttriggers.Str
 	}
 	if info, ok := mc.getStreamInfoLogged(playbackID); ok {
 		glog.Infof("Setting stream's manifestID=%s playbackID=%s active status to %v", info.id, playbackID, isActive)
-		_, err := mc.lapi.SetActive(info.id, isActive, info.startedAt)
-		if err != nil {
-			glog.Error(err)
+		ok, err := mc.lapi.SetActive(info.id, isActive, info.startedAt)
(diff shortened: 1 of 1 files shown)


mapic/mistapiconnector_app.go
```diff
@@ -189,27 +189,27 @@ func (mc *mac) NukeStream(playbackID string) {
...
 	if mc.baseStreamName != "" && strings.Contains(playbackID, "+") {
 		playbackID = strings.Split(playbackID, "+")[1]
 	}
 	if info, ok := mc.getStreamInfoLogged(playbackID); ok {
-		glog.Infof("Setting stream's manifestID=%s playbackID=%s active status to false", info.id, playbackID)
-		_, err := mc.lapi.SetActive(info.id, false, info.startedAt)
+		glog.Infof("Setting stream's manifestID=%s playbackID=%s active status to %v", info.id, playbackID, isActive)
+		_, err := mc.lapi.SetActive(info.id, isActive, info.startedAt)
 		if err != nil {
 			glog.Error(err)
```
thomshutt, line 201: Should we add some error context here? It looks like we don't get much from the SetActive method:

```go
func (lapi *Client) SetActive(id string, active bool, startedAt time.Time) (ok bool, err error) {
        if id == "" {
                return false, errors.New("empty id")
        }
```

thomshutt (approved):

mapic/mistapiconnector_app.go
leszko: Added in 5075f8fbf25e889db5a250e811f88a6ed5f6d324

PR #957, opened 2023-11-07
Use separate limits for regular-vod vs clipping-vod jobs

emranemran:
TODO: add/fix tests.

### DIFF
--- a/middleware/capacity.go
+++ b/middleware/capacity.go
@@ -39,7 +39,6 @@ func (c *CapacityMiddleware) HasCapacity(vodEngine *pipeline.Coordinator, next h
 				vodJobCount++
 			}
 		}
-		fmt.Println("XXX total Clip Jobs:", clipJobCount)
 
 		// Get this current request's job type (i.e. clipping or regular-vod request)
 		isClip, err := isClipRequest(r)
@@ -54,7 +53,6 @@ func (c *CapacityMiddleware) HasCapacity(vodEngine *pipeline.Coordinator, next h
 
 			if clipJobCount+int(inFlightClipReqs) >= config.MaxInFlightClipJobs {
 				w.WriteHeader(http.StatusTooManyRequests)
-				fmt.Println("XXX: too many clips in progress")
 				return
 			}
 		} else {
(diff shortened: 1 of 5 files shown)


middleware/capacity.go
```diff
@@ -1,33 +1,92 @@
...
+
 		// Keep a gauge of HTTP requests in flight
 		metrics.Metrics.HTTPRequestsInFlight.Add(1)
 		defer metrics.Metrics.HTTPRequestsInFlight.Add(-1)
 
-		requestsInFlight := c.requestsInFlight.Add(1)
-		defer c.requestsInFlight.Add(-1)
+		// Get total in-progress VOD jobs
+		allVodJobs := vodEngine.Jobs.GetJobs()
+		vodJobCount = len(allVodJobs)
```
victorges, line 33: Shouldn't `vodJobCount` be the jobs that are NOT clips? Otherwise, IIUC, what could happen on a worst case is that the clip jobs will starve the capacity for VOD jobs. 

e.g.: we have 10 clips running concurrently all the time, with clips finishing but new ones coming in. This means the `vodJobCount` here will always be higher than the limit, and only clip jobs will be accepted.

I guess this could be fixed by calculating `vodJobCount++` an an `else` around line 39 instead, when the given jobs is not a clip. WDYT?

victorges (approved):
LGTM, only a comment about how the non-clip limit is being handled

middleware/capacity.go
emranemran: Yeah I deleted some logic that accounted for that condition while i was refactoring the PR for review. Fixing now and what you suggested is simpler.

middleware/capacity_test.go
```diff
@@ -48,25 +102,44 @@ func TestItErrorsWhenNoJobCapacityAvailable(t *testing.T) {
...
-	handler := cm.HasCapacity(coordinator, next)
-
 	// Call the handler
-	responseRecorder := httptest.NewRecorder()
-	handler(responseRecorder, nil, nil)
-
+	handler(responseRecorder, vodReq, nil)
 	// Confirm we got an HTTP 429 response
 	require.Equal(t, http.StatusTooManyRequests, responseRecorder.Code)
+	// Confirm the handler didn't call the next middleware
```
thomshutt, line 121: Let's split the clipping failure mode out into a new test so that it's clearer what the cause is when a test fails

thomshutt (approved):

thomshutt:
Looks like your tests are triggering a race condition

emranemran:
> Looks like your tests are triggering a race condition

Fixed

middleware/capacity_test.go
emranemran: Done

PR #966, opened 2023-11-09
transmux: use stream based concatenation for clips

emranemran:
Occasionally, the mp4s of very short clips might stutter as it crosses segment boundaries. This was caused by incorrect PTS between segments which in turn is a result of us clipping and re-encoding the first and last segments "locally". Using a stream based concatenation instead of file based concatenation readjusts the timestamps and prevents these issues.

Note: Eventually, all mp4s will be generated using stream based concatenation if this works well for clipping.

### DIFF
--- a/video/transmux.go
+++ b/video/transmux.go
@@ -1,12 +1,14 @@
 package video
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 	"time"
 
 	ffmpeg "github.com/u2takey/ffmpeg-go"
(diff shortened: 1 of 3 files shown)


video/transmux.go
```diff
@@ -95,21 +97,89 @@ func MuxTStoFMP4(fmp4ManifestOutputFile string, inputs ...string) error {
...
-	// 1. create a .ts file for a given rendition
-	tsFile, err := os.Create(tsFileName)
-	if err != nil {
-		return totalBytes, fmt.Errorf("error creating file (%s) err: %w", tsFileName, err)
-	}
-	defer tsFile.Close()
-	// 2. for a given rendition, write all segment indices in ascending order to the single .ts file
-	for _, k := range segmentsList.GetSortedSegments() {
-		segBytes, err := tsFile.Write(segmentsList.SegmentDataTable[k])
+	if !useStreamBasedConcat {
```
thomshutt, line 102: This block is becoming pretty huge, could we split each part out into methods?

video/transmux.go
```diff
@@ -95,21 +97,89 @@ func MuxTStoFMP4(fmp4ManifestOutputFile string, inputs ...string) error {
...
+			}
 		}
-		totalBytes = totalBytes + int64(segBytes)
+		// Create a .ts file for a given rendition
+		tsFile, err := os.Create(tsFileName)
+		if err != nil {
+			return totalBytes, fmt.Errorf("error creating file (%s) err: %w", tsFileName, err)
+		}
+		defer tsFile.Close()
+		defer os.Remove(tsFileName)
```
thomshutt, line 164: In the other disk-based pieces we did, we had logic on app startup to clear up any dangling files that got left behind from interrupted processes. Are we able to do that here?

video/transmux.go
emranemran: Sure, I split out the ffmpeg bits into it's own separate fcn and left file management bits in the main fcn.

video/transmux.go
emranemran: There's several deferred os.Removes in the top level fcn calls that ensures all dangling files are cleaned up if there's a panic/exit. But I didn't see any app startup logic to delete - can you point me to it?

emranemran:
Updates: 
* addressed comments
* working on integration tests (might use a separate PR but trying to just get it done in this one)

emranemran:
**Updates**:
- Added unit test since it made more sense than cucumber test for this case

video/transmux.go
emranemran: Yeah, we're not doing this for clip/thumbnails/transmux/etc. Discussed with @mjh1 on a way to do this better and I think we'll create a top level dir that we can remove from init() all at one. Filed a separate ticket to get this done (VID-474).

mjh1 (approved):

PR #969, opened 2023-11-09
Increase VTT thumbnail resolution

mjh1:


thomshutt (approved):

thomshutt:
Looks like you need to update a test with the new value

PR #976, opened 2023-11-14
Balancer jam

iameli:


balancer/catalyst/catalyst_balancer.go
```diff
@@ -1,5 +1,7 @@
 package catalyst
```
iameli, line 1: Probably `catalyst_balancer` yeah?

balancer/catalyst/geo.go
```diff
@@ -0,0 +1,56 @@
...
+
+	// Order nodes by distance
+	sort.Slice(nodes, func(i, j int) bool {
+		return nodes[i].GeoDistance < nodes[j].GeoDistance
+	})
+
+	// Assign scores
+	var baseDistance = nodes[0].GeoDistance
+	var goodDistance = baseDistance + 1500
+	var okayDistance = baseDistance + 7500
```
pwilczynskiclearcode, line 44: Mist approach was different (from what [Jaron described](https://discord.com/channels/423160867534929930/1165251383956602950/1166714774650900532)). It doesn't use hard thresholds but sums different parameters and its weights.

If we want hard if-else logic with thresholds it would be nice to still make it configurable like mist weights so we wouldn't need redeployments for tuning this algorithm

iameli: Agreed, even if we want this:

```
0km = 2
1500km = 1
7500km = 0
```

No reason not to draw those points on a graph and interpolate between them, yeah? `750km = 1.5` kind of deal

balancer/catalyst/catalyst_balancer.go
thomshutt: The Golang way is `catalyst.Balancer` right?

iameli: "Folder names and package names should match" is a convention but stuff violates it all the time like `"github.com/peterbourgon/ff/v3"` being imported like that but used as `ff`. This defaulting to being imported as `catalyst` is weird to me because lots of things are Catalyst. Weeeee could call it `catabalancer` as an homage to the Istanbul cats? 🐈

thomshutt: 🐱 sold

balancer/catalyst/catalyst_balancer.go
```diff
@@ -0,0 +1,171 @@
...
+	return 2
+}
+
+func SelectNode(nodes []Node, streamID string, requestLatitude, requestLongitude float64) (Node, error) {
+	if len(nodes) == 0 {
+		return Node{}, fmt.Errorf("no nodes to select from")
+	}
+
+	topNodes := selectTopNodes(nodes, streamID, requestLatitude, requestLongitude, 3)
+	return topNodes[rand.Intn(len(topNodes))].Node, nil
```
pwilczynskiclearcode, line 120: I'd add some logging for live debugging/analysing selection mechanism:
```go
...
topNodes := selectTopNodes(nodes, streamID, requestLatitude, requestLongitude, 3)
glog.V(7).Infof("Selecting node: all available: %v", nodes)
glog.V(5).Infof("Selecting node: topNodes: %v", topNodes)
selected : = topNodes[rand.Intn(len(topNodes))].Node
glog.V(3).Infof("Selecting node: selected %v", selected)
return selected, nil
```

events/events.go
```diff
@@ -0,0 +1,126 @@
...
+				log.LogNoRequestID("catabalancer failed to get sys info", "err", err)
+				break
+			}
+			log.LogNoRequestID("catabalancer sys info ", "cpus", len(sysinfo.CPUInfo))
+
+			event := NodeStatsEvent{
+				Resource: "nodeStats",
+				NodeID:   nodeName,
+				NodeMetrics: catalyst.NodeMetrics{
+					CPUUsagePercentage:       int64(sysinfo.LoadInfo.Load5 * 100), // TODO how can we convert load into a percentage? i think we need to take into account the number of cpus
```
pwilczynskiclearcode, line 103: Load and CPU percentage usage are different metrics.
[Load and https://superuser.com/questions/1402079/linux-understanding-load-average-and-cpu](https://superuser.com/questions/1402079/linux-understanding-load-average-and-cpu)

You're probably looking for https://pkg.go.dev/github.com/shirou/gopsutil/v3@v3.23.10/cpu#Percent

emranemran: Yeah we need both CPU usage (to determine current cpu load) and load averages (to determine if load is going up or down). @mjh1 I can fix this up tomorrow. Will also add something for the bandwidth usage bits.

balancer/catalyst/catalyst_balancer_test.go
```diff
@@ -0,0 +1,346 @@
...
+		BandwidthUsagePercentage: 100,
+	},
+	Streams: map[string]Stream{},
+}
+
+func TestItReturnsItselfWhenNoOtherNodesPresent(t *testing.T) {
+	// Make the node handling the request unfavourable in terms of stats, to make sure
+	// it'll still pick itself if it's the only option
+	_, err := SelectNode([]ScoredNode{}, "some-stream-id", 0, 0)
+	require.EqualError(t, err, "no nodes to select from")
```
iameli-streams, line 48: This seems to contradict the test title - do we want to return ourself or that it errors out?

balancer/catalyst/sysinfo.go
```diff
@@ -0,0 +1,58 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Get load info
+	loadInfo, err := load.Avg()
+	if err != nil {
+		return nil, err
+	}
+	// TODO somebody check my maths here. taking the load average over 5 minutes, which is in the range 0-1 so multiply by 100 and divide by number of CPUs
```
iameli-streams, line 52: > which is in the range 0-1

I don't think this assumption is correct - I have a load average of 1.7 on the computer I'm typing this on right now. I just started compiling ffmpeg and now I'm at 5.85.

iameli-streams (changes requested):
This is starting to look really really solid! I have a couple organizational things in my head:

* What would you think about having the "what streams are where" state live in the _cluster_ package rather than the _balancer_ package? (Or potentially in another package entirely, if `cluster` is getting too bloated.) Balancer already calls into cluster, so there's no problem with making the data accessible there. A couple reasons:
* 1. Better separation of concerns; balancer just asks questions about what nodes have what streams rather than having to think about locking on its internal state.
* 2. This would also make it easier to implement alternative balancer algorithms in the future. 
* 3. We want to start implementing more kinds of events, such as informing a node that it's "retired" and shouldn't handle any more production traffic. This information is certainly relevant to the balancer but not _only_ to the balancer; we could imagine a message that makes us want to dump all of our existing users or something like that.

I could also live with doing that stuff as a refactor after the fact if we want to ship more quickly.

* It seems like the missing bit from the testing is some actual _simulation_ - given reasonable assumptions about the per-user CPU/memory/bandwidth consumption of a stream, what happens when we get 10,000 viewers all at once? I'll take a timeboxed swing at this today and see if I can make any progress.

balancer/catalyst/catalyst_balancer.go
mjh1: Yep good idea 👍 just want to check @iameli do you think it'll be ok to use logging rather than prometheus metrics? I'm just wondering if we're normally ok with those sorts of volumes of logging, it'll only be called once per playback session right so not too bad?

balancer/catalyst/catalyst_balancer.go
```diff
@@ -0,0 +1,272 @@
...
+		c.IngestStreams[id][streamID] = Stream{ID: streamID, LastSeen: time.Now()}
+		return
+	}
+	if c.Streams[id] == nil {
+		c.Streams[id] = make(Streams)
+	}
+	c.Streams[id][streamID] = Stream{ID: streamID, LastSeen: time.Now()}
+}
+
+func removeOldStreams(streams Streams) {
```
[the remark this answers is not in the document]
mjh1: Ah thanks good spot, will fix that!

balancer/catalyst/catalyst_balancer.go
```diff
@@ -0,0 +1,272 @@
...
+	for s, stream := range streams {
+		if stream.LastSeen.Before(time.Now().Add(-streamTimeout)) {
```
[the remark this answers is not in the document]
mjh1: Ah, Sub is slightly different but I can do: `time.Now().Sub(stream.LastSeen) >= streamTimeout`

balancer/catalyst/sysinfo.go
mjh1: whoops yep, Emran has sorted this out now 👍

balancer/catalyst/catalyst_balancer.go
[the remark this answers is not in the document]
mjh1: I really wanted to avoid having to acquire locks for the codepaths associated with end user requests, so I've removed the calls to `removeOldStreams` in those cases and instead just check if the stream is expired based on `LastSeen`. Does that sound ok?
We'll be getting calls to `UpdateStreams` every 5s for each other node where we do call `removeOldStreams` so we'll be cleaning up old memory that way.

balancer/catalyst/sysstats.go
```diff
@@ -0,0 +1,92 @@
...
+	systemUsage.LoadAvg = LoadAverage{
+		Load1Min:  avg.Load1,
+		Load5Min:  avg.Load5,
+		Load15Min: avg.Load15,
+	}
+
+	return systemUsage, nil
+}
+
+func GetSystemInfo() (*SystemInfo, error) {
```
mjh1, line 63: @emranemran should we just delete this func now it's unused?

balancer/catalyst/catalyst_balancer_test.go
mjh1: Ah whoops yeah looks like we meant to come back to this. Returning ourself seems like a better option if that's possible, the balancer could have an emergency `SelfNodeName` to return if nothing else is found?

mjh1:
> Contributor

For the simulation are you thinking a go unit test essentially making thousands of calls to `GetBestNode`?

balancer/catalyst/catalyst_balancer.go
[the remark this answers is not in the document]
mjh1: ahhh no looks like i have to use my way 🤦 
```
golangci-lint run
balancer/catalyst/catalyst_balancer.go:265:9: S1012: should use `time.Since` instead of `time.Now().Sub` (gosimple)
```

mjh1:
@iameli thats a really good point about the stream state refactor, definitely makes sense to do but yeah I think it would be better to save it for a later PR. 
let me know what you think of the changed I’ve made, do you think it’s ready to merge and I’ll create tickets to follow up with the remaining bits?

iameli:
> For the simulation are you thinking a go unit test essentially making thousands of calls to GetBestNode?

@mjh1 That'd be a piece of it, but we'd want to simulate actual CPU/RAM/bandwidth load so we can see how the algorithm performs given some real-world scenarios. Probably we'd want multiple nodes in that test too so we can see the result on the 5s sync time; are we going to totally overload our first nodes before it reports about incoming usage and gets overwhelmed? That kind of thing.

balancer/catalyst/catalyst_balancer.go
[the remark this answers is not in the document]
mjh1: Ok actually that's no good, go's concurrent read and write detector will kick in. Already seen it happen in staging. So i'll have to just minimise the amount of time we need to hold the locks.

balancer/catalyst/catalyst_balancer.go
```diff
@@ -0,0 +1,283 @@
...
+	"sync"
+	"time"
+)
+
+type CataBalancer struct {
+	Nodes         map[string]*Node
+	nodesLock     sync.Mutex
+	Streams       map[string]Streams     // Node name -> Streams
+	IngestStreams map[string]Streams     // Node name -> Streams
+	NodeMetrics   map[string]NodeMetrics // Node name -> NodeMetrics
```
pwilczynskiclearcode, line 20: https://pkg.go.dev/sync#Map - might be better for thread safety with its functions. Just sth to be considered in future.

balancer/catalyst/catalyst_balancer.go
mjh1: done!

mjh1:
I've started creating tickets under https://linear.app/livepeer/issue/VID-478/catabalancer for the remaining TODOs

thomshutt (approved):

PR #984, opened 2023-11-21
Debug handler

mjh1:


handlers/debug.go
```diff
@@ -9,6 +9,7 @@ import (
 	"github.com/livepeer/catalyst-api/pipeline"
 )
 
+// TODO add back in
 // DebugEndpointListenAndServe provides a handler to allow us to check internal state of the app for debugging
 func DebugEndpointListenAndServe(port int, vodEngine *pipeline.Coordinator) error {
```
mjh1, line 14: @iameli this is no longer enabled but we used to have this endpoint to be able to get a snapshot of the current jobs on the app. Previously I just gave it a new port, but do you think it would be safer and more sensible to add the handlers onto the existing internal port 7979?

mjh1: @victorges we used to have these endpoints for manually checking the internal job state. I know recently with your vod processing debugging you were looking in this area so I'm wondering if you think it would still be useful to have this and add it back in? I know we have a graph of the number of jobs being processed, but there isn't an easy way to see details of those jobs right?

PR #986, opened 2023-11-22
access-control: implement in-memory rate limiting

gioelecerati:
This should fix rate limiting, handling it in memory in catalyst instead of Studio
Goes along with https://github.com/livepeer/studio/pull/1982

### DIFF
--- a/handlers/accesscontrol/access-control.go
+++ b/handlers/accesscontrol/access-control.go
@@ -14,6 +14,7 @@ import (
 	"time"
 
 	"github.com/golang-jwt/jwt/v4"
+	"github.com/golang/glog"
 	"github.com/livepeer/catalyst-api/config"
 	"github.com/livepeer/catalyst-api/handlers/misttriggers"
 	"github.com/livepeer/catalyst-api/log"
@@ -64,7 +65,23 @@ var hitRecordCache = HitRecords{
 	data: make(map[string]*HitRecord),
 }
 
+func periodicCleanUpHitRecordCache() {
+	go func() {
+		for {
+			time.Sleep(time.Duration(30) * time.Second)
+			hitRecordCache.mux.Lock()
+			hitRecordCache = HitRecords{
+				data: make(map[string]*HitRecord),
+			}
+			hitRecordCache.mux.Unlock()
+		}
+	}()
+}
+
 func NewAccessControlHandlersCollection(cli config.Cli) *AccessControlHandlersCollection {
+
+	periodicCleanUpHitRecordCache()
+
 	return &AccessControlHandlersCollection{
 		cache: make(map[string]map[string]*PlaybackAccessControlEntry),
 		gateClient: &GateClient{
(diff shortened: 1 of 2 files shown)


thomshutt (approved):
Got it, thanks! LGTM once linting is fixed

handlers/accesscontrol/access-control.go
```diff
@@ -107,6 +121,18 @@ func (ac *AccessControlHandlersCollection) isAuthorized(ctx context.Context, pla
 		jwt = payload.JWT
 	}
 
+	if _, ok := hitRecordCache.data[playbackID]; ok {
+		hitRecordCache.mux.Lock()
+
+		if len(hitRecordCache.data[playbackID].hits) >= hitRecordCache.data[playbackID].rateLimit {
+			return false, nil
```
pwilczynskiclearcode, line 128: You wont' unlock `mux` if returning in here? right?
It's good to use `defer c.mux.Unlock()` like you did in cleanupCache() if not optimising for earliest-possible unlock

gioelecerati: Right, good point! I added the defer unlock and fixed the cucumber tests failing - will add a unit test in a follow up PR

handlers/accesscontrol/access-control.go
```diff
@@ -107,6 +121,19 @@ func (ac *AccessControlHandlersCollection) isAuthorized(ctx context.Context, pla
...
+	if _, ok := hitRecordCache.data[playbackID]; ok {
+		hitRecordCache.mux.Lock()
+		defer hitRecordCache.mux.Unlock()
+
+		if len(hitRecordCache.data[playbackID].hits) >= hitRecordCache.data[playbackID].rateLimit {
+			return false, nil
+		}
+
+		hitRecordCache.data[playbackID].hits = append(hitRecordCache.data[playbackID].hits, time.Now())
+		hitRecordCache.mux.Unlock()
```
thomshutt, line 133: Do we need this double unlock?

handlers/accesscontrol/access-control.go
```diff
@@ -198,26 +251,44 @@ func (ac *AccessControlHandlersCollection) cachePlaybackAccessControlInfo(playba
 	return nil
 }
 
-func (g *GateClient) QueryGate(body []byte) (bool, int32, int32, error) {
+func (g *GateClient) QueryGate(body []byte) (bool, int32, int32, int32, error) {
```
thomshutt, line 254: Could you add a comment on what these return params are please?

handlers/accesscontrol/access-control.go
```diff
@@ -181,12 +208,38 @@ func isStale(entry *PlaybackAccessControlEntry) bool {
 	return entry != nil && time.Now().After(entry.MaxAge) && !isExpired(entry)
 }
 
+func (c *HitRecords) cleanUpCache() {
```
thomshutt, line 211: Shall we put something in to only let this run every X seconds? Otherwise we're running over this locking loop really frequently in a high traffic situation

handlers/accesscontrol/access-control.go
gioelecerati: Sure, adding them

Considering that they grew and soon they are going to include other params (refreshInterval, for example) I want to put them into a separate `type` in the next PR

handlers/accesscontrol/access-control.go
gioelecerati: Right! removed it on line 133

PR #996, opened 2023-12-02
access-control: make access key as header work

gioelecerati:
This change finally make access keys and JWTs working as headers
Now that mist accepts `Livepeer-Access-Key` and `Livepeer-Jwt` as headers (https://github.com/livepeer/mistserver/commit/2e74b3cab6d95f3ed4f4e43b2aa208e54a52c5ba), catalyst-api is able to fetch them from the cookies.

Related docs https://github.com/livepeer/docs/pull/471

### DIFF
--- a/handlers/misttriggers/user_new.go
+++ b/handlers/misttriggers/user_new.go
@@ -17,7 +17,6 @@ type UserNewPayload struct {
 	URL          *url.URL
 	FullURL      string
 	SessionID    string
-	Cookies      []*http.Cookie
 	AccessKey    string
 	JWT          string
 }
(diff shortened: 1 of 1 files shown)


handlers/misttriggers/user_new.go
```diff
@@ -47,10 +46,17 @@ func ParseUserNewPayload(payload MistTriggerBody) (UserNewPayload, error) {
 func (d *MistCallbackHandlersCollection) TriggerUserNew(ctx context.Context, w http.ResponseWriter, req *http.Request, body MistTriggerBody) {
 	payload, err := ParseUserNewPayload(body)
 	cookies := req.Cookies()
-	accessKey := req.Header.Get("X-Livepeer-Access-Key")
-	jwt := req.Header.Get("X-Livepeer-JWT")
 
-	payload.Cookies = cookies // would remove probably when everything's working
+	var accessKey, jwt string
+	for _, cookie := range cookies {
```
victorges, line 51: So is the new method using cookies or headers on the player side? From the PR description it looks like Mist grabs the header and puts it on the cookie for the `USER_NEW` request? Any reason to translate the header to a cookie? Or is this actually making access key/jwt work as a cookie on the player side? If the latter, please update the PR title/description to reflect that.

victorges (approved):

handlers/misttriggers/user_new.go
gioelecerati: On the player side, it's a header - mist then transform that header as a cookie and pass it in the USER_NEW trigger!

Related thread: https://discord.com/channels/423160867534929930/1171019521151225856/1176876688991330405

PR #997, opened 2023-12-04
transcode: memory usage optimization

emranemran:
Transcoded segments are held in memory until the MP4 generation stage (after transcode stage is completed). This results in large usage in RAM when processing very long recordings that are several hours in length. This can occasionally result in kube pods hitting OOM.

To optimize this behavior, this PR makes the following changes:

* Create a buffered channel that holds transcoded segments
* Launch a disk-writing goroutine that consumes the buffered channel and writes segments to disk.
* Delete the byte array of the segment's data in hopes the garbage collector releases it from memory.
* Switch all MP4 generation to use stream-based concatenation instead of file-based concatenation.

### DIFF
--- a/video/transmux.go
+++ b/video/transmux.go
@@ -138,19 +138,26 @@ func ConcatTS(tsFileName string, segmentsList *TSegmentList, useStreamBasedConca
 		}()
 
 		// Write each segment to disk and add segment filename to the text file
-		for segName, segData := range segmentsList.GetSortedSegments() {
+		for segName := range segmentsList.GetSortedSegments() {
 			// Open a new file to write each segment to disk
 			segmentFilename := fileBaseWithoutExt + "_" + strconv.Itoa(segName) + ".ts"
-			segmentFile, err := os.Create(segmentFilename)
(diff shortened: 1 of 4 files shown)


video/transmux.go
```diff
@@ -138,19 +138,26 @@ func ConcatTS(tsFileName string, segmentsList *TSegmentList, useStreamBasedConca
...
+						if err != nil {
+							return totalBytes, fmt.Errorf("error creating individual segment file (%s) err: %w", segmentFilename, err)
+						}
+						defer segmentFile.Close()
+						// Write the segment data to disk
+						segBytes, err := segmentFile.Write(segmentsList.SegmentDataTable[segData])
+						if err != nil {
+							return totalBytes, fmt.Errorf("error writing segment %d err: %w", segName, err)
+						}
+			*/
```
victorges, line 154: Can we remove this dead code?

victorges (approved):
Nice!

video/transmux.go
emranemran: Yup, forgot to delete.

PR #998, opened 2023-12-04
Use errgroup to write segments to disk in background

mjh1:
Replace the channel based logic with errgroup, makes it a little simpler and easier to reason about.

### DIFF
--- a/transcode/transcode.go
+++ b/transcode/transcode.go
@@ -8,6 +8,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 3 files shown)


mjh1:
I just find it much easier to follow and really prefer to avoid channels wherever possible. I don't think the current implementation surfaces errors whereas with errgroup this it just built in. https://github.com/livepeer/catalyst-api/pull/998/files#diff-fb499d09ef134f20877abedf55ee66c60683bec63795c2cc7db80e5cb0b9c0a0L201

thomshutt (approved):

emranemran:
> I just find it much easier to follow and really prefer to avoid channels wherever possible. I don't think the current implementation surfaces errors whereas with errgroup this is just built in. current:
> 
> https://github.com/livepeer/catalyst-api/blob/513a513e5e30937f893ccada8fa92b2fd9e974d5/transcode/transcode.go#L211-L214

I'm the opposite - channels just kind of fit nicely into the produce-consumer model and the framework sort of made sense to me and I was able to reason around it. The errors are captured in various stages in the channel implementation and surfaced up (e.g. fail to get segment, or fail to write incorrect number of segments, etc). 

What's the test plan if we decide to merge this? The main reason the original change was done was to handle OOM conditions and the refactor meant a lot of the logic around mp4s had to be re-tested i.e. short and very long MP4s will need to be tested. 

Some other things that come to mind: 
* Resource usage - currently we never exceed 40% per pod during typical usage + mp4 generation. 
* Are transcoded segments arriving out of order handled? 
* Verifying all renditions are the same length and playable.

PR #1002, opened 2023-12-05
cors: StatusOK on preflight requests

gioelecerati:


iameli (commented):
Tests failing but broadly LGTM

iameli (approved):
LGTM but I'm assuming you've just made someone else's `accept-ranges` code obsolete? SOMEbody was adding that field. Might be worth finding that and removing it (and checking that there's nothing else happening there that you're missing)

gioelecerati:
> LGTM but I'm assuming you've just made someone else's `accept-ranges` code obsolete? SOMEbody was adding that field. Might be worth finding that and removing it (and checking that there's nothing else happening there that you're missing)

If I understood correctly, that test was set for the VOD and webrtc preflight requests and it's still valid in http.go - my change was related to live streaming

Maybe @hjpotter92 you can chime and take a look, as it was your change?

iameli:
> If I understood correctly, that test was set for the VOD and webrtc preflight requests and it's still valid in http.go - my change was related to live streaming

If I understand correctly your change is in the CORS middleware and so will affect EVERY HTTP request comnig through the pipeline. Which is fine, I think we want to have permissive CORS in all cases, but I'm just guessing that the other "OPTIONS" handlers are now dead code

PR #1003, opened 2023-12-06
[VID-483] CDN-fallback improvements

pwilczynskiclearcode:
Two new features added:
* setting percentage of traffic that goes to CDN and directly 
* redirecting to per-catalyst CDN URL - to distribute the traffic around the world instead of single `fra-0` - [terraform changes](https://github.com/livepeer/livepeer-infra/pull/1798) ... mist doesn't support switching nodes mid-playback... so we need to redirect a view to the closest's node CDN zone.

https://linear.app/livepeer/issue/VID-483/cdn-fallback-to-balance-between-catalyst-removing-fra-0-spof


handlers/geolocation/gelocation_test.go
```diff
@@ -391,25 +392,32 @@ func TestCdnRedirectHLS(t *testing.T) {
...
 	n.Config.CdnRedirectPrefix, _ = url.Parse("https://external-cdn.com/mist")
-	n.Config.CdnRedirectPlaybackIDs = []string{CdnRedirectedPlaybackID}
+	n.Config.CdnRedirectPlaybackPct = map[string]float64{CdnRedirectedPlaybackID: 1.0}
 
 	// this playbackID is configured to be redirected to CDN
 	requireReq(t, fmt.Sprintf("/hls/%s/index.m3u8", CdnRedirectedPlaybackID)).
 		result(n).
 		hasStatus(http.StatusTemporaryRedirect).
-		hasHeader("Location", fmt.Sprintf("http://external-cdn.com/mist/hls/video+%s/index.m3u8", CdnRedirectedPlaybackID))
+		hasHeader("Location", fmt.Sprintf("http://someurl.com.external-cdn.com/mist/hls/video+%s/index.m3u8", CdnRedirectedPlaybackID))
```
pwilczynskiclearcode, line 401: On bunny side I'd need to setup plenty of hostnames:
```
fra-prod-catalyst-0.lp-playback.studio.redirect.lp-playback.studio
fra-prod-catalyst-1.lp-playback.studio.redirect.lp-playback.studio
fra-prod-catalyst-2.lp-playback.studio.redirect.lp-playback.studio
lon-prod-catalyst-0.lp-playback.studio.redirect.lp-playback.studio
lon-prod-catalyst-1.lp-playback.studio.redirect.lp-playback.studio
lon-prod-catalyst-2.lp-playback.studio.redirect.lp-playback.studio
...
```
and write Bunny Rules that matching these hostnames and serving content from individual catalysts

main.go
```diff
@@ -98,7 +100,7 @@ func main() {
 	fs.StringVar(&cli.NodeName, "node", hostname, "Name of this node within the cluster")
 	config.SpaceSliceFlag(fs, &cli.BalancerArgs, "balancer-args", []string{}, "arguments passed to MistUtilLoad")
 	fs.StringVar(&cli.NodeHost, "node-host", "", "Hostname this node should handle requests for. Requests on any other domain will trigger a redirect. Useful as a 404 handler to send users to another node.")
-	config.CommaSliceFlag(fs, &cli.CdnRedirectPlaybackIDs, "cdn-redirect-playback-ids", []string{}, "PlaybackIDs to be redirected to CDN.")
+	config.CommaSliceFlag(fs, &cli.CdnRedirectPlaybackIDs, "cdn-redirect-playback-ids", []string{}, "PlaybackIDs to be redirected and percentage of traffic. E.g. 'dbe3q3g6q2kia036:100,6736xac7u1hj36pa:0.01'")
```
victorges, line 103: Independent of that, wdyt of declaring a local variable here to hold the result of this `CommaSliceFlag`, so the `cli` struct doesn't maintain an intermediary field as if it were a separate option?
```suggestion
        var cdnRedirectPlaybackIDs string
	config.CommaSliceFlag(fs, &cdnRedirectPlaybackIDs, "cdn-redirect-playback-ids", []string{}, "PlaybackIDs to be redirected and percentage of traffic. E.g. 'dbe3q3g6q2kia036:100,6736xac7u1hj36pa:0.01'")
```

main.go
```diff
@@ -143,6 +145,25 @@ func main() {
...
+	cli.CdnRedirectPlaybackPct = make(map[string]float64)
+
+	for _, redirect := range cli.CdnRedirectPlaybackIDs {
+		elements := strings.Split(redirect, ":")
+		if len(elements) == 2 {
+			percentage, err := strconv.ParseFloat(elements[1], 64)
+			if err != nil || percentage > 100.0 || percentage < 0.0 {
+				glog.Fatalf("Invalid config %s - should be between 0.0 and 100.00", redirect)
+			} else {
+				cli.CdnRedirectPlaybackPct[elements[0]] = percentage / 100.0
```
victorges, line 157: Oh I see... I was confused on whether this was a 0-100 or 0-1 value. Might be worth a comment in the `CdnRedirectPlaybackPct` explaining that it's provided as 0-100 but stored as 0-1 for the load balancing logic

pwilczynskiclearcode:
@victorges I rewritten it slightly. Storing `0.0` to `100.0` in memory, instead of 0..1.

config/cli.go
```diff
@@ -212,6 +217,36 @@ func CommaSliceFlag(fs *flag.FlagSet, dest *[]string, name string, value []strin
 	})
 }
 
+// handles -foo=value1:10.3,value2:99.9,value3:0.1,value4:100,value5:0
+func CommaWithPctSliceFlag(fs *flag.FlagSet, dest *map[string]float64, name string, value map[string]float64, usage string) {
```
victorges, line 221: neat!

config/cli.go
```diff
@@ -212,6 +217,36 @@ func CommaSliceFlag(fs *flag.FlagSet, dest *[]string, name string, value []strin
...
+			return nil
+		}
+		for _, redirect := range elements {
+			elements := strings.Split(redirect, ":")
+			if len(elements) == 2 {
+				percentage, err := strconv.ParseFloat(elements[1], 64)
+				if err != nil || percentage > 100.0 || percentage < 0.0 {
+					return fmt.Errorf("invalid config %s - should be between 0.0 and 100.00", redirect)
+				} else {
+					(*dest)[elements[0]] = percentage
```
victorges, line 236: **nit:** This will actually mutate the map provided as default (`value` arg). If that value is provided as `nil`, this will also panic.

Maybe we want to always initialize the `*dest` on the beginning of this function (aka move line 226 to 224) instead.

victorges (approved):
re-LGTM! just one nit on the flag parsing func which is not really a problem in the current usage

config/cli.go
pwilczynskiclearcode: I think the mutation isn't that problem in this context. I'd need to copy the default map contents to make it mutation-safe... I think we can have simple implementation as `CommaWithPctSliceFlag()` is called only once

PR #1006, opened 2023-12-06
clients: ignore SPS errors when probing input files

emranemran:
SPS errors are indicative of an encoding issue but typically don't cause issues with transcode and playback. We should ignore these as they're already ignored post-segment and pre-transcode stage.

Follow-on change from VID-450 which was fixed here: https://github.com/livepeer/catalyst-api/pull/928/files


mjh1 (commented):
Does it work if we just add this new message here: https://github.com/livepeer/catalyst-api/blob/8719227911ac2722f670da000fc533cae4ad6c8e/clients/input_copy.go#L27

PR #1007, opened 2023-12-07
Fix source load balancing

mjh1:


balancer/catabalancer/catalyst_balancer.go
```diff
@@ -276,10 +279,15 @@ func (c *CataBalancer) MistUtilLoadSource(ctx context.Context, stream, lat, lon
 	defer c.nodesLock.Unlock()
 	for _, node := range c.Nodes {
 		if s, ok := c.IngestStreams[node.Name][stream]; ok && !isStale(s.Timestamp, c.metricTimeout) {
-			return node.DTSC, nil
+			dtsc := node.DTSC
+			if dtsc == "" {
+				dtsc = "dtsc://" + node.Name
```
mjh1, line 284: @iameli is this a valid thing to do? I found with the multi node test in the catalyst repo that there wasn't a dtsc tag populated. So are there some situations where this happens and this fallback is ok? Or should we fix the test to include dtsc tags?

iameli: Hmmmmm it'd still be nice to be able to support non-standard ports on the DTSC address.

mjh1: Yeah definitely, bit annoying but I've made a ticket to come back to it (in the interests of time) https://linear.app/livepeer/issue/VID-485/make-use-of-the-dtsc-serf-tag

PR #1015, opened 2023-12-11
-no-cdn-redirect-prefix-catalyst-subdomain flag (inverted)

pwilczynskiclearcode:
https://github.com/livepeer/catalyst-api/blob/main/config/cli.go#L302-L303
> // MistController has trouble giving us `-boolean-flag=false` values, so we use `-no-boolean-flag=true` instead 🤷‍♂️


victorges (approved):
oh yeah mist does that

PR #1019, opened 2023-12-12
Check stream field for CDN routing

mjh1:


api/http.go
```diff
@@ -17,8 +18,8 @@ import (
 	"github.com/livepeer/catalyst-api/pipeline"
 )
 
-func ListenAndServe(ctx context.Context, cli config.Cli, vodEngine *pipeline.Coordinator, bal balancer.Balancer, c cluster.Cluster) error {
-	router := NewCatalystAPIRouter(cli, vodEngine, bal, c)
+func ListenAndServe(ctx context.Context, cli config.Cli, vodEngine *pipeline.Coordinator, bal balancer.Balancer, c cluster.Cluster, mapic mistapiconnector.IMac) error {
```
iameli, line 21: Not a blocker but this is another area where I wish we had a more generic "state" struct that was getting passed around — rather than this HTTP handler needing to learn what a "mapic" is, we'd have a more generic (and easily-mockable) struct where it can just ask "is stream ABC123 routed to CDN?"

mjh1:
Not useful enough to continue with for now

PR #1042, opened 2023-12-18
Condense stream update messages into a single message

thomshutt:
To try and avoid overloading Serf

### DIFF
--- a/events/events.go
+++ b/events/events.go
@@ -3,6 +3,7 @@ package events
 import (
 	"encoding/json"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/hashicorp/serf/serf"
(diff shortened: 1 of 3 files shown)


events/events.go
```diff
@@ -139,29 +159,34 @@ func StartMetricSending(nodeName string, latitude float64, longitude float64, c
...
-					continue
+			streamsEvent := NodeStreamsEvent{
+				Resource: nodeStreamsEventResource,
+				NodeID:   nodeName,
+			}
+			var nonIngestStreams, ingestStreams []string
+			for streamID := range mistState.ActiveStreams {
+				if mistState.IsIngestStream(streamID) {
+					ingestStreams = append(ingestStreams, streamID)
 				}
```
mjh1, line 171: Are you missing adding to `nonIngestStreams`?

thomshutt: oof great catch, fixed

mjh1 (approved):

PR #1045, opened 2023-12-18
analytics: write to database nonblocking

iameli:
So that if it's slow we don't block responding to Mist's USER_END.

### DIFF
--- a/handlers/analytics/user_end.go
+++ b/handlers/analytics/user_end.go
@@ -28,6 +28,11 @@ func (a *AnalyticsHandler) HandleUserEnd(ctx context.Context, payload *misttrigg
 
 	// No need to block our response to Mist; everything else in a goroutine
 	go func() {
+		defer func() {
+			if rec := recover(); rec != nil {
+				glog.Errorf("panic writing to analytics database err=%s payload=%v", rec, payload)
+			}
+		}()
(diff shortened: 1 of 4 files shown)


handlers/analytics/user_end.go
```diff
@@ -25,41 +26,44 @@ func (a *AnalyticsHandler) HandleUserEnd(ctx context.Context, payload *misttrigg
...
-		payload.StreamNames[len(payload.StreamNames)-1], // stream_id
-		len(payload.StreamNames),                        // stream_id_count
-		payload.Protocols[len(payload.Protocols)-1],     // protocol
-		len(payload.Protocols),                          // protocol_count
-		payload.IPs[len(payload.IPs)-1],                 // ip_address
-		len(payload.IPs),                                // ip_address_count
-		strings.Join(payload.Tags, ","),                 // tags
+	// No need to block our response to Mist; everything else in a goroutine
+	go func() {
+		insertDynStmt := `insert into "` + USER_END_TABLE_NAME + `"(
```
victorges, line 31: out of caution a recover
```suggestion
		defer func() {
			if rec := recover(); rec != nil {
				glog.Errorf("panic writing to analytics database err=%s payload=%v", rec, payload)
			}
		}()
		insertDynStmt := `insert into "` + USER_END_TABLE_NAME + `"(
```

victorges (approved):
LGTM

PR #1046, opened 2023-12-19
Increase Serf event size limit + truncate json names to keep messages small

thomshutt:


mjh1 (approved):

main.go
```diff
@@ -361,11 +361,11 @@ func handleClusterEvents(ctx context.Context, mapic mistapiconnector.IMac, bal b
 	}
 }
 
-func processClusterEvent(mapic mistapiconnector.IMac, bal balancer.Balancer, e serf.UserEvent) {
+func processClusterEvent(mapic mistapiconnector.IMac, bal balancer.Balancer, userEvent serf.UserEvent) {
 	go func() {
-		e, err := events.Unmarshal(e.Payload)
+		e, err := events.Unmarshal(userEvent.Payload)
```
pwilczynskiclearcode, line 366: We could use `Protocol Buffers`, `FlatBuffers` or golang native binary marshalling if we needed further compression

thomshutt: True! I think this hopefully buys us enough breathing room to explore alternative strategies

PR #1048, opened 2023-12-19
Implement JWT blocking

mjh1:


thomshutt (approved):

handlers/accesscontrol/access-control.go
```diff
@@ -94,6 +96,13 @@ func (ac *AccessControlHandlersCollection) IsAuthorized(playbackID string, paylo
 		acReq.AccessKey = accessKey
 		cacheKey = "accessKey_" + accessKey
 	} else if jwt != "" {
+		for _, blocked := range ac.blockedJWTs {
+			if jwt == blocked {
+				glog.Errorf("blocking JWT. playbackId=%v jwt=%v", acReq.Stream, jwt)
+				return false, fmt.Errorf("JWT is blocked: %s", jwt)
```
victorges, line 102: @mjh1 I _think_ we might not want to return an error here otherwise it will become a 500 to mist, which fails open. Not sure if someone else handles this error, but might be worth the double check

victorges: Fixed here https://github.com/livepeer/catalyst-api/pull/1051

PR #1052, opened 2023-12-20
Clean up token gating logs

mjh1:
👋 goodbye excessive logging


handlers/accesscontrol/access-control.go
```diff
@@ -129,13 +127,11 @@ func (ac *AccessControlHandlersCollection) GetPlaybackAccessControlInfo(playback
 	ac.mutex.RUnlock()
 
 	if isExpired(entry) {
-		glog.Infof("Cache expired for playbackId=%v cacheKey=%v", playbackID, cacheKey)
```
pwilczynskiclearcode, line 132: maybe we could log them on the verbose level or sth:
`glog.V(7).Infof("Cache expired for playbackId=%v cacheKey=%v", playbackID, cacheKey)`
in case we would need to debug caching

handlers/accesscontrol/access-control.go
```diff
@@ -69,7 +69,6 @@ func (ac *AccessControlHandlersCollection) HandleUserNew(ctx context.Context, pa
 	}
 
 	if playbackAccessControlAllowed {
-		glog.Infof("Playback access control allowed for playbackId=%v", playbackID)
```
thomshutt, line 72: Can be a separate PR, but this should be a metric

gioelecerati (approved):

PR #1063, opened 2023-12-27
ignore SPS errors when probing

emranemran:
When probing, we should ignore SPS errors as they indicate a slightly out-of-spec H264 stream. These files can
still be transcoded successfully so we should be less stringent here.


emranemran:
cc @thomshutt and @mjh1 for retroactive review once they're back.

leszko (approved):

mjh1 (commented):
lgtm

PR #1065, opened 2023-12-27
transcode/transmux: set MP4 generation limit

emranemran:
Generate MP4s only for the first 6hrs. Any input (e.g. recordings) that are longer will still get transcoded up to it's full duration but MP4s will be available for only the first 6hrs.


victorges (approved):
LGTM

PR #1073, opened 2024-01-04
Implement logger for Retryable http client

mjh1:
The default was to always produce logs which we don't generally need

### DIFF
--- a/log/httpLogger.go
+++ b/log/httpLogger.go
@@ -1,0 +1,39 @@
+package log
+
+import (
+	"github.com/golang/glog"
+	"github.com/hashicorp/go-retryablehttp"
+)
+
+var _ retryablehttp.LeveledLogger = retryableHTTPLogger{}
+
+type retryableHTTPLogger struct {
+}
+
+func NewRetryableHTTPLogger() retryablehttp.LeveledLogger {
+	return retryableHTTPLogger{}
+}
+
+func (r retryableHTTPLogger) Error(msg string, keysAndValues ...interface{}) {
+	if glog.V(3) {
+		LogNoRequestID(msg, keysAndValues...)
+	}
+}
+
+func (r retryableHTTPLogger) Warn(msg string, keysAndValues ...interface{}) {
+	if glog.V(4) {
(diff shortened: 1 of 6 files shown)


log/httpLogger.go
```diff
@@ -0,0 +1,39 @@
+package log
+
+import (
+	"github.com/golang/glog"
+	"github.com/hashicorp/go-retryablehttp"
+)
+
+var _ retryablehttp.LeveledLogger = retryableHTTPLogger{}
```
leszko, line 8: Why do we need this line? 🤔

leszko (approved):

log/httpLogger.go
```diff
@@ -0,0 +1,39 @@
...
+
+func (r retryableHTTPLogger) Info(msg string, keysAndValues ...interface{}) {
+	if glog.V(5) {
+		LogNoRequestID(msg, keysAndValues...)
+	}
+}
+
+func (r retryableHTTPLogger) Debug(msg string, keysAndValues ...interface{}) {
+	if glog.V(6) {
+		LogNoRequestID(msg, keysAndValues...)
```
pwilczynskiclearcode, line 37: `LogNoRequestID()` uses `kitlog` which supports [level-filtering](https://pkg.go.dev/github.com/go-kit/log/level) but unsure why we use this lib. `glog` seems much more widely used.

log/httpLogger.go
mjh1: It confirms at compile time that the interface is satisfied, but I've realised it's not needed now I've added the constructor as that returns the interface and performs the same check at compile time.

log/httpLogger.go
mjh1: yeah.. it does look odd, i'll see if there's a better way but glog didn't seem to have the same function signature with `msg string, keysAndValues ...interface{}`

log/httpLogger.go
leszko: Then, I suggest to remove it :)

leszko (approved):

PR #1077, opened 2024-01-08
Add retries for writing segments

mjh1:


handlers/ffmpeg/ffmpeg.go
```diff
@@ -73,7 +75,13 @@ func (h *HandlersCollection) NewFile() httprouter.Handle {
 		// and pass the one ffmpeg gives us to UploadToOSURL instead
 		targetURLBase := reg.ReplaceAllString(job.SegmentingTargetURL, "")
 
-		if err := clients.UploadToOSURL(targetURLBase, filename, body, config.SEGMENT_WRITE_TIMEOUT); err != nil {
+		if err := backoff.Retry(func() error {
+			err := clients.UploadToOSURL(targetURLBase, filename, body, config.SEGMENT_WRITE_TIMEOUT)
```
leszko, line 79: I'm not sure it will work, because `body` is `data io.Reader`, so I think that you cannot read from this twice.

leszko (changes requested):

handlers/ffmpeg/ffmpeg.go
mjh1: ah thanks good spot!

thomshutt: Let's also get a test in here, since we shouldn't be relying on manually catching this stuff

leszko (approved):

PR #1086, opened 2024-01-12
Rafal/audio sync issues

leszko:


leszko:
Closing, because apparently the mp4box does not solve the root issue for all scenarios.

PR #1088, opened 2024-01-15
Re-encode audio while concatenating TS files into MP4

leszko:
This solves the issue with audio/video sync.

Fix https://linear.app/livepeer/issue/PS-198/hello-team-we-have-received-reports-of-audio-sync-issues-in-the-final#comment-c1a08e6e

### DIFF
--- a/video/transmux.go
+++ b/video/transmux.go
@@ -197,7 +197,9 @@ func concatStreams(segmentList, outputTsFileName string) error {
 		"f":    "concat", // Use stream based concatenation (instead of file based concatenation)
 		"safe": "0"}).    // Must be 0 since relative paths to segments are used in segmentListTxtFileName
 		Output(outputTsFileName, ffmpeg.KwArgs{
-			"c": "copy", // Don't accidentally transcode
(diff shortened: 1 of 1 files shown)


video/transmux.go
```diff
@@ -197,7 +197,9 @@ func concatStreams(segmentList, outputTsFileName string) error {
 		"f":    "concat", // Use stream based concatenation (instead of file based concatenation)
 		"safe": "0"}).    // Must be 0 since relative paths to segments are used in segmentListTxtFileName
 		Output(outputTsFileName, ffmpeg.KwArgs{
-			"c": "copy", // Don't accidentally transcode
+			"c:v": "copy", // Don't accidentally transcode video
+			"c:a": "aac",  // Re-encode audio to avoid audio/video sync issues
+			"af":  "aresample=async=100",
```
thomshutt, line 202: Could we have a comment on this line too please?

thomshutt (approved):

video/transmux.go
leszko: Updated

mjh1 (approved):

PR #1110, opened 2024-01-30
Enrich Mist Prometheus Metrics

leszko:
Add `user_id` label to playback-related metrics. We can add more stream-related labels if needed.

Additional comments:
- Adding labels is done on-the-fly (if we see some performance issues, then we can include some caching)
- Mist metrics are exposed together with other metrics under `/metrics`
- Adding labels is done by a simple string replacement, I considered using some Prometheus metrics parsing, but it seemed to be more code and more works, so ended up with this solution

Related PRs:
- https://github.com/livepeer/catalyst/pull/835
- https://github.com/livepeer/livepeer-infra/pull/1931

fix https://linear.app/livepeer/issue/ENG-1581/whitelist-usersaccounts-so-they-never-raise-viewers-by-stream-alert

### DIFF
--- a/main.go
+++ b/main.go
@@ -87,6 +87,7 @@ func main() {
 	fs.StringVar(&cli.MistHost, "mist-host", "127.0.0.1", "Hostname of the Mist server")
 	fs.StringVar(&cli.MistUser, "mist-user", "", "username of MistServer")
 	fs.StringVar(&cli.MistPassword, "mist-password", "", "password of MistServer")
+	fs.StringVar(&cli.MistPrometheus, "mist-prometheus", "", "Mist path for the prometheus metrics endpoint")
 	fs.DurationVar(&cli.MistConnectTimeout, "mist-connect-timeout", 5*time.Minute, "Max time to wait attempting to connect to Mist server")
 	fs.StringVar(&cli.MistStreamSource, "mist-stream-source", "push://", "Stream source we should use for created Mist stream")
 	fs.StringVar(&cli.MistHardcodedBroadcasters, "mist-hardcoded-broadcasters", "", "Hardcoded broadcasters for use by MistProcLivepeer")
--- a/mapic/mistmetrics.go
+++ b/mapic/mistmetrics.go
@@ -3,21 +3,61 @@ package mistapiconnector
 import (
 	"fmt"
 	"github.com/golang/glog"
+	"github.com/julienschmidt/httprouter"
+	"io"
+	"net/http"
 	"strings"
 )
 
+func (mc *mac) MistMetricsHandler() httprouter.Handle {
+	return func(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+		mistMetrics, err := mc.queryMistMetrics()
+		if err != nil {
+			http.Error(w, fmt.Sprintf("error fetching Mist prometheus metrics: %s", err), http.StatusInternalServerError)
+			return
+		}
+
+		enrichedMistMetrics := mc.enrichMistMetrics(mistMetrics)
+		_, err = w.Write([]byte(enrichedMistMetrics))
+		if err != nil {
+			http.Error(w, fmt.Sprintf("error writing enriched metrics: %s", err), http.StatusInternalServerError)
+			return
+		}
+	}
+}
+
+func (mc *mac) queryMistMetrics() (string, error) {
+	mistMetricsURL := fmt.Sprintf("http://%s:%d/%s", mc.config.MistHost, mc.config.MistPort, mc.config.MistPrometheus)
+	resp, err := http.Get(mistMetricsURL)
+	if err != nil {
+		return "", err
+	}
+	defer resp.Body.Close()
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return "", err
+	}
+	return string(body), nil
+}
+
+// enrichMistMetrics adds additional labels to Mist streams
 func (mc *mac) enrichMistMetrics(metrics string) string {
 	res := strings.Builder{}
 	lines := strings.Split(metrics, "\n")
 	for i, line := range lines {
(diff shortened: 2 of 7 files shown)


mapic/mistmetrics.go
```diff
@@ -0,0 +1,92 @@
...
+
+func (mc *mac) enrichLabels(playbackID string) string {
+	res := mc.streamLabel(playbackID)
+	si, err := mc.getStreamInfo(playbackID)
+	if err != nil {
+		glog.Warning("could not enrich Mist metrics for stream=%s err=%v", playbackID, err)
+	} else if si == nil || si.stream == nil {
+		glog.Warning("could not enrich Mist metrics for stream=%s, cannot fetch stream info", playbackID)
+	} else {
+		res += fmt.Sprintf(`,user_id="%s"`, si.stream.UserID)
```
thomshutt, line 84: Nice, I didn't realise we actually had this info

thomshutt (approved):

main.go
```diff
@@ -87,6 +87,7 @@ func main() {
 	fs.StringVar(&cli.MistHost, "mist-host", "127.0.0.1", "Hostname of the Mist server")
 	fs.StringVar(&cli.MistUser, "mist-user", "", "username of MistServer")
 	fs.StringVar(&cli.MistPassword, "mist-password", "", "password of MistServer")
+	fs.StringVar(&cli.MistPrometheus, "mist-prometheus", "", "Mist path for the prometheus metrics endpoint")
```
pwilczynskiclearcode, line 90: You don't actually use `cli.MistPrometheus` to define the path. I think it's fine to have hardcoded `/mistmetrics` path. Probably no need to ever change it.

mapic/mistmetrics.go
```diff
@@ -0,0 +1,92 @@
...
+	match := mc.streamMetricsRe.FindStringSubmatch(line)
+	if len(match) > 1 {
+		return match[1], true
+	}
+	return "", false
+}
+
+func (mc *mac) enrichLabels(playbackID string) string {
+	res := mc.streamLabel(playbackID)
+	si, err := mc.getStreamInfo(playbackID)
```
pwilczynskiclearcode, line 78: does it take it from studio API? Does it always do a request or use some internal cache?

I'm just worried about too many repeating requests in case of multiple playbackIDs..

main.go
pwilczynskiclearcode: On the other hand why wouldn't we have it always enabled?

leszko: The metrics are always exposed from Catalyst API with the `/mistmetrics` endpoint. This is for the endpoint exposed from Mist. And it corresponds to this: https://github.com/livepeer/livepeer-infra/blob/40c743cae8f63421d7344032831dc0a9cb2ea01e/values/catalyst.values.yaml#L6

mapic/mistmetrics.go
leszko: This is cached in catalyst-api, so it won't make a network call to studio.

pwilczynskiclearcode:
Merging it with existing `/metrics` would simplify prometheus config as it wouldn't need to set it up to scrape this additional endpoint.

mapic/mistmetrics_test.go
```diff
@@ -0,0 +1,53 @@
...
+	inLines := strings.Split(mistMetrics, "\n")
+	resLines := strings.Split(res, "\n")
+	require.Equal(t, len(inLines), len(resLines))
+
+	expLines := []string{
+		`version{app="MistServer",version="729ddd4b42980d0124c72a46f13d8e0697293e94",release="Generic_x86_64"} 1`,
+		`mist_sessions{stream="video+077bh6xx5bx5tdua",user_id="abcdefgh-123456789",sessType="viewers"}1`,
+		`mist_latency{stream="video+077bh6xx5bx5tdua",user_id="abcdefgh-123456789",source="sin-prod-catalyst-3.lp-playback.studio"}1795`,
+		`mist_sessions{stream="video+51b13mqy7sgw520w",user_id="hgfedcba-987654321",sessType="viewers"}5`,
+		`mist_latency{stream="video+51b13mqy7sgw520w",user_id="hgfedcba-987654321",source="prg-prod-catalyst-0.lp-playback.studio"}1156`,
```
pwilczynskiclearcode, line 48: We'd need to either disable scraping mist metrics directly (we currently do it, right?) 
or add some suffix/prefix to enriched metrics.

leszko:
> Merging it with existing `/metrics` would simplify prometheus config as it wouldn't need to set it up to scrape this additional endpoint.

Yeah, the reason I didn't merge it together is that if, for some reason, this `/mistmetrics` does not perform well, then it's simpler to (temporarily) stop using it. Otherwise, if we have any issues with this, we would need to stop fetching all metrics.

mapic/mistmetrics_test.go
leszko: Yes, I think we should stop scaping mist and scrape this instead. I'll send a PR to `livepeer-infra`.

pwilczynskiclearcode:
I think we should aim to avoid blacklisting/whitelisting on the grafana-side... but rather have it on the studio-side.... so we'd need to pull another information from studio - `is_whitelisted` flag... But that's a longer story, I guess

pwilczynskiclearcode:
I think that having two endpoints might require being served at two different ports. I can't find a simple way to extract prometheus metrics from two endpoints on the single port.... not sure

leszko:
> I think that having two endpoints might require being served at two different ports. I can't find a simple way to extract prometheus metrics from two endpoints on the single port.... not sure

Ok, merged everything into `/metrics`

main.go
pwilczynskiclearcode, line 90: Should this be `fs.BoolVar` instead?

pwilczynskiclearcode (approved):

PR #1114, opened 2024-01-30
FLV pull

mjh1:


handlers/geolocation/geolocation.go
```diff
@@ -128,17 +141,44 @@ func (c *GeolocationHandlersCollection) HandleStreamSource(ctx context.Context,
 	dtscURL, err := c.Balancer.MistUtilLoadSource(context.Background(), payload.StreamName, latStr, lonStr)
 	if err != nil {
 		glog.Errorf("error querying mist for STREAM_SOURCE: %s", err)
+
+		url, err := c.getStreamPull(strings.TrimPrefix(payload.StreamName, "video+"))
+		if err != nil {
+			return "", err
+		}
+		if url != "" {
```
emranemran, line 149: should we check to make sure url has a `flv` here just to make sure no other type of input is attempted to be started?

emranemran:
Changes lgtm - just the comment from Victor about removing studio dependency.

handlers/geolocation/geolocation.go
mjh1: One thing Jaron mentioned was that if the url doesn't end `.flv` then we should change the protocol in the url to `flv://` so not sure we can check if it truly is flv ourselves? I think I'll implement this suffix check logic in a follow up PR

emranemran (approved):

PR #1118, opened 2024-01-31
access-control: added refresh_interval to gate & invalidate sessions to mapic

gioelecerati:


handlers/accesscontrol/access-control.go
```diff
@@ -79,12 +100,31 @@ func periodicCleanUpHitRecordCache() {
 	}()
 }
 
-func NewAccessControlHandlersCollection(cli config.Cli) *AccessControlHandlersCollection {
+func periodicRefreshIntervalCache(mapic mistapiconnector.IMac, accessControlCache map[string]map[string]*PlaybackAccessControlEntry) {
+	go func() {
+		for {
+			time.Sleep(time.Duration(5) * time.Second)
```
mjh1, line 106: you can just do `time.Sleep(5 * time.Second)`

handlers/accesscontrol/access-control.go
```diff
@@ -61,16 +62,36 @@ type HitRecords struct {
 	mux  sync.Mutex
 }
 
+type RefreshIntervalRecord struct {
+	RefreshInterval int32
+	LastRefresh     time.Time
+}
+
+type RefreshInterval struct {
+	data map[string]*RefreshIntervalRecord
```
mjh1, line 71: could you leave a comment of what the key is here, playbackID?

handlers/accesscontrol/access-control.go
```diff
@@ -61,16 +62,36 @@ type HitRecords struct {
...
+}
+
+type GateConfig struct {
+	MaxAge               int32 `json:"max_age"`
+	StaleWhileRevalidate int32 `json:"stale_while_revalidate"`
```
mjh1, line 76: could you change these to `time.Duration`?

handlers/accesscontrol/access-control.go
```diff
@@ -61,16 +62,36 @@ type HitRecords struct {
...
+	RateLimit            int32 `json:"rate_limit"`
+	RefreshInterval      int32 `json:"refresh_interval"`
```
mjh1, line 78: i think this can be `time.Duration` too

handlers/accesscontrol/access-control.go
```diff
@@ -236,6 +282,12 @@ func (ac *AccessControlHandlersCollection) cachePlaybackAccessControlInfo(playba
 		}
 	}
 
+	if refreshInterval > 0 {
+		if _, ok := refreshIntervalCache.data[playbackID]; !ok {
+			refreshIntervalCache.data[playbackID] = &RefreshIntervalRecord{RefreshInterval: refreshInterval, LastRefresh: time.Now()}
```
mjh1, line 287: do we need to lock when updating this map? since it's being read in the 5 second sleep loop in another routine

handlers/accesscontrol/access-control.go
```diff
@@ -79,12 +100,31 @@ func periodicCleanUpHitRecordCache() {
...
+			}
+		}
+	}()
+}
 
-	periodicCleanUpHitRecordCache()
+func NewAccessControlHandlersCollection(cli config.Cli, mapic mistapiconnector.IMac) *AccessControlHandlersCollection {
+	accessControlCache := make(map[string]map[string]*PlaybackAccessControlEntry)
+	periodicCleanUpRecordCache()
+	periodicRefreshIntervalCache(mapic, accessControlCache)
```
mjh1, line 124: maybe we should just make this function part of AccessControlHandlersCollection then we can just access the cache field in the normal way.
e.g `func (ac *AccessControlHandlersCollection) periodicRefreshIntervalCache(...`
then we can access `ac.cache`

handlers/accesscontrol/access-control.go
```diff
@@ -79,12 +100,31 @@ func periodicCleanUpHitRecordCache() {
...
+	go func() {
+		for {
+			time.Sleep(time.Duration(5) * time.Second)
+			for key := range refreshIntervalCache.data {
+				if time.Since(refreshIntervalCache.data[key].LastRefresh) > time.Duration(refreshIntervalCache.data[key].RefreshInterval)*time.Second {
+					refreshIntervalCache.data[key].LastRefresh = time.Now()
+					mapic.InvalidateAllSessions(key)
+					playbackIdCache := accessControlCache[key]
+					for cacheKey := range playbackIdCache {
+						delete(playbackIdCache, cacheKey)
```
mjh1, line 113: We need to lock with the mutex here i think before updating

handlers/accesscontrol/access-control.go
gioelecerati: Don't want to touch old acl code so close to trovo launch, can do it a followup pr, if that's fine to you

thomshutt (approved):

PR #1123, opened 2024-02-01
handlers: add headers as queryparam for pull sources

emranemran:


handlers/geolocation/geolocation.go
```diff
@@ -170,7 +170,17 @@ func (c *GeolocationHandlersCollection) getStreamPull(playbackID string) (string
 		return "", nil
 	}
 
-	return stream.Pull.Source, nil
+	if len(stream.Pull.Headers) == 0 {
+		return stream.Pull.Source, nil
+	}
```
victorges, line 175: btw, nil maps in go behave just like an empty map when reading (will just skip the loop below), so this check is optional

victorges (approved):
neat

PR #1126, opened 2024-02-02
clients: input-copy: add -analyzeduration to input probe logic

emranemran:
For recordings, the first segment can have missing or delayed audio and video tracks as the streamer begins the stream. Currently, the probe of an m3u8 file will fail as ffprobe attempts opening the first segment (e.g. 0.ts) and fails to determing codec information for either the audio or video track. This causes the InputVideo struct to get populated with incorrect info (e.g. width and height is set to 0) which causes downstream transcode issues. Setting -analyzeduration 15M allows ffprobe to inspect up to 15s of audio and video tracks to determine the track information.

### DIFF
--- a/clients/input_copy.go
+++ b/clients/input_copy.go
@@ -64,7 +64,7 @@ func (s *InputCopy) CopyInputToS3(requestID string, inputFile, osTransferURL *ur
 	}
 
 	log.Log(requestID, "starting probe", "source", inputFile.String(), "dest", osTransferURL.String())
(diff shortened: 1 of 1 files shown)


emranemran:
FYI: we already do this for output files: https://github.com/livepeer/catalyst-api/blob/0a1354d1bb9f2c02589ca962c4abe8139ba5ce2e/video/profiles.go#L239

thomshutt (approved):

PR #1135, opened 2024-02-09
Handle large a/v offset in first segment of recordings.

emranemran:
transcode: drop first segment if a/v tracks have a large offset

For HLS input (recordings), sometimes the first segment can experience a
    large delta between the audio and video track's start time (in seconds).
    This is mainly because when a streamer begins the stream, the user's
    hardware config and OBS settings might cause the video track to start
    later than the audio (or vice versa). These segments can be transcoded
    but muxing them into an MP4 will fail. For these cases, the first segment
    should be dropped from the transcode payload.

### DIFF
--- a/pipeline/coordinator.go
+++ b/pipeline/coordinator.go
@@ -109,19 +109,21 @@ type JobInfo struct {
 	SegmentingDone     time.Time
 	TranscodingDone    time.Time
 
-	sourceBytes        int64
-	sourceSegments     int
-	sourceDurationMs   int64
-	sourceCodecVideo   string
-	sourceCodecAudio   string
-	sourceWidth        int64
-	sourceHeight       int64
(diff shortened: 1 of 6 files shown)


thomshutt (approved):
If it works then LGTM, but do we have any sense why this offset would cause problems when muxing to MP4?

Also it looks like the Cucumber tests are failing

emranemran:
> If it works then LGTM, but do we have any sense why this offset would cause problems when muxing to MP4?
> 
> Also it looks like the Cucumber tests are failing

Because MP4s need a/v starting PTS to be very close to one another otherwise the headers cannot be generated correctly. The 1s is arbitrary and we may update in the future. The offending recordings typically have 8s+ offsets between a/v tracks and the video typically starts later.

PR #1143, opened 2024-02-14
Forbid starting stream when it's suspended or deleted

leszko:
fix https://linear.app/livepeer/issue/PS-227/qiang-we-are-testing-new-camera-firmware-and-observing-strange-issue

### DIFF
--- a/mapic/mistapiconnector_app.go
+++ b/mapic/mistapiconnector_app.go
@@ -274,6 +274,10 @@ func (mc *mac) handlePushRewrite(ctx context.Context, payload *misttriggers.Push
 	} else {
 		glog.Errorf("Shouldn't happen streamID=%s", stream.ID)
 	}
+	if stream.Deleted || stream.Suspended {
+		// Do not allow to start deleted or suspended streams
+		return "", nil
+	}
 	glog.Infof("Responded with '%s'", responseName)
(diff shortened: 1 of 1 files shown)


thomshutt:
@leszko I'd tried to do this in https://github.com/livepeer/catalyst-api/pull/1133 - should we get rid of that?

leszko:
> @leszko I'd tried to do this in #1133 - should we get rid of that / any idea why it's not working here?

Not sure TBH. What was the purpose of your change? When I tried in the box, even without your change and my change, I couldn't play the pull stream when it was suspended.

thomshutt:
@leszko my change was to stop terminated Pull streams coming back to life when a new viewer initiated a playback

leszko:
> @leszko my change was to stop terminated Pull streams coming back to life when a new viewer initiated a playback

Hmm, I tried in the box, and even without your change, a new playback didn't work after the stream was suspended... 🤔

thomshutt:
I think we probably need both though - yours covers the autopush and mine covers new viewers

leszko:
> I think we probably need both though - yours covers the autopush and mine covers new viewers

Yeah, that's what I think as well. One is ingest, one is watching.

thomshutt (approved):

PR #1147, opened 2024-02-19
Add Analytics Log endpoint

leszko:
Add `/analytics/log` endpoint which will be used to pass the analytics from player => analytics pipeline.

Design Doc: https://www.notion.so/livepeer/Studio-2-0-Realtime-Viewership-Backend-2511fd91b0854d09958eed1fa548f447

fix: https://linear.app/livepeer/issue/ENG-1649/implement-http-endpoint-for-periodic-events

Note that:
- Currently, the endpoint does not do anything other than logging, but I'd like to get it into staging, because:
  - It will help with the Player integration
  - It will help with checking if NGINX is passing geo data correctly
  - It will help with checking if the endpoint is exposed correctly with NGINX
- The features is behind the flag `analytics`
  - I plan to have it enabled in staging, but disabled for now in prod
  - The flag is string not bool, because bool is broken with passing from Mist => Catalyst API (the hack with inverted bool is ugly, we should not use it)
  
CC: @0xcadams

### DIFF
--- a/api/http.go
+++ b/api/http.go
@@ -66,7 +66,7 @@ func NewCatalystAPIRouter(cli config.Cli, vodEngine *pipeline.Coordinator, bal b
 	analyticsApiHandlers := &handlers.AnalyticsHandlersCollection{}
 
 	router.GET("/ok", withLogging(catalystApiHandlers.Ok()))
-	if cli.EnableAnalytics {
+	if cli.EnableAnalytics == "true" || cli.EnableAnalytics == "enabled" {
 		router.POST("/analytics/log", analyticsApiHandlers.Log())
 	}
 
--- a/api/http.go
+++ b/api/http.go
@@ -63,8 +63,12 @@ func NewCatalystAPIRouter(cli config.Cli, vodEngine *pipeline.Coordinator, bal b
 		Config:   cli,
 		Lapi:     lapi,
 	}
+	analyticsApiHandlers := &handlers.AnalyticsHandlersCollection{}
 
 	router.GET("/ok", withLogging(catalystApiHandlers.Ok()))
+	if cli.EnableAnalytics {
+		router.POST("/analytics/log", analyticsApiHandlers.Log())
+	}
 
 	// Playback endpoint
 	playback := middleware.LogAndMetrics(metrics.Metrics.PlaybackRequestDurationSec)(
--- a/api/http.go
+++ b/api/http.go
@@ -63,10 +63,10 @@ func NewCatalystAPIRouter(cli config.Cli, vodEngine *pipeline.Coordinator, bal b
 		Config:   cli,
 		Lapi:     lapi,
 	}
-	analyticsApiHandlers := &handlers.AnalyticsHandlersCollection{}
 
 	router.GET("/ok", withLogging(catalystApiHandlers.Ok()))
 	if cli.EnableAnalytics == "true" || cli.EnableAnalytics == "enabled" {
+		analyticsApiHandlers := handlers.NewAnalyticsHandlersCollection()
 		router.POST("/analytics/log", analyticsApiHandlers.Log())
 	}
 
--- a/handlers/analytics.go
+++ b/handlers/analytics.go
@@ -1,0 +1,130 @@
+package handlers
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/golang/glog"
+	"github.com/julienschmidt/httprouter"
+	"github.com/livepeer/catalyst-api/errors"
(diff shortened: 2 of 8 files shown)


api/http.go
```diff
@@ -62,8 +62,12 @@ func NewCatalystAPIRouter(cli config.Cli, vodEngine *pipeline.Coordinator, bal b
 		Config:   cli,
 		Lapi:     lapi,
 	}
+	analyticsApiHandlers := &handlers.AnalyticsHandlersCollection{}
```
mjh1, line 65: maybe make a constructor function for this as I often find we end up needing some logic when setting up a struct like this and it can be a pain in the future updating everywhere.

handlers/analytics.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/mmcloughlin/geohash"
+	"github.com/xeipuuv/gojsonschema"
+	"io"
+	"net/http"
+	"strconv"
+)
+
+const GEO_HASH_PRECISION = 3
+
+type AnalyticsHandler struct {
```
mjh1, line 18: Is this struct used anywhere or will we need it in a further PR?

api/http.go
thomshutt, line 65: Could this go down inside the conditional too?

handlers/analytics.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+	if len(missingHeader) > 0 {
+		return &res, fmt.Errorf("missing geo headers: %v", missingHeader)
+	}
+
+	return &res, nil
+}
+
+func getOrAddMissing(key string, headers http.Header, missingHeaders []string) (string, []string) {
+	if val, ok := headers[key]; ok {
```
mjh1, line 105: It's probably better to use the helper function `headers.Get(key string)` and check if val is `""`. It just returns the first value so then no need to do `val[0]`

handlers/analytics.go
```diff
@@ -0,0 +1,130 @@
...
+	payload, err := io.ReadAll(r.Body)
+	if err != nil {
+		return nil, err
+	}
+	result, err := schema.Validate(gojsonschema.NewBytesLoader(payload))
+	if err != nil {
+		return nil, err
+	}
+	if !result.Valid() {
+		return nil, err
```
thomshutt, line 122: This should be a new error - I think at the moment it will always be nil

handlers/analytics.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+	var log AnalyticsLog
+	if err := json.Unmarshal(payload, &log); err != nil {
+		return nil, err
```
thomshutt, line 126: We should wrap these errors with a `fmt.Errorf("<what we were trying to do>: %w", err)` so it's clear where they came from

thomshutt (approved):
LGTM - left a few comments about small changes but feel free to merge without a re-review

handlers/analytics.go
```diff
@@ -0,0 +1,130 @@
...
+		if err != nil {
+			glog.Warning("error parsing geo info from analytics log request header, err=%v", err)
+		}
+
+		// TODO: ENG-1650, Process analytics data and remove logging
+		glog.Info("Processing analytics log: log=%v, geo=%v", log, geo)
+	}
+}
+
+func parseAnalyticsGeo(r *http.Request) (*AnalyticsGeo, error) {
```
mjh1, line 75: maybe AnalyticsGeo could just be a value rather than a pointer, it's just a simple struct and not sure there's a need for it to be a pointer?

mjh1 (commented):
pretty much lgtm, just some nits

api/http.go
leszko: Good spot, updated in 966e638a1cc48551294ffc960a8f82364098cb9c

handlers/analytics.go
leszko: Updated in 966e638a1cc48551294ffc960a8f82364098cb9c

handlers/analytics.go
leszko: Updated in 966e638a1cc48551294ffc960a8f82364098cb9c

api/http.go
leszko: Updated in 966e638a1cc48551294ffc960a8f82364098cb9c

handlers/analytics.go
leszko: leftover, removed.

handlers/analytics.go
leszko: Updated in 966e638a1cc48551294ffc960a8f82364098cb9c

PR #1149, opened 2024-02-22
geolocation: Allow overriding lat/lon with qs

victorges:
This is to allow overriding the lat/lon used for load balancing with custom values.

It is needed for the API service to be able to trigger the pull stream to start in a location
close to the pull source, instead of close to where the API is.

The headers themselves are not overridable, since our nginx configuration uses `proxy_set_header`
which forces the value even if one is already present.

I thought about adding some kind of authorization for the override, but it feels like this context
there's no problem if a user tried to override this. They will just get redirected to a worse node. If
they are malicious, they could just hit that (and all other) nodes directly anyway.

Usage: https://github.com/livepeer/studio/pull/2073

### DIFF
--- a/handlers/geolocation/geolocation.go
+++ b/handlers/geolocation/geolocation.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"net/url"
 	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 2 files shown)


handlers/geolocation/geolocation.go
```diff
@@ -36,8 +36,13 @@ func (c *GeolocationHandlersCollection) RedirectHandler() httprouter.Handle {
 		host := r.Host
 		pathType, prefix, playbackID, pathTmpl := parsePlaybackID(r.URL.Path)
 		redirectPrefixes := c.Config.RedirectPrefixes
-		lat := r.Header.Get("X-Latitude")
-		lon := r.Header.Get("X-Longitude")
+
+		query := r.URL.Query()
+		lat, lon := query.Get("lat"), query.Get("lon")
```
emranemran, line 41: nit: since these are query params, do we need some error checking to make sure they're in range? i.e. 
```
The numbers are in decimal degrees format and range from -90 to 90 for latitude and -180 to 180 for longitude
```

emranemran (approved):

pwilczynskiclearcode (approved):

handlers/geolocation/geolocation.go
victorges: Done! Since these are not critical inputs (same reason for not adding some security layer), I just ignore the values if they are invalid (incl from the headers). Updated PR

PR #1156, opened 2024-03-04
Add logging and redirect for /analytics/log requests

leszko:
Redirect: We want to redirect `GET /analytics/log` from `https://playback.livepeer.monster/analytics/log` => `https://mdw-staging-staging-catalyst-0.livepeer.monster/analytics/log`, so that player sticks to one catalyst node while sending analytics logs for assets (for stream it's already covered by the playback redirects).

Logging: I'm actually not sure if we should log every analytics log request. A Catalyst node currently can serve ~2.5k viewers, which means that during the peak, we'll log additional `500` logs / second (~2.5k req each 5s). It seems fine, but I'd like to double check with @pwilczynskiclearcode if that's not excessive.

fix https://linear.app/livepeer/issue/ENG-1712/fix-redirects-and-cors

### DIFF
--- a/api/http.go
+++ b/api/http.go
@@ -67,7 +67,8 @@ func NewCatalystAPIRouter(cli config.Cli, vodEngine *pipeline.Coordinator, bal b
 	router.GET("/ok", withLogging(catalystApiHandlers.Ok()))
 	if cli.EnableAnalytics == "true" || cli.EnableAnalytics == "enabled" {
 		analyticsApiHandlers := handlers.NewAnalyticsHandlersCollection(mapic, lapi, cli.AnalyticsMetricsURL, cli.NodeName)
-		router.POST("/analytics/log", withCORS(analyticsApiHandlers.Log()))
+		router.POST("/analytics/log", withLogging(withCORS(analyticsApiHandlers.Log())))
+		router.GET("/analytics/log", withLogging(withCORS(geoHandlers.RedirectHandler())))
 	}
 
 	// Playback endpoint


pwilczynskiclearcode:
Logging? You mean ingress-nginx logs?
We could disable logging `/analytics/log` by adding it into [the list of exclusions](https://github.com/livepeer/livepeer-infra/blob/master/values/ingress-nginx.values.yaml#L24)

pwilczynskiclearcode:
The player is georouted when connecting to `/hls` or `/webrtc` anyway. So maybe it could reuse the georouted URL (domain) for `/analytics/log` as well? Otherwise we geolocate twice for stream and analytics separately.

api/http.go
```diff
@@ -67,7 +67,8 @@ func NewCatalystAPIRouter(cli config.Cli, vodEngine *pipeline.Coordinator, bal b
 	router.GET("/ok", withLogging(catalystApiHandlers.Ok()))
 	if cli.EnableAnalytics == "true" || cli.EnableAnalytics == "enabled" {
 		analyticsApiHandlers := handlers.NewAnalyticsHandlersCollection(mapic, lapi, cli.AnalyticsMetricsURL, cli.NodeName)
-		router.POST("/analytics/log", withCORS(analyticsApiHandlers.Log()))
+		router.POST("/analytics/log", withLogging(withCORS(analyticsApiHandlers.Log())))
+		router.GET("/analytics/log", withLogging(withCORS(geoHandlers.RedirectHandler())))
```
pwilczynskiclearcode, line 71: Could you extend [the tests](https://github.com/livepeer/catalyst-api/blob/main/handlers/analytics_test.go#L145-L162) so they check CORS headers and GET request as well?

leszko:
> Logging? You mean ingress-nginx logs? We could disable logging `/analytics/log` by adding it into [the list of exclusions](https://github.com/livepeer/livepeer-infra/blob/master/values/ingress-nginx.values.yaml#L24)
> 
> We still have some insights using [`nginx_ingress_controller_requests`](https://eu-metrics-monitoring.livepeer.live/grafana/goto/RekLVIASR?orgId=1) prometheus metric so yes, we could get rid of the access logs.

Discussed offline.

api/http.go
leszko: Hmm, CORS and redirect kind-of not belong to `analytics.go`, because they are added from outside (in `http.go`). I see that for other cases we don't test it here, only in cucumber: https://github.com/livepeer/catalyst-api/blob/10b16410706c875ff220b3dfe4e1a6524e3b05c4/test/features/playback.feature#L51

We could add something similar, but technically it's already tested, so probably to safe to leave it as it is.

thomshutt (approved):

PR #1157, opened 2024-03-05
Send analytics logs to Kafka

leszko:
Changes:
- Send analytics logs to Kafka (instead of  Prometheus / Victoria Metrics)
- Add more fields to Kafka Event

Fix https://linear.app/livepeer/issue/ENG-1702/send-raw-analytics-logs-to-kafka


api/http.go
```diff
@@ -66,9 +67,13 @@ func NewCatalystAPIRouter(cli config.Cli, vodEngine *pipeline.Coordinator, bal b
 
 	router.GET("/ok", withLogging(catalystApiHandlers.Ok()))
 	if cli.EnableAnalytics == "true" || cli.EnableAnalytics == "enabled" {
-		analyticsApiHandlers := handlers.NewAnalyticsHandlersCollection(mapic, lapi, cli.AnalyticsMetricsURL, cli.NodeName)
-		router.POST("/analytics/log", withLogging(withCORS(analyticsApiHandlers.Log())))
-		router.GET("/analytics/log", withLogging(withCORS(geoHandlers.RedirectHandler())))
+		analyticsApiHandlers, err := handlers.NewAnalyticsHandlersCollection(mapic, lapi, cli.KafkaBootstrapServers, cli.KafkaUser, cli.KafkaPassword, cli.KafkaTopic)
+		if err != nil {
+			glog.Errorf("failed to configure analytics handlers, analytics endpoint won't be enabled, err=%v", err)
```
victorges, line 72: I believe startup should fail if we get an error here. Better to block a deploy than never realize we broke our data pipeline.

handlers/analytics.go
```diff
@@ -35,34 +37,60 @@ type AnalyticsLog struct {
...
 	extFetcher   analytics.IExternalDataFetcher
 	logProcessor analytics.ILogProcessor
 }
 
-func NewAnalyticsHandlersCollection(streamCache mistapiconnector.IStreamCache, lapi *api.Client, metricsURL string, host string) AnalyticsHandlersCollection {
+func NewAnalyticsHandlersCollection(streamCache mistapiconnector.IStreamCache, lapi *api.Client, kafkaBootstrapServers, kafkaUser, kafkaPassword, kafkaTopic string) (AnalyticsHandlersCollection, error) {
+	lp, err := analytics.NewLogProcessor(kafkaBootstrapServers, kafkaUser, kafkaPassword, kafkaTopic)
+	if err != nil {
+		return AnalyticsHandlersCollection{}, err
+	}
```
victorges, line 89: **nit:** WDYT of receiving only the `analytics.ILogProcessor` as an input to this constructor, instead of the args for creating the processor? Will keep the interface much cleaner IMO and consistent with how the `lapi` is handled. 

The specific Kafka stuff would be initialized on the place that creates this instead (plus more explicit error handling) and we can deal with only the interface here.

main.go
```diff
@@ -122,7 +122,10 @@ func main() {
 	fs.IntVar(&cli.SerfEventBuffer, "serf-event-buffer", 100000, "Size of serf 'recent event' buffer, outside of which things are dropped")
 	fs.IntVar(&cli.SerfMaxQueueDepth, "serf-max-queue-depth", 100000, "Size of Serf queue, outside of which things are dropped")
 	fs.StringVar(&cli.EnableAnalytics, "analytics", "disabled", "Enables analytics API: enabled or disabled")
-	fs.StringVar(&cli.AnalyticsMetricsURL, "analytics-metrics-url", "", "URL of Prometheus DB to store analytics data")
+	fs.StringVar(&cli.KafkaBootstrapServers, "kafka-bootstrap-servers", "", "URL of Kafka Bootstrap Servers")
+	fs.StringVar(&cli.KafkaUser, "kafka-user", "", "Kafka Username")
+	fs.StringVar(&cli.KafkaPassword, "kafka-password", "", "Kafka Password")
+	fs.StringVar(&cli.KafkaTopic, "kafka-topic", "", "Kafka Topic used to send analytics logs")
```
victorges, line 128: Should this be more specific in case we have other topics in the future? e.g.
```suggestion
	fs.StringVar(&cli.AnalyticsKafkaTopic, "analytics-kafka-topic", "", "Kafka Topic used to send analytics logs")
```

victorges (approved):
LGTM

api/http.go
leszko: ok, changed to `Fatalf()` which will stop catalyst-api.

handlers/analytics.go
leszko: I like it :) Updated.

handlers/analytics/log_processor.go
```diff
@@ -1,212 +1,162 @@
...
-	for _, mv := range v {
-		if mv.count > 0 {
-			count += 1
-			sumRebufferRate += mv.sumRebufferRate / float64(mv.count)
+	var msgs []kafka.Message
+	for _, d := range p.logs {
+		key := []byte(d.SessionID)
+		value, err := json.Marshal(d)
+		if err != nil {
+			glog.Errorf("invalid analytics log event, cannot sent to Kafka, err=%v", err)
```
thomshutt, line 148: Can we get an identifying field in here too? e.g SessionID

handlers/analytics/log_processor.go
```diff
@@ -1,212 +1,162 @@
...
-		sumRebufferRate: mv.sumRebufferRate + rebufferRate,
-		sumErrorRatio:   mv.sumErrorRatio + errorRatio,
-	}
+	p.logs = append(p.logs, d)
 }
 
-func (p *LogProcessor) sendMetrics() {
+func (p *LogProcessor) sendEvents() {
 	if len(p.logs) > 0 {
 		glog.Info("sending analytics logs")
```
thomshutt, line 137: Let's get the len into this message

handlers/analytics/log_processor.go
```diff
@@ -1,212 +1,162 @@
...
-	resp, err := client.Do(req)
+	err := p.writer.WriteMessages(context.Background(), msgs...)
 	if err != nil {
-		return err
-	}
-	defer resp.Body.Close()
-
-	if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusNoContent {
-		return fmt.Errorf("non-OK status code: %d", resp.StatusCode)
+		glog.Errorf("error while sending analytics log to Kafka, err=%v", err)
```
thomshutt, line 160: Doesn't have to be in this PR, but this should probably be a metric that we can alert on

thomshutt (approved):

main.go
```diff
@@ -122,7 +122,10 @@ func main() {
 	fs.IntVar(&cli.SerfEventBuffer, "serf-event-buffer", 100000, "Size of serf 'recent event' buffer, outside of which things are dropped")
 	fs.IntVar(&cli.SerfMaxQueueDepth, "serf-max-queue-depth", 100000, "Size of Serf queue, outside of which things are dropped")
 	fs.StringVar(&cli.EnableAnalytics, "analytics", "disabled", "Enables analytics API: enabled or disabled")
-	fs.StringVar(&cli.AnalyticsMetricsURL, "analytics-metrics-url", "", "URL of Prometheus DB to store analytics data")
+	fs.StringVar(&cli.KafkaBootstrapServers, "kafka-bootstrap-servers", "", "URL of Kafka Bootstrap Servers")
+	fs.StringVar(&cli.KafkaUser, "kafka-user", "", "Kafka Username")
+	fs.StringVar(&cli.KafkaPassword, "kafka-password", "", "Kafka Password")
+	fs.StringVar(&cli.AnalyticsKafkaTopic, "analytics-kafka-topic", "", "Kafka Topic used to send analytics logs")
```
pwilczynskiclearcode, line 128: If we want to use just a single kafka topic we should partition it by at least `EventType`+`CreatorID`.

leszko: Good point. I guess it should be `session_id`, not creator_id, but yeat. Our initial idea with @ecmulli was to partition it only by `session_id`. This should be enough to sink data into Clickhouse. And if we plan to end up sending them to the same DB table, then do you see any value in partitioning (also) by `event.type`?

leszko: For the context, for now we'll have 2 event types: `heartbeat` and `error`.

handlers/analytics/log_processor.go
leszko: Added sessionID

handlers/analytics/log_processor.go
leszko: Ok, created an XS task in Linear: https://linear.app/livepeer/issue/ENG-1730/add-metrics-and-alerting

main.go
pwilczynskiclearcode: I guess there will be some service reading it from kafka and writing into Clickhouse. If it gets stuck it'd better if it got stuck (because of e.g. some parsing issue) only for one type of event. Also some super-huge number of error messages shouldn't block processing heartbeats and the opposite.

leszko: Yeah, it makes sense to me.

leszko: Updated to use session_id+event_type as the kafka key.

handlers/analytics/log_processor.go
```diff
@@ -1,212 +1,169 @@
...
-func (p *LogProcessor) toRebufferRatioMetric(k labelsKey, v map[string]metricValue, nowMs int64) string {
-	var count int
-	var sumRebufferRate float64
-	for _, mv := range v {
-		if mv.count > 0 {
-			count += 1
-			sumRebufferRate += mv.sumRebufferRate / float64(mv.count)
+	var msgs []kafka.Message
+	for _, d := range p.logs {
+		key, err := json.Marshal(KafkaKey{SessionID: d.SessionID, EventType: d.EventType})
```
pwilczynskiclearcode, line 148: optimization: 
no need to send JSON as a key. it could be simply:
```go
key := []byte(d.SessionID + d.EventType)
```

leszko: Hmm, currently we set the JSON schema for the key, so I guess this won't work. We could consider not having the schema and just using the concatenated string instead.

pwilczynskiclearcode: nvm then. No need for preoprimizations

PR #1166, opened 2024-03-07
Fix empty creatorID

leszko:


victorges (approved):
Ops I think I cause this haha

PR #1174, opened 2024-03-12
Add redirect for /analytics/log

leszko:
We want to redirect `GET https://playback.livepeer.monster/analytics/log` => `https://mdw-staging-staging-catalyst-0.livepeer.monster/analytics/log`

The context for this is this [Discord thread](https://discord.com/channels/423160867534929930/1200457179015155823/1212467699997745202); in short: when sending analytics logs from player=>catalyst, we want to stick to the same node (mainly for the optimization reasons), which works this way:
- streams => player GET playback `https://playback.livepeer.monster/hls/<playback-id>` which gets redirected to `https://mdw-staging-staging-catalyst-0.livepeer.monster/hls/<playback-id>`; the same node `mdw-staging-staging-catalyst-0.livepeer.monster` is used to send analytics logs
- assets => here we don't have a playback, so we need to somehow resolve the node address and stick to it; the idea is that player uses the same mechanism and redirects on `GET https://playback.livepeer.monster/analytics/log` => `https://mdw-staging-staging-catalyst-0.livepeer.monster/analytics/log`, then it sticks to one node `mdw-staging-staging-catalyst-0.livepeer.monster`

fix https://linear.app/livepeer/issue/ENG-1712/fix-redirects

### DIFF
--- a/handlers/geolocation/geolocation.go
+++ b/handlers/geolocation/geolocation.go
@@ -128,6 +128,15 @@ func (c *GeolocationHandlersCollection) RedirectHandler() httprouter.Handle {
 	}
 }
 
+// RedirectConstPathHandler redirects const path into the self catalyst node
+func (c *GeolocationHandlersCollection) RedirectConstPathHandler() httprouter.Handle {
+	return func(w http.ResponseWriter, r *http.Request, params httprouter.Params) {
+		rURL := fmt.Sprintf("%s://%s%s", protocol(r), c.Config.NodeName, r.URL.Path)
+		glog.V(6).Infof("generated redirect url=%s", rURL)
+		http.Redirect(w, r, rURL, http.StatusTemporaryRedirect)
+	}
+}
+
 // respond to a STREAM_SOURCE request from Mist
 func (c *GeolocationHandlersCollection) HandleStreamSource(ctx context.Context, payload *misttriggers.StreamSourcePayload) (string, error) {
 	lat := c.Config.NodeLatitude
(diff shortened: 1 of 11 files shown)


handlers/geolocation/geolocation.go
```diff
@@ -128,6 +128,15 @@ func (c *GeolocationHandlersCollection) RedirectHandler() httprouter.Handle {
 	}
 }
 
+// RedirectConstPathHandler redirects const path into the self catalyst node
+func (c *GeolocationHandlersCollection) RedirectConstPathHandler() httprouter.Handle {
```
thomshutt, line 132: Is this redirecting to itself? I'm not clear on what redirect issue this is fixing but could it just not redirect and handle the request?

leszko: Ahh, yeah that's confusing. I've updated the description. PTAL.

Technically, yes, we don't need any redirect, but then each analytics log for the same viewing session may be sent to a different catalyst node. We don't want this because of the performance (some playback-related data is cached in the catalyst node).

thomshutt: Gotcha, makes sense now thanks!

thomshutt (approved):

handlers/geolocation/geolocation.go
pwilczynskiclearcode: @leszko just might be worth to document it here, in the code.

leszko: Updated in 4bbf94c9b6036dd5edfa8427d7d36d677c6a759f

PR #1178, opened 2024-03-14
Add additional User Agent fields to Kafka Event

leszko:
- Remove `cpu`
- Add `device_model` and `device_brand`

fix https://linear.app/livepeer/issue/ENG-1722/additional-fields-to-kafka-event-cpu-device-brand-device-model

### DIFF
--- a/handlers/analytics.go
+++ b/handlers/analytics.go
@@ -4,6 +4,13 @@ import (
 	"crypto/sha256"
 	"encoding/json"
 	"fmt"
+	"github.com/ua-parser/uap-go/uaparser"
+	"io"
+	"net"
+	"net/http"
+	"strconv"
+	"time"
+
 	"github.com/golang/glog"
 	"github.com/julienschmidt/httprouter"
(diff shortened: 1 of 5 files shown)


thomshutt (approved):

handlers/analytics.go
```diff
@@ -191,8 +195,10 @@ func getOrAddMissing(key string, headers http.Header, missingHeaders []string) (
 	return "", missingHeaders
 }
 
-func toAnalyticsData(log *AnalyticsLog, geo AnalyticsGeo, extData analytics.ExternalData) []analytics.LogData {
+func (c *AnalyticsHandlersCollection) toAnalyticsData(log *AnalyticsLog, geo AnalyticsGeo, extData analytics.ExternalData) []analytics.LogData {
 	ua := useragent.Parse(log.UserAgent)
```
thomshutt, line 199: Do they both contain different fields (i.e so we can't just completely switch to the new parser)?

leszko: Unfortunately, they provide different fields for us 😠

PR #1183, opened 2024-03-18
Split "page_url" into "domain", "path", and "params"

leszko:
Fix https://linear.app/livepeer/issue/ENG-1810/split-page-url-into-domain-path-and-params


handlers/analytics/log_processor.go
```diff
@@ -58,7 +58,9 @@ type LogData struct {
 	PlaybackID            string       `json:"playback_id"`
 	ViewerHash            string       `json:"viewer_hash"`
 	Protocol              string       `json:"protocol"`
-	PageURL               string       `json:"page_url"`
+	Domain                string       `json:"domain"`
+	Path                  string       `json:"path"`
+	Params                string       `json:"params"`
```
0xcadams, line 63: I'd like to see `Hash` here as well - it likely is the least important or interesting value in a URL, but some apps use it for routing and it would be a field we shouldn't drop: https://developer.mozilla.org/en-US/docs/Web/API/URL/hash

handlers/analytics.go
```diff
@@ -276,5 +280,5 @@ func hashViewer(log *AnalyticsLog, geo AnalyticsGeo) string {
 		return log.UID
 	}
 	// If user didn't define the unique viewer ID, then we hash IP and user agent data
```
0xcadams, line 282: Nit: could we update comment here?

0xcadams:
```suggestion
	// If user didn't define the unique viewer ID, then we hash domain, IP and user agent data
```

0xcadams (approved):
LGTM, some small changes requested

ecmulli (approved):

ecmulli:
@leszko 
I've added three additional columns to the `viewership_events` table. Once we have events, we'll need to create a new Clickpipe that writes into the same table. 

1. Pause the existing one
2. Create a new clickpipe 
3. Upload data to "Existing Table"
4. make sure the new fields are mapped to the columns. They are named the same so it should be trivial. 

I can do this tomorrow but wrote these down in case you want to give it a try.

handlers/analytics/log_processor.go
leszko: Added

PR #1187, opened 2024-03-21
Add lock to stream pull to prevent multiple nodes starting pulling for the same stream

leszko:
Currently, when multiple playbacks are started for the same stream, different catalyst nodes start to pull the stream separately which results in a number of issues (like multiple same sessions created for the same stream, broadcaster fails because it's confused by the same `playbackId`/`manifestId`).

To solve it, Studio acts like a distributed lock with the lease timeout. In result what happens is that when the playback is started concurrently, then one of them will fail and need to retry until the stream is replicated between Mist nodes.

Note that we could consider some other distributed lock mechanism than Studio DB, e.g., we can consider replacing Serf with some other distributed in-memory DB which would act as a service discovery, distributed cache, etc. Some options are Redis, Hazelcast, Consul. 

Related PRs:
- https://github.com/livepeer/go-api-client/pull/61
- https://github.com/livepeer/studio/pull/2108

fix https://linear.app/livepeer/issue/PS-407/duplicate-streams

### DIFF
--- a/handlers/geolocation/geolocation.go
+++ b/handlers/geolocation/geolocation.go
@@ -223,7 +223,7 @@ func (c *GeolocationHandlersCollection) getStreamPull(playbackID string) (string
 	}
 
 	if err := c.Lapi.LockPull(stream.ID, lockPullLeaseTimeout); err != nil {
-		return "", fmt.Errorf("failed to lock pull, err=%v", err)
+		return "", errLockPull
 	}
 
 	if len(stream.Pull.Headers) == 0 {
--- a/handlers/geolocation/geolocation.go
+++ b/handlers/geolocation/geolocation.go
@@ -175,7 +175,7 @@ func (c *GeolocationHandlersCollection) HandleStreamSource(ctx context.Context, 
 			glog.Infof("replying to Mist STREAM_SOURCE with stream pull request=%s response=%s", payload.StreamName, pullURL)
 			return pullURL, nil
 		}
-		// another Mist node is currently pulling the stream, sleep and retry
+		glog.Warningf("another node is currently pulling the stream, waiting %v and retrying", streamSourceRetryInterval)
 		time.Sleep(streamSourceRetryInterval)
 	}
 	return "push://", nil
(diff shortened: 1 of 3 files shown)


victorges (approved):
LGTM

handlers/geolocation/geolocation.go
```diff
@@ -200,6 +203,10 @@ func (c *GeolocationHandlersCollection) getStreamPull(playbackID string) (string
 		return "", nil
 	}
 
+	if err := c.Lapi.LockPull(stream.ID, lockPullLeaseTimeout); err != nil {
+		return "", fmt.Errorf("failed to lock pull, err=%v", err)
```
mjh1, line 207: Do you know if this will result in a playback error going back to the client or is there a retry somewhere?

mjh1: Ah no it will end up defaulting to returning `"push://"` so I think we probably need to either return the error in this case to allow a retry to happen (and check the playback behaviour for that)

leszko: Updated with the retries. PTAL.

handlers/geolocation/geolocation.go
```diff
@@ -150,31 +160,43 @@ func (c *GeolocationHandlersCollection) HandleStreamSource(ctx context.Context,
...
+		pullURL, err := c.getStreamPull(playbackID(payload.StreamName))
+		if err != nil && !errors.Is(err, errLockPull) {
 			glog.Errorf("getStreamPull failed for %s: %s", payload.StreamName, err)
+			return "push://", nil
 		} else if pullURL != "" {
 			glog.Infof("replying to Mist STREAM_SOURCE with stream pull request=%s response=%s", payload.StreamName, pullURL)
 			return pullURL, nil
 		}
+		// another Mist node is currently pulling the stream, sleep and retry
+		time.Sleep(streamSourceRetryInterval)
```
mjh1, line 179: Might be useful to have a log line here at least in the beginning for debugging

handlers/geolocation/geolocation.go
```diff
@@ -20,6 +22,14 @@ import (
 	"github.com/livepeer/go-api-client"
 )
 
+const (
+	streamSourceRetries       = 10
+	streamSourceRetryInterval = 15 * time.Second
```
mjh1, line 27: I think we should probably reduce this a bit to try to speed up the response as much as possible. Maybe 2-5s?

handlers/geolocation/geolocation.go
leszko: Good point, updated.

mjh1 (approved):

PR #1192, opened 2024-03-26
ffmpeg with nice/ionice

pwilczynskiclearcode:


video/transmux.go
```diff
@@ -25,6 +25,7 @@ func MuxTStoMP4(tsInputFile, mp4OutputFile string) ([]string, error) {
 	// transmux the .ts file into a standalone MP4 file
 	ffmpegErr := bytes.Buffer{}
 	err := ffmpeg.Input(tsInputFile).
+		SetFfmpegPath("/path/to/ffmpeg-nice.sh").
```
pwilczynskiclearcode, line 28: https://pkg.go.dev/github.com/u2takey/ffmpeg-go#Stream.SetFfmpegPath

video/transmux.go
```diff
@@ -81,7 +82,7 @@ func MuxTStoFMP4(fmp4ManifestOutputFile string, inputs ...string) error {
 
 	timeout, cancel := context.WithTimeout(context.Background(), 10*time.Minute)
 	defer cancel()
-	cmd := exec.CommandContext(timeout, "ffmpeg", args...)
+	cmd := exec.CommandContext(timeout, "/path/to/ffmpeg-nice.sh", args...)
```
pwilczynskiclearcode, line 85: This is pseudo-code. I'm not even sure if these two places are the correct ones for VOD processing

emranemran:
What problem are we trying to solve? nice-ing to such a low priority may make the full transcode cycle much slower. Did this change make some problems go away?

video/transmux.go
emranemran: This is just the MP4 generation step. There's ffmpeg usage in different places for for segmenting, thumbnailing, clipping, etc. Just grep-ing `ffmpeg` or `CommandContext | grep ffmpeg` should get you most of the instances.

pwilczynskiclearcode: So we're not fully sure yet but it looks like some `ffmpeg` for big VODs are congesting both disk writes on datapacket servers and generate big egress spikes resulting in entire server slowdown and k3s restarts. 
[We're testing with Rafał](https://discord.com/channels/423160867534929930/1219450408439648418/1222181474275430451) if moving VOD processing onto just one of the nodes in the cluster results in less or zero restarts of the rest of the nodes.

PR #1197, opened 2024-03-27
Add lock to stream pull to prevent multiple nodes starting pulling for the same stream

leszko:
The previous PR was reverted because it caused prod issues. It was related to the retry loop inside STREAM_SOURCE, so I suggest we don't do any retries, but just fail the playback if it's currently being pulled.

Original PR: https://github.com/livepeer/catalyst-api/pull/1187
Revert PR: https://github.com/livepeer/catalyst-api/pull/1194


mjh1 (commented):
Could you just also add 67281_11879_11879 and 67281_11878_11878 thanks

mjh1 (approved):

mjh1 (approved):

PR #1206, opened 2024-04-02
Record the number of bytes read from storage in a metric

mjh1:
Since it is hard to get access logging from storage providers this will give us an idea of where the egress usage is coming from, making it easier to optimise.

### DIFF
--- a/clients/object_store_client.go
+++ b/clients/object_store_client.go
@@ -26,6 +26,51 @@ func DownloadOSURL(osURL string) (io.ReadCloser, error) {
 	return fileInfoReader.Body, nil
 }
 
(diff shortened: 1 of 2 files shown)


clients/object_store_client.go
```diff
@@ -26,6 +26,51 @@ func DownloadOSURL(osURL string) (io.ReadCloser, error) {
 	return fileInfoReader.Body, nil
 }
 
+func NewTeeReadCloser(readCloser io.ReadCloser, writer io.Writer) io.ReadCloser {
```
leszko, line 29: How about adding a unit test for this?

leszko (approved):
Looks good. Added one comment. Other than that, I think we could keep an eye on the  performance if it was not declined, because of this `TeeReadCloser`.

PR #1212, opened 2024-04-07
transmux: use smaller chunks when concatenating long list of files

emranemran:
When very large files are ingested, the segmented file list can be a very long list. This list of files are passed to ffmpeg commands which will may fail with a 'command: arg list too long' error (linux error and not ffmpeg error). This happens because the arg length exceeds the MAX_ARG limit which varies from kernel to kernel. To workaround this issue, we break down the list into smaller chunks and then concat those chunks to get the final concatenated file.

### DIFF
--- a/video/transmux.go
+++ b/video/transmux.go
@@ -18,6 +18,7 @@ import (
 
 const (
 	Mp4DurationLimit = 21600 //MP4s will be generated only for first 6 hours
+	MaxArgLimit      = 250
 )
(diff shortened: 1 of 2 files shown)


thomshutt (approved):

thomshutt:
I think this looks good, but I'm having trouble envisioning the final ffmpeg comand - could you throw a before / after example in here please?

PR #1215, opened 2024-04-09
Fix node ingest for stream pull

leszko:


handlers/geolocation/geolocation.go
```diff
@@ -242,6 +244,14 @@ func (c *GeolocationHandlersCollection) getStreamPull(playbackID string) (string
 		return "", nil
 	}
 
+	if stream.PullRegion != "" && c.Config.OwnRegion != stream.PullRegion && retryCount < streamSourceMaxWrongRegionRetries {
+		if retryCount == 0 {
+			glog.Infof("Stream pull requested in the incorrect region, sending playback request to the correct region region, playbackID=%s, region=%s", playbackID, stream.PullRegion)
+			c.sendPlaybackRequestAsync(playbackID, stream.PullRegion)
```
emranemran, line 250: Does this need to happen async? If `sendPlaybackRequestAsync` fails for any reason, then the request for playback will just fall-through and never make it to any node right?

leszko: Yes, that's intentional. I think the logic should be that if this request fails or if another node does not pull the stream, the current node should do the stream-pull. 

For example, let's say the `pullRegion` is set to `fra`.
1. We receive this request from a node in `lon`
2. We find that the right region is `fra`, so we send a request from here
3. The node in `lon` receives the requests and pull the stream

All good. But imagine the node in `lon` is down and/or replying slowly. If we make this call sync, then the whole pull stream stops working, because one node in `lon` is down. Keeping it async, we have a fallback scenario, that after 3s, this node in `fra` will start pulling anyway (assuming it was not started by the `lon` node).

emranemran (approved):

handlers/geolocation/geolocation.go
```diff
@@ -242,6 +244,14 @@ func (c *GeolocationHandlersCollection) getStreamPull(playbackID string) (string
 		return "", nil
 	}
 
+	if stream.PullRegion != "" && c.Config.OwnRegion != stream.PullRegion && retryCount < streamSourceMaxWrongRegionRetries {
+		if retryCount == 0 {
+			glog.Infof("Stream pull requested in the incorrect region, sending playback request to the correct region region, playbackID=%s, region=%s", playbackID, stream.PullRegion)
```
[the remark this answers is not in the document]
leszko: Updated

PR #1217, opened 2024-04-10
Update logging to distinguish different cases for STREAM_SOURCE errors

leszko:


handlers/geolocation/geolocation.go
```diff
@@ -160,25 +161,31 @@ func (c *GeolocationHandlersCollection) HandleStreamSource(ctx context.Context,
...
 			} else {
 				// start stream pull
 				glog.Infof("replying to Mist STREAM_SOURCE with stream pull request=%s response=%s", payload.StreamName, pullURL)
 				return pullURL, nil
 			}
 		}
-		if !errors.Is(err, errLockPull) {
+		if errors.Is(err, errNoStreamSourceForActiveStream) {
+			// stream is active, but STREAM_SOURCE cannot be found
+			glog.Errorf("error querying mist for STREAM_SOURCE: %s", errMist)
```
thomshutt, line 186: Could we throw the work "active stream" in here please? Just so I can remember which message is which in the future without referring to the code

i.e `error querying mist for active stream STREAM_SOURCE: %s`

thomshutt (approved):

handlers/geolocation/geolocation.go
leszko: Ok

PR #1229, opened 2024-04-17
Use DNS redirect instead of MistUtilLoad

leszko:
When the given Catalyst node is not overloaded, then use it instead of using the node returned from MistUtilLoad


leszko:
After discussing offline, closing this, we'll try to use tagging in MistUtilLoad.

PR #1247, opened 2024-05-02
Check write permission for output locations

mjh1:
Annoyingly this relies on having delete permission (to delete the temp file once created) but I couldn't find any other way to achieve this check. Pretty sure there are cases where delete is already needed anyway when we overwrite certain output files.

### DIFF
--- a/clients/object_store_client.go
+++ b/clients/object_store_client.go
@@ -93,32 +93,6 @@ func UploadToOSURLFields(osURL, filename string, data io.Reader, timeout time.Du
 	return nil
 }
 
-func CheckWritePermission(url *url.URL) error {
-	if url.Scheme == "memory" {
-		// delete is not supported for memory and it's only used for tests so just return here
-		return nil
-	}
-
-	tmpFileName := "tmp"
-	urlString := url.String()
-
-	// check write permission by uploading a file
-	err := UploadToOSURL(urlString, tmpFileName, strings.NewReader(""), time.Second)
-	if err != nil {
-		return fmt.Errorf("failed to upload tmp file for %s: %w", log.RedactURL(urlString), err)
-	}
-
-	storageDriver, err := drivers.ParseOSURL(urlString, true)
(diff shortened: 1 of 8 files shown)


thomshutt:
Why don't we just call it `metadata.json`, put some metadata about the video in it and then leave it there

clients/object_store_client.go
```diff
@@ -93,6 +93,32 @@ func UploadToOSURLFields(osURL, filename string, data io.Reader, timeout time.Du
 	return nil
 }
 
+func CheckWritePermission(url *url.URL) error {
+	if url.Scheme == "memory" {
+		// delete is not supported for memory and it's only used for tests so just return here
+		return nil
+	}
+
+	tmpFileName := "tmp"
```
victorges, line 102: Maybe add a random UUID or timestamp here to avoid concurrency issues with multiple nodes running this?

victorges (approved):
LGTM

clients/object_store_client.go
mjh1: I think we should be ok from that point of view since we're already writing files statically named e.g. 'index.m3u8' etc? For normal vod jobs this will be a location unique to the job and for transcode api the users have defined their output location.

mjh1:
@thomshutt could you take another look? thanks

test/features/vod.feature
```diff
@@ -25,12 +25,17 @@ Feature: VOD Streaming
...
+    And I submit to the internal "/api/vod" endpoint with "<payload>"
     And receive a response within "3" seconds
     Then I get an HTTP response with code "400"
     And my "failed" vod request metrics get recorded
 
+    Examples:
+      | payload                                             |
+      | an invalid upload vod request                       |
+      | a valid upload vod request with no write permission |
+
```
thomshutt, line 38: Love to see it

thomshutt (approved):

PR #1261, opened 2024-05-16
transcode,video: Add support for copying HLS input segments

victorges:
- [x] Unit  tests
- [x] Test on the box
- [x] Test on staging

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ build:
 .PHONY: build-linux
 build-linux:
 	# Useful for cross-compiling from Mac for testing on an environment
-	env GOOS=linux GOARCH=amd64 go build -o "$(GO_BUILD_DIR)catalyst-api-linux" main.go
+	env GOOS=linux GOARCH=amd64 go build -ldflags="$(ldflags)" -o "$(GO_BUILD_DIR)catalyst-api-linux" main.go
 
 .PHONY: fmt
 fmt:
--- a/clients/broadcaster_remote.go
+++ b/clients/broadcaster_remote.go
@@ -63,7 +63,7 @@ func (c *RemoteBroadcasterClient) TranscodeSegmentWithRemoteBroadcaster(segment 
 		return TranscodeResult{}, fmt.Errorf("pickRandomBroadcaster failed %v", err)
 	}
 
-	return transcodeSegment(segment, sequenceNumber, durationMillis, broadcasterURL, manifestId, profiles, "")
+	return transcodeSegment(segment, sequenceNumber, durationMillis, broadcasterURL, manifestId, "")
 }
 
 // findBroadcaster contacts Livepeer API for a broadcaster to use if localBroadcaster is not defined
@@ -109,9 +109,13 @@ func CreateStream(c Credentials, streamName string, profiles []video.EncodedProf
 	if err != nil {
 		return "", fmt.Errorf("appending stream to api url %s: %v", c.CustomAPIURL, err)
 	}
-	// prepare payload
-	payload := createStreamPayload{Name: streamName}
-	payload.Profiles = append(payload.Profiles, profiles...)
+	if err := validateProfiles(profiles); err != nil {
+		return "", err
+	}
+	payload := createStreamPayload{
+		Name:     streamName,
+		Profiles: append([]video.EncodedProfile(nil), profiles...),
+	}
 	payloadBytes, err := json.Marshal(&payload)
 	if err != nil {
 		return "", fmt.Errorf("POST url=%s json encode error %v struct=%v", requestURL, err, payload)
--- a/transcode/transcode.go
+++ b/transcode/transcode.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"io"
 	"math"
 	"net/url"
 	"os"
@@ -74,8 +75,10 @@ func RunTranscodeProcess(transcodeRequest TranscodeSegmentRequest, streamName st
 
 	// transcodeProfiles are desired constraints for transcoding process
 	transcodeProfiles, err := video.SetTranscodeProfiles(inputInfo, transcodeRequest.Profiles)
-	if err != nil || len(transcodeProfiles) == 0 {
+	if err != nil {
 		return outputs, segmentsCount, fmt.Errorf("failed to set playback profiles: %w", err)
+	} else if len(transcodeProfiles) == 0 {
+		return outputs, segmentsCount, fmt.Errorf("no transcode profiles could be resolved")
 	}
 
 	// Download the "source" manifest that contains all the segments we'll be transcoding
@@ -499,7 +502,7 @@ func getHlsTargetURL(tsr TranscodeSegmentRequest) (*url.URL, error) {
 func transcodeSegment(
 	segment segmentInfo, streamName, manifestID string,
 	transcodeRequest TranscodeSegmentRequest,
-	transcodeProfiles []video.EncodedProfile,
+	encodedProfiles []video.EncodedProfile,
 	targetOSURL *url.URL,
 	transcodedStats []*video.RenditionStats,
 	renditionList *video.TRenditionList,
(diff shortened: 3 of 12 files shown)


transcode/transcode.go
```diff
@@ -507,37 +511,53 @@ func transcodeSegment(
...
 	err := backoff.Retry(func() error {
 		ctx, cancel := context.WithTimeout(context.Background(), clients.MaxCopyFileDuration)
 		defer cancel()
 		rc, err := clients.GetFile(ctx, transcodeRequest.RequestID, segment.Input.URL.String(), nil)
 		if err != nil {
 			return fmt.Errorf("failed to download source segment %q: %s", segment.Input, err)
 		}
+		defer rc.Close()
+
+		var r io.Reader = rc
```
emranemran, line 545: Maybe better to stick this block into a separate function so we can unit test it?

transcode/transcode.go
```diff
@@ -568,47 +605,57 @@ func transcodeSegment(
 	duration := time.Since(start)
 	metrics.Metrics.TranscodeSegmentDurationSec.Observe(duration.Seconds())
 
-	for _, transcodedSegment := range tr.Renditions {
-		renditionIndex := getProfileIndex(transcodeProfiles, transcodedSegment.Name)
-		if renditionIndex == -1 {
-			return fmt.Errorf("failed to find profile with name %q while parsing rendition segment", transcodedSegment.Name)
+	for renditionIndex, profile := range encodedProfiles {
```
emranemran, line 608: same here, pretty simple function but maybe better to break it up into a separate fcn to unit test?

emranemran:
Changes LGTM. One thing that wasn't clear to me was how the manifest for the copied source is handled. The output.m3u8 of a recording has files listed as `source/1.ts`etc so that will need to be modified in the output dir. I didn't see that being handled in the code.

video/profiles.go
```diff
@@ -147,7 +151,18 @@ func GenerateSingleProfileWithTargetParams(videoTrack InputTrack, videoProfile E
 	return profiles
 }
 
-func GetDefaultPlaybackProfiles(video InputTrack) ([]EncodedProfile, error) {
+func GetSourceCopyProfile(video InputTrack) EncodedProfile {
+	return EncodedProfile{
+		Copy:    true,
+		Name:    "source",
```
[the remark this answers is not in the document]
victorges: Sure

video/profiles.go
```diff
@@ -217,6 +236,10 @@ type EncodedProfile struct {
 	ColorDepth   int64  `json:"colorDepth,omitempty"`
 	ChromaFormat int64  `json:"chromaFormat,omitempty"`
 	Quality      uint   `json:"quality,omitempty"`
+	// Copy is a flag to indicate that the profile should be a copy of the input video, no transcoding required. Copying
+	// cannot be specified externally, but is automatically set when the input is in HLS format. This field is not
+	// supported on broadcasters trancode request, so should be used only for internal logic.
+	Copy bool `json:"-"`
```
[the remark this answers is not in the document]
victorges: If we leave it blank it will still be serialized when calling the broadcaster and it would likely explode cause the field doesn't exist (unless it doesn't validate that, but it's still wrong to send a non-existent field IMO).

We could use an `omitempty` instead, given we never send any `Copy: true` values to broadcasters (given that validation I added), but it feels unclear IMO. So `-` makes it clear that this field is never to be serialized (it's go's `json` way to ignore a field).

clients/broadcaster_remote.go
```diff
@@ -107,9 +107,13 @@ func CreateStream(c Credentials, streamName string, profiles []video.EncodedProf
 	if err != nil {
 		return "", fmt.Errorf("appending stream to api url %s: %v", c.CustomAPIURL, err)
 	}
-	// prepare payload
-	payload := createStreamPayload{Name: streamName}
-	payload.Profiles = append(payload.Profiles, profiles...)
+	if err := validateProfiles(profiles); err != nil {
```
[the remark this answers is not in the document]
victorges: Yeah, we will not send any profile. This is to catch coding errors, as copy profiles should never be sent to the broadcaster. If they are sent, this is a mistake like a "null reference exception". I think it would be misleading to make the filter here, as it could lead to behavioral bugs instead which might be harder to catch than if it consistently failed, for example. 

I could make this even more explicit by turning into a `panic` instead, but I think that would be too much. WDYT?

transcode/transcode.go
victorges: I think it's a little overkill since this is only 2 ifs, the complex/testable logic is inside `io.TeeReader` and `io.Copy` IMO. I definitely have a bias for too few tests tho, so I'll follow your suggestion anyway!

video/profiles.go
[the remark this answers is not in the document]
victorges: Let's continue this discussion on https://github.com/livepeer/catalyst-api/pull/1261#discussion_r1607518382

clients/broadcaster_remote.go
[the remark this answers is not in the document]
victorges: Will add unit test

Makefile
```diff
@@ -13,7 +13,7 @@ build:
 .PHONY: build-linux
 build-linux:
 	# Useful for cross-compiling from Mac for testing on an environment
-	env GOOS=linux GOARCH=amd64 go build -o "$(GO_BUILD_DIR)catalyst-api-linux" main.go
+	env GOOS=linux GOARCH=amd64 go build -ldflags="$(ldflags)" -o "$(GO_BUILD_DIR)catalyst-api-linux" main.go
```
emranemran, line 16: why was this required?

transcode/transcode.go
```diff
@@ -570,47 +597,99 @@ func transcodeSegment(
...
+		// This is a copy-only job, so skip transcoding with the broadcaster
+		_, err := io.Copy(source, in)
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to copy source segment: %s", err)
+		}
+		return nil, source, nil
+	}
+
+	// Otherwise if there are profiles to transcode, then tee the source segment to the buffer
+	return io.TeeReader(in, source), source, nil
```
emranemran, line 628: is the second return param (`source`) filled with the io.TeeReader call first when it returns? i.e is source just returned as empty buffer?

emranemran (approved):

emranemran:
Changes LGTM - left a few questions. Also tagging @mjh1 to get more 👀 

* Was testing on staging done with varying files? like 4k vs 1080p? 
* Was clipping testing done? This can be done on staging after merging.

victorges:
Confirmed we don't need this for clips!

transcode/transcode.go
mjh1: What do you think of only allowing copy profiles on their own or would that end up making the API too weird and confusing? Fishtank will just use copy on its own right?

transcode/transcode.go
```diff
@@ -570,47 +597,99 @@ func transcodeSegment(
...
+			return fmt.Errorf("failed to find rendition with name %q while parsing transcode result", profile.Name)
 		}
 
-		targetRenditionURL, err := url.JoinPath(targetOSURL.String(), transcodedSegment.Name)
+		targetRenditionURL, err := url.JoinPath(targetOSURL.String(), profile.Name)
 		if err != nil {
 			return fmt.Errorf("error building rendition segment URL %q: %s", targetRenditionURL, err)
 		}
 
 		if transcodeRequest.GenerateMP4 {
```
mjh1, line 663: So we definitely want to be generating MP4s for fishtanks long recordings?

victorges: Yeah the logic should stay the same here! Generate the mp4s if they are shorter than our threshold (I think 6h? or 12h)

transcode/transcode.go
victorges: It is actually part of the requirements that we _always_ want copy-only for the source profile of recordings. Not only when no transcoding is requested. So it also behaves similar to the profiles in livestreams themselves, where you don't need to specify that the source is made available, you only handle what transcoding is gonna be made additionally from that.

mjh1: oh i see so for the regular recording scenario that we have now that will change to a source copy-only plus transcoded renditions?

PR #1262, opened 2024-05-17
We always want thumbs index to start from zero as users are expecting this

mjh1:
Clip manifests generally have their segment index starting at a non-zero value since they're taken from the middle of a live stream. This was causing the thumb filenames to start with this same index rather than zero. This wasn't great from a user perspective because often they want to just use the filename `keyframe_0.jpg` to get the first frame.

E.g. clip manifest:
```
#EXTM3U
#EXT-X-VERSION:3
#EXT-X-MEDIA-SEQUENCE:0
#EXT-X-TARGETDURATION:8
#EXTINF:2.875,
clip_16.ts
#EXT-X-DISCONTINUITY
#EXTINF:7.125,
clip_17.ts
#EXT-X-ENDLIST
```

Current thumbs output:
```
keyframes_16.jpg
keyframes_17.jpg
```

Fixed thumbs output:
```
keyframes_0.jpg
keyframes_1.jpg
```

### DIFF
--- a/thumbnails/thumbnails.go
+++ b/thumbnails/thumbnails.go
@@ -16,7 +16,6 @@ import (
 	"github.com/cenkalti/backoff/v4"
 	"github.com/grafov/m3u8"
 	"github.com/livepeer/catalyst-api/clients"
-	"github.com/livepeer/catalyst-api/log"
 	"github.com/livepeer/go-tools/drivers"
 	ffmpeg "github.com/u2takey/ffmpeg-go"
 	"golang.org/x/sync/errgroup"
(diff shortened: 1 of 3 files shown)


thumbnails/thumbnails.go
```diff
@@ -218,15 +235,23 @@ func processSegment(input string, thumbOut string) error {
...
+}
+
+func thumbFilename(segmentURI string, segmentOffset int64) (string, error) {
+	i, err := segmentIndex(segmentURI)
 	if err != nil {
-		return "", fmt.Errorf("thumbFilename failed for %s: %w", segmentURI, err)
+		return "", err
 	}
-	return fmt.Sprintf("keyframes_%d.jpg", i), nil
+	return fmt.Sprintf("keyframes_%d.jpg", i-segmentOffset), nil
```
leszko, line 256: Doesn't it complicate things that now the segments will have different numbering than thumbnails? 🤔

leszko (approved):

thumbnails/thumbnails.go
mjh1: That's what I thought initially but then I realised that the output manifests do start from zero, it's only the input manifest which doesn't

PR #1265, opened 2024-05-21
Add Analytics Logs and Kafka Metrics

leszko:


mapic/metrics/census.go
```diff
@@ -225,6 +237,41 @@ func InitCensus(nodeID, version, namespace string) {
...
+			Name:        "log_processor_write_errors_count",
+			Measure:     Census.mLogProcessorWriteErrors,
+			Description: "Number of log processors errors while writing to Kafka",
+			TagKeys:     baseTags,
+			Aggregation: view.Count(),
+		},
+		{
+			Name:        "analytics_log_errors",
+			Measure:     Census.mAnalyticsLogErrors,
+			Description: "Number of errors while posting analytics log",
```
pwilczynskiclearcode, line 250: Is it posting or fetching issue?
https://github.com/livepeer/catalyst-api/pull/1265/files#diff-9d26d9175b31b67c2fcc702fa795bafd990d74b3fcac9f55df02ca461b25f41fR119

leszko: This is mostly the error with the request itself. Like the user posts the wrong `playback_id` or some other data is wrong. Not sure what will be the better name... any thoughts?

handlers/analytics/log_processor.go
```diff
@@ -166,6 +169,14 @@ func (p *LogProcessor) sendEvents() {
 
 	err := p.writer.WriteMessages(context.Background(), msgs...)
```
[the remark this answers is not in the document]
leszko: Added in 0a9fee84c7ef7adb07740618754203059cc7ae5d

FYI: I also switched from census lib to promhttp, because I realize we used census only for mapic.

pwilczynskiclearcode (approved):

leszko:
Good questions @pwilczynskiclearcode 
1. We **don't push any billing data** and all this data is kind-of "opt out" from the user perspective because it's based on a player sending the data. A user can block sending log data and still watch the stream. Or may use a player that just don't send this data. So, not having this data **is not critical in the service** delivery sense.
2. Saying that, obviously we also treat this data seriously, because it's useful data. And we plan to charge users for analytics, so I think the right assumption is that "losing data from one player is ok, but losing data from a significant number of customers is not ok". 
3. Let's think about the unhappy scenario:
  - Kafka is down (or full): I think the risk is low, because Confluent is more reliable than our infa; in that case:
     - Switching to another node does not help
     - We lose the data (the other option would be to switch [this line](https://github.com/livepeer/catalyst-api/blob/ac54bfda03992c3843775b5b4aad1793850d0751/handlers/analytics/log_processor.go#L165) and first send and later remove from the internal memory, but I'm afraid of piling up the mem data in case of some misconfiguration, so for now, let's just have a metric and alert on that)
   - One of our catalyst node has connectivity issues: the risk is higher, in that case:
     - Switching to another node would help, but...
     - I think that in that case most probably the given node will also not serve the traffic correctly, so we should just get that node out of the rotation

thomshutt (approved):

pwilczynskiclearcode:
@leszko One another comment.

https://github.com/segmentio/kafka-go/blob/main/writer.go#L551-L580 we should call sth like `defer writer.Close()` when exiting (somewhere [in here](https://github.com/livepeer/catalyst-api/blob/main/api/http.go#L47-L48))?

PR #1280, opened 2024-06-06
Add rate limiting to stream pull

leszko:


handlers/geolocation/geolocation.go
```diff
@@ -220,7 +257,7 @@ func (c *GeolocationHandlersCollection) HandleStreamSource(ctx context.Context,
 			// stream is active, but STREAM_SOURCE cannot be found
 			glog.Errorf("error querying mist for active stream STREAM_SOURCE: %s", errMist)
```
thomshutt, line 258: While we're in here are there any logging changes that would help us debug the root cause here? e.g adding a stream ID to this log line

thomshutt (approved):

handlers/geolocation/geolocation.go
leszko: I improved logging in: 92f4a20df2fc4cd0514ffbb55e23462bab6781b8

handlers/geolocation/geolocation.go
```diff
@@ -257,6 +297,11 @@ func (c *GeolocationHandlersCollection) getStreamPull(playbackID string, retryCo
 		return "", nil
 	}
 
+	// To prevent overloading the LAPI, we limit the rate at which we query for stream pulls
+	if c.streamPullRateLimit.shouldLimit(playbackID) {
```
mjh1, line 301: Why did you need to change it from just a single `acquire` call? Doesn't this mean that there will be a window of time between the call to `shouldLimit` and `acquire` where multiple routines could call `acquire`?

leszko: Yes, but it's ok for multiple goroutines to `acquire()` in parallel. But your comment made me think I should rename this function to `mark()`. Otherwise, it sounds like a semaphore.

And to your question, I changed it because I realized that this function is also executed for streams that are not pull, but just ingest. Then, we don't want to make any retries or rate limit for those. In particular, if there are multiple playback requests at the same to the same non-active stream, we don't want to retry it.

gioelecerati (approved):

mjh1 (approved):

PR #1281, opened 2024-06-06
mapic: multistream: question mark support

gioelecerati:


mapic/mistapiconnector_app.go
```diff
@@ -708,7 +708,12 @@ func (mc *mac) reconcileMultistream(mistState clients.MistState) {
 	for k, v := range cachedMap {
 		if v.enabled && !mistMap[toKey(k.stream, k.target)] {
 			glog.Infof("adding AUTO_PUSH for stream=%s target=%s", k.stream, k.target)
-			if err := mc.mist.PushAutoAdd(k.stream, k.target); err != nil {
+			// Support for streamKeys with a question mark
```
victorges, line 711: I think it'd be useful to explain why this has any effect of supporting `?` on the streamkey. i.e. be explicit about the mist behavior and why we need to do this
```suggestion
			// Mist strips anything after the last `?` of the URL. So add an extra one in the end so it doesn't mess with the desired target URL which may contain a `?` itself.
```

victorges (approved):

leszko (approved):
LGTM

@gioelecerati Did you check in the box if it actually works (both cases: with and without `?`)?

gioelecerati:
> LGTM
> 
> @gioelecerati Did you check in the box if it actually works (both cases: with and without `?`)?

Yeah it's working in both cases in the box

PR #1293, opened 2024-06-14
Lock for the whole of GetState() to avoid parallel mist API calls

mjh1:
I was still observing hundreds of mist API calls being made on startup, much better than the thousands we had before but it still has the potential to blow up. This should fix that issue.

The cache uses a RWMutex which means multiple routines can end up with a cache miss at the same time and therefore calling through to the mist API. We want to avoid this and have the other routines wait so we lock for the whole of the GetState function.

### DIFF
--- a/clients/mist_client.go
+++ b/clients/mist_client.go
@@ -48,7 +48,7 @@ type MistClient struct {
 	TriggerCallback string
 	configMu        sync.Mutex
 	cache           *cache.Cache
-	// The cache uses a RWMutex which means multiple routines can end up with a cache miss at the same time and end up
+	// The go-cache library uses a RWMutex which means multiple routines can end up with a cache miss at the same time and end up
 	// calling through to the mist API. We want to avoid this and have the other routines wait so we lock for the
 	// whole of the GetState function.
 	stateCacheMu sync.Mutex
--- a/clients/mist_client.go
+++ b/clients/mist_client.go
@@ -48,6 +48,10 @@ type MistClient struct {
 	TriggerCallback string
 	configMu        sync.Mutex
 	cache           *cache.Cache
+	// The cache uses a RWMutex which means multiple routines can end up with a cache miss at the same time and end up
+	// calling through to the mist API. We want to avoid this and have the other routines wait so we lock for the
+	// whole of the GetState function.
+	stateCacheMu sync.Mutex
 }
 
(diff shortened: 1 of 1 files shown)


clients/mist_client.go
```diff
@@ -48,6 +48,10 @@ type MistClient struct {
 	TriggerCallback string
 	configMu        sync.Mutex
 	cache           *cache.Cache
+	// The cache uses a RWMutex which means multiple routines can end up with a cache miss at the same time and end up
```
leszko, line 51: You mention `RWMutex`, but you actually use `sync.Mutex` 🙃

clients/mist_client.go
```diff
@@ -427,6 +431,9 @@ func (mc *MistClient) GetStreamInfo(streamName string) (MistStreamInfo, error) {
 }
 
 func (mc *MistClient) GetState() (MistState, error) {
+	mc.stateCacheMu.Lock()
```
leszko, line 434: Shouldn't we just lock inserting to the cache? Then you can actually use this `RWMutex`. Otherwise it needs to wait for the whole Mist communication.

clients/mist_client.go
mjh1: I meant the cache library https://github.com/patrickmn/go-cache/blob/master/cache.go#L43

clients/mist_client.go
mjh1: That's the whole issue this PR is concerning, because the library uses an `RWMutex` we still see bursts of parallel requests to mist. If you think it safer I can wait for [my metric addition](https://github.com/livepeer/catalyst-api/commit/e84a3c84a6864c626949c94dbc05e9e740355ddd) to go to prod and definitively check the number of requests we're seeing before proceeding with this PR.

clients/mist_client.go
leszko: I see. Thanks for upating the comment!

clients/mist_client.go
leszko: Yeah, if there is no rush, then I think it's safer. 

TBH I think this _should_ be ok, but locking for longer may have some impact.

leszko (approved):
LGTM and as discussed, if possible, we can wait for the metrics first.

mjh1:
Didn't end up needing this. It can be revived if necessary at a later date

PR #1296, opened 2024-06-19
Add viewer limit

leszko:
Related to: https://www.notion.so/livepeer/Cap-50k-Fishtank-viewers-a828ef8d037a4f27b9a6769d91c50b23

**Related PRs:**
- https://github.com/livepeer/livepeer-data/pull/194
- https://github.com/livepeer/studio/pull/2223

**Comments:**
- The solution is not 100% bullet-proof (we may encounter viewers exceeding the limit); that's done on purpose, because we prefer to allow a few viewers above the limit rather than increasing the TTFF. The reasons why viewers may be allowed above the limit:
  - data about the concurrent view count is note refresh (livepeer data takes data from Victoria Metrics nad it has some delay)
  - we cache the concurrent view count for 30s
  - we refresh the concurrent view count **after** allowing the user (because we don't want to make a sync call)
  - viewer limit may not be refreshed (10 min cache)
- We should do some refactoring and changes (but I prefer to do it in a separate PRs), in particular:
  - rate limit can be removed and replaced with the viewer limit mechanism
  - parsing Gate API response should use the standard json umarshall mechanism

### DIFF
--- a/handlers/accesscontrol/access-control.go
+++ b/handlers/accesscontrol/access-control.go
@@ -94,6 +94,7 @@ type ConcurrentViewersCache struct {
 type ConcurrentViewersCacheEntry struct {
 	ViewCount   int32
 	LastRefresh time.Time
+	mux         sync.Mutex
 }
 
 type RefreshIntervalRecord struct {
@@ -367,20 +368,27 @@ func (ac *AccessControlHandlersCollection) refreshConcurrentViewerCacheAsync(pla
 		viewerLimitCache.mux.RUnlock()
 
 		concurrentViewersCache.mux.Lock()
-		defer concurrentViewersCache.mux.Unlock()
-
-		if _, ok := concurrentViewersCache.data[playbackID]; !ok {
-			concurrentViewersCache.data[playbackID] = &ConcurrentViewersCacheEntry{}
+		vc, ok := concurrentViewersCache.data[playbackID]
+		if !ok {
+			vc = &ConcurrentViewersCacheEntry{}
+			concurrentViewersCache.data[playbackID] = vc
 		}
-		if time.Since(concurrentViewersCache.data[playbackID].LastRefresh) > 30*time.Second {
+		concurrentViewersCache.mux.Unlock()
+
+		vc.mux.Lock()
+		if time.Since(vc.LastRefresh) > 30*time.Second {
 			viewCount, err := ac.dataClient.QueryServerViewCount(viewerLimit.UserID)
 			if err != nil {
 				glog.Errorf("Error querying server view count: %v", err)
 				return
 			}
+
+			concurrentViewersCache.mux.Lock()
 			concurrentViewersCache.data[playbackID].ViewCount = viewCount
 			concurrentViewersCache.data[playbackID].LastRefresh = time.Now()
+			concurrentViewersCache.mux.Unlock()
 		}
+		vc.mux.Unlock()
 	}()
 }
 
--- a/handlers/accesscontrol/access-control.go
+++ b/handlers/accesscontrol/access-control.go
@@ -340,10 +340,6 @@ func checkRateLimit(playbackID string) bool {
 
 // checkViewerLimit is used to limit viewers per user globally (as configured with Gate API)
 func (ac *AccessControlHandlersCollection) checkViewerLimit(playbackID string) bool {
-	// We don't want to make any blocking calls, so refreshing the cache async
-	// The worse that can happen is that we allow a few users above the limit for a few seconds
-	defer func() { ac.refreshConcurrentViewerCache(playbackID) }()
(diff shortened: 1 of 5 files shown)


handlers/accesscontrol/access-control.go
```diff
@@ -73,21 +78,48 @@ type HitRecords struct {
...
+// ViewerLimitCache comes from Gate API
+type ViewerLimitCache struct {
+	data map[string]*ViewerLimitCacheEntry
+	mux  sync.RWMutex
+}
+
+type ViewerLimitCacheEntry struct {
+	ViewerLimitPerUser int32
+	UserID             string
+}
```
victorges, line 90: - Do we need to cache these separately from the gate API response?
- Shouldn't this have some `LastRefresh` or similar field?

handlers/accesscontrol/access-control.go
```diff
@@ -452,6 +584,20 @@ func (g *GateClient) QueryGate(body []byte) (bool, GateConfig, error) {
 			}
 			refreshInterval = int32(refreshIntervalFloat64)
 		}
+		if ri, ok := result["viewer_limit_per_user"]; ok {
```
victorges, line 587: Not a regression, but any idea why we parse the JSON manually here instead of using `json:"field_name"` tags?

handlers/accesscontrol/access-control.go
```diff
@@ -379,6 +498,19 @@ func (ac *AccessControlHandlersCollection) cachePlaybackAccessControlInfo(playba
...
+	if gateConfig.ViewerLimitPerUser != 0 && gateConfig.UserID != "" {
+		if _, ok := viewerLimitCache.data[playbackID]; !ok {
+			viewerLimitCache.data[playbackID] = &ViewerLimitCacheEntry{}
+		}
+		viewerLimitCache.data[playbackID].ViewerLimitPerUser = gateConfig.ViewerLimitPerUser
+		viewerLimitCache.data[playbackID].UserID = gateConfig.UserID
+	} else {
+		delete(viewerLimitCache.data, playbackID)
+	}
+	viewerLimitCache.mux.Unlock()
```
victorges, line 512: Hmm, WDYT of changing `PlaybackAccessControlEntry` to store the `GateConfig` object instead and avoiding this separate cache? Then to get the viewer limit for a given playback ID (e.g. [L378](https://github.com/livepeer/catalyst-api/pull/1296/files#diff-e4212fa5c0109280d6faf7edd53a45935b2460fca63acb0e9fadf092e1fc225bR378)), we would instead do `GetPlaybackAccessControlInfo` (which can change to return to be the full `PlaybackAccessControlEntry`, not just the boolean `Allow`). 

This would mean we change the viewer limit check to after [L328](https://github.com/livepeer/catalyst-api/pull/1296/files#diff-e4212fa5c0109280d6faf7edd53a45935b2460fca63acb0e9fadf092e1fc225bR328), which IMO makes more sense: we only check the viewer limit after fetching it (rn from what I could tell, the very first playback won't be checked against the limit right? the separate cache only gets initialized after it is used)

handlers/accesscontrol/data_client.go
```diff
@@ -0,0 +1,62 @@
...
+	if userID == "" {
+		return 0, fmt.Errorf("userID is empty")
+	}
+
+	url := fmt.Sprintf("%s/views/internal/server/now?userId=%s", d.Endpoint, userID)
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return 0, fmt.Errorf("failed to create request, err=%v", err)
+	}
+	req.Header.Set("Content-Type", "application/json")
```
victorges, line 39: Not needed as there's no body
```suggestion
```

victorges (approved):
LGTM with one suggestion on the access control flow

handlers/accesscontrol/access-control.go
leszko: No, that was one of my comments in the PR description. I'll address if while removing the rate limit logic. In this Linear ticket: https://linear.app/livepeer/issue/ENG-2120/change-rate-limite-to-viewer-limit

handlers/accesscontrol/access-control.go
leszko: Yeah, that could work this way. Though, I'm not sure I want to touch that flow right now.

This code is quite messy and very sensitive, because: 1) It makes a blocking call 2) It affects all viewers for all customers. Also, the rate limit is implemented with its own cache, so it would still be inconsistent. I'll reconsider it while removing `rate limit` code.

victorges: Cool. Yeah, I think this only makes sense if the `rate_limit` gets refactored too. Let's keep it separate.

PR #1297, opened 2024-06-19
Remove request count client metric

mjh1:
We already have a count metric automatically created from the duration metric.


leszko (approved):

thomshutt:
@mjh1 could you add a description about why we're doing this please?

PR #1300, opened 2024-06-20
mapic: Fix multistream target URL `?` handling

victorges:
It turns out that Mist has a special parsing of these push URLs which is not a regular URL parsing. It removes anything after the last `?` in the URL and uses that as its config. If there are multiple `?` in the URL (e.g. query-string already existed in target URL), it will strip only (from) the last one and keep the previous.

So instead of merging Mist's qs with the existing ones, we actually need to create a separate `?` section for Mist. This guarantees both that Mist gets its config properly, as well as that it preserves the URL to call the target service.


thomshutt:
Could you add some unit tests to make this easier to reason about please?

gioelecerati (approved):

victorges (commented):
LGTM

PR #1302, opened 2024-06-20
clients: Handle not found errors gracefully

victorges:
Currently, 404 errors are not treated any differently on clients code. This means that if we get a 
404 error from a storage we keep retrying until the end of the retry policy. This was a problem for
the storage fallback code so I'm starting with this refact, which improves behavior anyway.

After this:
- They stop backoff retry loops (as long as it gets returned or wrapped)
- If they're included in the final error of the job, they will make the `unretriable` flag be `true` so studio doesn't retry
- The returned errors can be detected as 404s easily (necessary for storage fallback logic)

### Changes
- Implement a "not exist" error in the lib https://github.com/livepeer/go-tools/pull/121, which is consumed here.
- Create a `ObjectNotFoundError` type in catalyst. This will be especially useful for the storage fallback code to identify this error.
- Add the 404 handling on each client and make sure errs are wrapped properly

### Context
This was a problem for storage fallback logic for the manifest file, because we need to:
- Always download from the 2 storages to be able to compare the most recent one
- Not wait for the entire retry loop to finish in case the resource simply doesn't exist in one of the storages

### DIFF
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"strings"
 
+	"github.com/cenkalti/backoff/v4"
 	"github.com/livepeer/catalyst-api/log"
 	"github.com/xeipuuv/gojsonschema"
 )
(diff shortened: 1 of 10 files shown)


errors/errors.go
```diff
@@ -63,21 +65,55 @@ func WriteHTTPBadBodySchema(where string, w http.ResponseWriter, errors []gojson
...
+}
+
+func NewObjectNotFoundError(msg string, cause error) error {
+	if cause != nil {
+		msg = fmt.Sprintf("ObjectNotFoundError: %s: %s", msg, cause)
+	} else {
+		msg = fmt.Sprintf("ObjectNotFoundError: %s", msg)
+	}
+	// every not found is unretriable
+	return Unretriable(ObjectNotFoundError{msg: msg, cause: cause})
```
mjh1, line 108: @victorges I wonder if we could simplify this code a bit by using golang's [multiple wrapped errors](https://tip.golang.org/doc/go1.20#errors) feature. Probably not worth it but curious to see what you think.

mjh1 (approved):

PR #1306, opened 2024-06-24
backoff: Fix usage of backoff library

victorges:
This fixes 2 misuses of the backoff lib found in `catalyst-api` code:
- Re-using singleton instances of a backoff. They are not meant to be shared so we should create one on every call
- Not calling `Reset` on the exponential backoff after configuring it. This (confusing) requirement is necessary if we change the `IniitalDelay`. By not calling it we were basically ignoring any changes and going with the default 500ms.

### DIFF
--- a/clients/object_store_client.go
+++ b/clients/object_store_client.go
@@ -184,8 +184,8 @@ func newExponentialBackOffExecutor() *backoff.ExponentialBackOff {
 	backOff := backoff.NewExponentialBackOff()
 	backOff.InitialInterval = 200 * time.Millisecond
 	backOff.MaxInterval = maxRetryInterval
-	backOff.MaxElapsedTime = 0 // don't impose a timeout as part of the retries
-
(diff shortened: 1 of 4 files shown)


clients/object_store_client.go
```diff
@@ -184,8 +184,8 @@ func newExponentialBackOffExecutor() *backoff.ExponentialBackOff {
 	backOff := backoff.NewExponentialBackOff()
 	backOff.InitialInterval = 200 * time.Millisecond
 	backOff.MaxInterval = maxRetryInterval
-	backOff.MaxElapsedTime = 0 // don't impose a timeout as part of the retries
-
+	backOff.MaxElapsedTime = 0 // don't impose a timeout as part of the retries=
```
mjh1, line 187: Trailing `=`? :)

mjh1 (approved):

clients/object_store_client.go
victorges: oh yeah i think i was coding with the box running and my mac was toasting lol

PR #1319, opened 2024-07-15
Use constant port for MistUtilLoad and reconcile periodically to make sure it's running

leszko:
The context for this change is that we sometimes see multiple MistUtilLoad instances running, sometimes zero instances running. Therefore we should always make sure that only 1 instance is running all the time.

This change will also make simpler to implement Zero Impact Catalyst API Deployments, because we'll statically configure MistUtilPort. Additionally, this change includes periodic members refresh (in case the serf member event is missed).

Design Doc: https://www.notion.so/livepeer/Zero-Impact-Catalyst-API-Deployments-c2b15232f3a2450ba3e4803130ecd2be?pvs=4#5ce68dfca1c341bfb6b21361d36d19df

Tagging @iameli , because you may remember the context of why we're running MistUtilLoad using a random port.

### DIFF
--- a/balancer/mist/mist_balancer.go
+++ b/balancer/mist/mist_balancer.go
@@ -19,7 +19,6 @@ import (
 	"github.com/golang/glog"
 	"github.com/livepeer/catalyst-api/balancer"
 	"github.com/livepeer/catalyst-api/cluster"
-	"golang.org/x/sync/errgroup"
 )
 
 var mistUtilLoadSingleRequestTimeout = 15 * time.Second
@@ -50,14 +49,12 @@ func NewBalancer(config *balancer.Config) balancer.Balancer {
 
 // start this load balancer instance, execing MistUtilLoad if necessary
 func (b *MistBalancer) Start(ctx context.Context) error {
-	group, ctx := errgroup.WithContext(ctx)
-	group.Go(func() error {
-		return b.execBalancer(ctx, b.config.Args)
-	})
-	group.Go(func() error {
-		return b.waitForStartup(ctx)
-	})
-	return group.Wait()
+	b.killPreviousBalancer(ctx)
+
+	go func() {
+		b.reconcileBalancerLoop(ctx, b.config.Args)
+	}()
+	return b.waitForStartup(ctx)
 }
 
 // wait for the mist LB to be available. can be called multiple times.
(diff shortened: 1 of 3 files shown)


hjpotter92 (approved):

balancer/mist/mist_balancer.go
```diff
@@ -245,6 +242,45 @@ func (b *MistBalancer) formatNodeAddress(server string) string {
...
+		err := b.execBalancer(ctx, balancerArgs)
+		if err != nil {
+			glog.Warningf("Error starting MistUtilLoad: %v", err)
+		}
+	}
+}
+
+// isBalancerRunning checks if MistUtilLoad is running.
+func (b *MistBalancer) isBalancerRunning(ctx context.Context) bool {
+	cmd := exec.CommandContext(ctx, "pgrep", "-f", "MistUtilLoad")
```
mjh1, line 279: Maybe we could emit a warning log or something if we find more than one instance running? Should be impossible but just in case..

leszko: I think that to do it, we'd need to parse the command output. TBH I think it's not worth doing it, because we shouldn't have more than 1 MistUtilLoad running now. If each binds to the same port, they'll just fail to start anyway.

balancer/mist/mist_balancer.go
```diff
@@ -245,6 +242,45 @@ func (b *MistBalancer) formatNodeAddress(server string) string {
...
+			return
+		case <-ticker.C:
+		}
+	}
+}
+
+// reconcileBalancer makes sure that MistUtilLoad is up and running.
+func (b *MistBalancer) reconcileBalancer(ctx context.Context, balancerArgs []string) {
+	if !b.isBalancerRunning(ctx) {
+		err := b.execBalancer(ctx, balancerArgs)
```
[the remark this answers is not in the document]
leszko: Good idea. Added in 1c514a541b2283d59868f997d92519c52823245b

thomshutt:
Do we have any way of reproducing / testing this problem in Staging?

My general concerns would be along the lines of something being wrong with the existing load balancer processes when it gets into this state and now we're also not spinning up a new one

balancer/mist/mist_balancer.go
```diff
@@ -245,6 +242,46 @@ func (b *MistBalancer) formatNodeAddress(server string) string {
...
+// killPreviousBalancer cleans up the previous MistUtilLoad process if it exists.
+// It uses pkill to kill the process.
+func (b *MistBalancer) killPreviousBalancer(ctx context.Context) {
+	cmd := exec.CommandContext(ctx, "pkill", "-f", "MistUtilLoad")
+	cmd.Run()
+}
+
+// reconcileBalancerLoop makes sure that MistUtilLoad is up and running all the time.
+func (b *MistBalancer) reconcileBalancerLoop(ctx context.Context, balancerArgs []string) {
+	ticker := time.NewTicker(1 * time.Minute)
```
thomshutt, line 254: Does this mean that we potentially have a minute without a functioning load balancer in the worst case?

leszko: Yes

leszko:
> Do we have any way of reproducing / testing this problem in Staging?

No, we don't know what happens in prod/staging. But we've seen 2 scenarios in prod:
- 2 instances of MistUtilLoad running (my guess is that what happened is that when restarting catalyst-api, the old MistUtilLoad process changed the parent process id to PID 1 and stayed alive, then it was hanging unused - still causing some 5xx troubles for our monitoring)
- 0 instances of MistUtilLoad running (my guess is that it just seg faulted or something and no one re-created it)

> My general concern would be along the lines of something being wrong with the existing load balancer processes when it gets into this state and now we're also not spinning up a new one
This is the opposite of what happens in this PR.
- Before this PR: MistUtilLoad crashes => we don't recreate it until catalyst-api is restarted
- After this PR: MistUtilLoad crashes => we check and recreate it every 1 min

thomshutt (approved):

thomshutt:
LGTM then - I think you just need to fix whatever the linting error is

PR #1321, opened 2024-07-15
clip: handle end segment offsets that are miniscule

emranemran:
The end time specified for a clip may end up in a segment where it's only several nano/micro seconds in length from the beginning of that segment. For example, it might be that the start/end times specified by the user causes the logic to clip the end segment from t=0s to t=0.000001s. In this case, passing these params to ffmpeg would cause it to select 0.0.0.000 for both the -ss (trim from) and -to (trim to) arguments which would cause the command to error out and eventually the clip as well.

To resolve this case, we round down to the nearest 3 decimal places. Any deltas < 0.0001s would get rounded down to 0s in which case the existing logic would correctly pick the previous segment.


video/clip_test.go
```diff
@@ -153,12 +153,3180 @@ func TestClippingSucceedsWhenValidManifestIsUsed(t *testing.T) {
...
+
+	segs, _, err = ClipManifest("1234", plE, 9271.241, 10401.36)
+	length = len(segs)
+	require.NoError(t, err)
+	require.Equal(t, 113, length)
+	require.Equal(t, uint64(1039), uint64(segs[112].SeqId))
+
 }
+
+const manifestE = `#EXTM3U
```
mjh1, line 179: Can you just save this to a fixture file?

mjh1 (approved):

thomshutt:
Change looks good, but is it worth setting a minimum clip duration at the request level? I don't think there's any real use case for a < 1s clip. We could even do 0.5s if we want to be conservative

PR #1323, opened 2024-07-18
Add Docker build

leszko:
We plan to enable running catalyst-api separately to catalyst, therefore it needs its own Docker image. 

Note that it I didn't include `w3` tool, because Web3 Storage has shut down, so we should remove the support for Web3 Storage.

Design Doc: https://www.notion.so/livepeer/Zero-Impact-Catalyst-API-Deployments-c2b15232f3a2450ba3e4803130ecd2be?pvs=4#5ce68dfca1c341bfb6b21361d36d19df

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -1,0 +1,76 @@
+name: Build and publish catalyst-api docker images
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+      - dev
+    tags:
+      - "v*"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  docker:
+    name: Build and publish docker image
+    runs-on: ubuntu-latest
+    permissions:
+      packages: write
+      contents: read
+
+    steps:
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      - name: Login to DockerHub
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.CI_DOCKERHUB_USERNAME }}
+          password: ${{ secrets.CI_DOCKERHUB_TOKEN }}
+
+      - name: Log in to the Container registry
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ github.token }}
+
+      - name: Extract metadata (tags, labels) for Docker
+        id: meta
+        uses: docker/metadata-action@v4
+        with:
+          images: |
+            livepeer/${{ github.event.repository.name }}
+            ghcr.io/${{ github.repository }}
+          tags: |
+            type=sha
+            type=ref,event=pr
+            type=ref,event=tag
+            type=sha,format=long
+            type=ref,event=branch
+            type=semver,pattern={{version}},prefix=v
+            type=semver,pattern={{major}}.{{minor}},prefix=v
+            type=raw,value=latest,enable={{is_default_branch}}
+            type=raw,value=${{ github.event.pull_request.head.ref }}
+            type=raw,value=stable,enable=${{ startsWith(github.event.ref, 'refs/tags/v') }}
+
+      - name: Build and push
+        uses: docker/build-push-action@v4
+        with:
+          platforms: linux/amd64, linux/arm64
+          push: true
+          build-args: |
+            version=${{ github.ref_type == 'tag' && github.ref_name || github.sha }}
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+ADD	go.mod go.sum	./
+RUN	go mod download
+
+ADD	.	.
+RUN make build
+
+ARG	GIT_VERSION
+ENV	GIT_VERSION="${GIT_VERSION}"
+
+FROM	ubuntu:22.04	AS	catalyst
```
leszko, line 28: I copied it the base image from `livpeeer/catalyst`, @hjpotter92 @pwilczynskiclearcode do you think we should try a smaller image?

Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+ARG	BUILD_TARGET
+
+RUN	apt update && apt install -yqq wget software-properties-common \
+	&& add-apt-repository -y ppa:ubuntuhandbook1/ffmpeg6
+
+RUN	apt update && apt install -yqq \
+	curl \
+	ca-certificates \
+	musl \
+	python3 \
```
leszko, line 43: Not sure we need all these dependencies. I guess it's for VOD. @thomshutt @mjh1 could you confirm we need those for vod?

thomshutt: I'm not sure - other than ca-certificates we might be able to get rid of some of those others. It'll probably need a bit of trial and error though unfortunately

Dockerfile
leszko: Do we want to try some liter (apline?) images? Do you know if it could work? @hjpotter92 Do you maybe know?

leszko:
@hjpotter92 also I see the build failed because the Docker Hub registry does not exist. How to create it?

```
buildx failed with: ERROR: failed to solve: failed to push livepeer/catalyst-api:pr-1323: push access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed
```

Dockerfile
```diff
@@ -0,0 +1,53 @@
+ARG	GIT_VERSION=unknown
+ARG	BUILD_TARGET
+ARG	FROM_LOCAL_PARENT
```
hjpotter92, line 3: you can remove unused args

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,76 @@
...
+            type=raw,value=${{ github.event.pull_request.head.ref }}
+            type=raw,value=stable,enable=${{ startsWith(github.event.ref, 'refs/tags/v') }}
+
+      - name: Build and push
+        uses: docker/build-push-action@v4
+        with:
+          platforms: linux/amd64, linux/arm64
+          push: true
+          build-args: |
+            version=${{ github.ref_type == 'tag' && github.ref_name || github.sha }}
```
hjpotter92, line 69: since the arg in dockerfile is called `GIT_VERSION`, it should be renamed accordingly here.

Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+FROM	ubuntu:22.04	AS	catalyst
+
+ENV	DEBIAN_FRONTEND=noninteractive
+
+LABEL	maintainer="Amritanshu Varshney <amritanshu+github@livepeer.org>"
+
+ARG	BUILD_TARGET
+
+RUN	apt update && apt install -yqq wget software-properties-common \
+	&& add-apt-repository -y ppa:ubuntuhandbook1/ffmpeg6
```
hjpotter92, line 37: didn't we bump to ffmpeg7?

Dockerfile
hjpotter92: we stopped using alpine because it made debugging on the pods difficult. devs. needed to install base packages before they could even do top/kill etc.

hjpotter92 (approved):

Dockerfile
leszko: Yes, updated.

.github/workflows/docker.yaml
leszko: Updated

Dockerfile
leszko: Ok, I removed most of them (including ffmpeg), because I think we'll run VOD nodes still as the whole Catalyst bundle. And for the livestream API, we don't need it.

thomshutt (approved):

Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+
+RUN	apt update && apt install -yqq \
+	curl \
+	ca-certificates \
+	musl \
+	python3 \
+	procps \
+	vnstat \
+	ffmpeg \
+	gstreamer1.0-tools gstreamer1.0-plugins-good gstreamer1.0-plugins-base gstreamer1.0-plugins-bad \
```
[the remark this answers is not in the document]
leszko: Ok, I found the related [Linear](https://linear.app/livepeer/issue/ENG-1443/add-gstreamer-to-catalyst-container-add-opus-aac-audio-transcoding) and [Discord](https://discord.com/channels/423160867534929930/1108406937919430716/1108709559498969099). We don't need it in catalyst-api.

PR #1326, opened 2024-07-18
Change HTTP Error code and error message when we don't have write permissions to the S3 storage

leszko:
It's related to the credentials leak to the user's error.


handlers/upload.go
```diff
@@ -335,7 +335,7 @@ func checkWritePermission(reqID, externalID string, urls ...*url.URL) error {
 		err := clients.UploadToOSURL(u.String(), "metadata.json", strings.NewReader(fmt.Sprintf(`{"external_id": "%s"}`, externalID)), time.Second)
 		if err != nil {
 			log.LogError(reqID, "failed write permission check", err, "url", log.RedactURL(urlString))
-			return fmt.Errorf("failed write permission check for %s", urlString)
+			return fmt.Errorf("failed write permission check")
```
thomshutt, line 338: I think we have an obfuscation method somewhere, could we use that here? It is quite useful to know the URL that failed

leszko: Ok, updated. 

But stil, it gets returned to users which we probably don't want. Is there any way that we return it to task-runner, but not to the end users?

thomshutt: Yeah, we don't want users seeing that - I thought your change to make it an Internal Server Error would stop that though

leszko: I think you're right. We don't pass Internal Server Errors to the users. Then, I guess this PR is good now.

thomshutt (approved):

PR #1330, opened 2024-07-18
Change Serf communication from local to http calls

leszko:
We plan to run Serf member on a different host than catalyst-api, therefore it needs to communicate remotely rather than directly accessing Serf.

Design Doc: https://www.notion.so/livepeer/Zero-Impact-Catalyst-API-Deployments-c2b15232f3a2450ba3e4803130ecd2be

### Rationale behind this approach

Stateless catalyst-api needs to receive Serf user events to perform a few actions: multistream update, nuke stream, stop sessions, etc.

Technically, we have 2 options here:
- Option 1: Catalyst receives a user event and triggers Stateless Catalyst API
- Option 2: Stateless Catalyst API joins the serf cluster and receives the messages on its own

I chose Option 1, mainly because:
1. Joining serf cluster by Catalyst API is not *so* straightforward, it requires
   - exposing serf advertise port to the public (currently we have a static configuration using `hostPort`, but for Stateless Catalyst API it would need to be somehow dynamic)
   - tagging different types of nodes and testing scenarios
2. We plan to anyway replace serf with Redis or something similar


thomshutt (approved):
This looks good, it hopefully also opens the door nicely to having something other than Serf manage the cluster state in the future

PR #1335, opened 2024-07-22
Add logging about reason for not generating an MP4

thomshutt:
Currently there's no way of distinguishing between the different reasons when MP4 generation doesn't happen and so this extra logging means we will now always be able to differentiate.

### DIFF
--- a/pipeline/coordinator.go
+++ b/pipeline/coordinator.go
@@ -323,18 +323,19 @@ func (c *Coordinator) StartUploadJob(p UploadJobPayload) {
 		if p.C2PA {
 			si.C2PA = c.C2PA
 		}
-		si.SourceFile = osTransferURL.String()  // OS URL used by mist
-		si.SignedSourceURL = signedNewSourceURL // http(s) URL used by mediaconvert
-		si.InputFileInfo = inputVideoProbe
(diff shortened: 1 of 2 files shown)


pipeline/coordinator.go
```diff
@@ -323,18 +323,19 @@ func (c *Coordinator) StartUploadJob(p UploadJobPayload) {
 		if p.C2PA {
 			si.C2PA = c.C2PA
 		}
-		si.SourceFile = osTransferURL.String()  // OS URL used by mist
+		si.SourceFile = osTransferURL.String() // OS URL used by mist
+		log.AddContext(p.RequestID, "new_source_url", si.SourceFile)
+
 		si.SignedSourceURL = signedNewSourceURL // http(s) URL used by mediaconvert
+		log.AddContext(p.RequestID, "signed_url", si.SignedSourceURL)
```
thomshutt, line 330: Refactor: Moved these to line up with the places where we get the info

pipeline/coordinator.go
```diff
@@ -377,26 +378,26 @@ func checkClipResolution(p UploadJobPayload, inputVideoProbe *video.InputVideo,
...
 	}
 
-	if fragMp4TargetUrl != nil {
-		return true
+	if fragMp4TargetUrl == nil {
+		return false, "missing MP4 target URL"
 	}
 
-	return false
+	return true, ""
```
thomshutt, line 400: Flipped the logic to make it more of a standard Golang flow of error checking and then success

leszko (approved):
👍

PR #1339, opened 2024-07-23
Add 1s cache for lapi.GetStreamByPlaybackID()

leszko:
It causes the prod failure, because of the cascading failures

Related Discord thread: https://discord.com/channels/423160867534929930/1265176750544130068/1265181494763327562

### DIFF
--- a/mapic/utils.go
+++ b/mapic/utils.go
@@ -1,0 +1,58 @@
+package mistapiconnector
+
+import (
+	"github.com/livepeer/go-api-client"
+	"sync"
+	"time"
+)
+
+const lapiCacheDuration = 1 * time.Second
+
+type ApiClientCached struct {
+	lapi        *api.Client
+	streamCache map[string]entry
+	mu          sync.RWMutex
+	ttl         time.Duration
+}
+
+func NewApiClientCached(lapi *api.Client) *ApiClientCached {
+	return &ApiClientCached{
+		lapi:        lapi,
+		streamCache: make(map[string]entry),
+		ttl:         lapiCacheDuration,
+	}
+}
+
+type entry struct {
+	stream   *api.Stream
+	err      error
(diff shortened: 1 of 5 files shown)


mapic/utils.go
```diff
@@ -0,0 +1,58 @@
+package mistapiconnector
+
+import (
+	"github.com/livepeer/go-api-client"
+	"sync"
+	"time"
+)
+
+const lapiCacheDuration = 1 * time.Second
```
victorges, line 9: Do we need the latest version of a stream for any config that can be changed? Maybe this cache could be even higher

mapic/utils.go
```diff
@@ -0,0 +1,58 @@
...
+	// Check again in case another goroutine has updated the cache in the meantime
+	e, ok = a.streamCache[playbackId]
+	if ok && e.stream != nil && e.updateAt.Add(a.ttl).After(time.Now()) {
+		return e.stream, e.err
+	}
+	// No value in the cache, fetch from Livepeer API and store the result in a cache
+	stream, err := a.lapi.GetStreamByPlaybackID(playbackId)
+	a.streamCache[playbackId] = entry{
+		stream:   stream,
+		err:      err,
```
victorges, line 54: do we need to cache error cases if we only use cache when it's successful? Could be clearer to have an early return for `err != nil` here instead and skip the err in cache

victorges (approved):
LGTM

mapic/utils.go
leszko: Not sure. I'm just afraid to set it higher (and 1s is enough for what we encountered)

mapic/utils.go
leszko: We do want to cache errors as well. That was actually the root cause of the issue that we never cached failed requests so they piled up.

PR #1345, opened 2024-07-25
thumbnails: png output

gioelecerati:
- Use `png` as output for thumbnails
- Remove `q:v` quality flag from thumbnails generation

`q:v` was added previously as an attempt to fix thumbnail quality - however reports of low quality thumbnails were due to some specific HDR videos. Having `png` as output fixes the issue


thomshutt:
Could we reduce the default resolution a little bit too please? I think the one now is excessive and it'll hopefully keep filesizes under control after the move to PNG

thomshutt (approved):

PR #1347, opened 2024-07-26
Improvements for Zero Impact Catalyst API Deployments

leszko:
This PR includes 2 ugly changes that are needed to deploy catalyst-api separately from catalyst:

### Change 1: Proxy `/api/events` requests

The events are used, e.g.,  to notify multistream target change and to nuke a stream. They are received in catalyst, then propagated to all nodes using Serf, and then forwarded to each catalyst-api instance. The change included here is to expose `/api/events` not only in catalyst, but also in catalyst-api (which then proxies the request to catalyst). It's needed because Studio API communicates events with catalyst-api (not directly with catalyst).

Alternatively, we could have two catalyst params in Studio API (but I think it would be "even more" ugly.

### Change 2: Hack to resolve Catalyst API URL if not specified

Catalyst needs to communicate with each corresponding Statelss Catalyst API. E.g. `staging-catalyst-0` needs to communicate with `staging-catalyst-api-0`. At first I thought it will be just an env variable, but then I realized that to introduce a new env variable we need to restart each catalyst node 🤦 . So to avoid restarting, I added a dirty hack to resolve dynamically the corresponding catalyst-api instance in the code (if not specified).

### DIFF
--- a/handlers/events.go
+++ b/handlers/events.go
@@ -1,7 +1,6 @@
 package handlers
 
 import (
-	"bytes"
 	"encoding/json"
 	"fmt"
 	"github.com/golang/glog"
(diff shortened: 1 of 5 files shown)


thomshutt (approved):

handlers/events.go
```diff
@@ -75,6 +80,39 @@ func (d *EventsHandlersCollection) Events() httprouter.Handle {
 	}
 }
 
+// ProxyEvents is a handler of Catalyst API called by Studio API.
+// It proxies the requests to Catalyst.
+func (d *EventsHandlersCollection) ProxyEvents() httprouter.Handle {
+	// Proxy the request to d.eventsEndpoint
+	return func(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+		// Read the request body
+		body, err := io.ReadAll(req.Body)
```
mjh1, line 89: Could we not just forward the body into the new request rather than reading it into memory?

leszko: Right. Updated in f4949ad7bb2ead7d6512b23fd4d4fd644efdafa2

PR #1352, opened 2024-07-31
analytics.go: Fix typo in error message formatting

hjpotter92:
i was getting tired of seeing `%v` appear in logs:
```
W0731 05:15:49.119430 2566432 analytics.go:117] cannot parse geo info from analytics log request header, err=%vmissing geo headers: [X-Region-Name]
```

### DIFF
--- a/handlers/analytics.go
+++ b/handlers/analytics.go
@@ -114,12 +114,12 @@ func (c *AnalyticsHandlersCollection) Log() httprouter.Handle {
 		}
 		geo, err := parseAnalyticsGeo(r)
 		if err != nil {
-			glog.Warning("cannot parse geo info from analytics log request header, err=%v", err)
+			glog.Warningf("cannot parse geo info from analytics log request header, err=%v", err)
 		}
 		extData, err := c.extFetcher.Fetch(log.PlaybackID)
 		if err != nil {
 			metrics.Metrics.AnalyticsMetrics.AnalyticsLogsErrors.Inc()
-			glog.Warning("error enriching analytics log with external data, err=%v", err)
+			glog.Warningf("error enriching analytics log with external data, err=%v", err)
 			cerrors.WriteHTTPBadRequest(w, "Invalid playback_id", nil)
 		}
 


pwilczynskiclearcode:
We're [flooding the logs](https://eu-metrics-monitoring.livepeer.live/grafana/goto/6hI1_49IR?orgId=1) with these warnings/errors. It's mostly `X-Region-Name` missing... maybe we could live without `Region` and stop spamming the logs

hjpotter92:
> We're [flooding the logs](https://eu-metrics-monitoring.livepeer.live/grafana/goto/6hI1_49IR?orgId=1) with these warnings/errors. It's mostly `X-Region-Name` missing... maybe we could live without `Region` and stop spamming the logs

yeah, probably might be better if we only log for cases when multiple headers are missing (for eg.):
```
W0731 09:48:49.223187 2835021 analytics.go:117] cannot parse geo info from analytics log request header, err=%vmissing geo headers: [X-City-Country-Name X-City-Country-Code X-Region-Name]
```

pwilczynskiclearcode:
> > We're [flooding the logs](https://eu-metrics-monitoring.livepeer.live/grafana/goto/6hI1_49IR?orgId=1) with these warnings/errors. It's mostly `X-Region-Name` missing... maybe we could live without `Region` and stop spamming the logs
> 
> yeah, probably might be better if we only log for cases when multiple headers are missing (for eg.):
> 
> ```
> W0731 09:48:49.223187 2835021 analytics.go:117] cannot parse geo info from analytics log request header, err=%vmissing geo headers: [X-City-Country-Name X-City-Country-Code X-Region-Name]
> ```

[There are no cases](https://eu-metrics-monitoring.livepeer.live/grafana/goto/_GBEuV9IR?orgId=1) of missing headers other than `X-Region-Name` so either excluding just this header from the warning or reporting only if more than one is missing like you suggested should be good.

thomshutt:
@ecmulli is this header missing expected or something we need to investigate? If it's expected then agree we should stop spamming the logs about it

ecmulli:
> @ecmulli is this header missing expected or something we need to investigate? If it's expected then agree we should stop spamming the logs about it

It is not expected but I like @pwilczynskiclearcode 's suggestion of only logging when multiple are missing.

leszko:
> > @ecmulli is this header missing expected or something we need to investigate? If it's expected then agree we should stop spamming the logs about it
> 
> It is not expected but I like @pwilczynskiclearcode 's suggestion of only logging when multiple are missing.

I think we should not log when `X-Region-Name` is missing (as suggested by Paweł). We expect this field sometimes not being present. @hjpotter92 I think you can merge your PR and I'll fire a new one with this logic.

leszko (approved):

PR #1354, opened 2024-08-07
fix: added ID field to analytics parsing

0xcadams:
Added an ID field to the analytics logs, which is generated from the Player and passed through to CH.


handlers/analytics.go
```diff
@@ -45,8 +46,9 @@ type AnalyticsLog struct {
 
 type AnalyticsLogEvent struct {
 	// Shared fields by all events
-	Type      string `json:"type"`
-	Timestamp int64  `json:"timestamp"`
+	Id        *string `json:"id"`
```
victorges, line 49: lopri/nit, but by [go standards](https://go.dev/wiki/CodeReviewComments#initialisms) we keep acronyms fully uppercase (or fully lowercase). So these new fields should be `ID`

handlers/analytics_test.go
```diff
@@ -83,6 +85,7 @@ func TestHandleLog(t *testing.T) {
 						"some_field": "some value"
 					},
 					{
+						"id": "abcde12345",
```
victorges, line 88: Perhaps worth adding a test case without an `id` present if you're worried about the backward compatible behavior?

handlers/analytics/log_processor.go
```diff
@@ -28,6 +29,7 @@ type LogProcessor struct {
 
 type LogDataEvent struct {
 	// Heartbeat event
+	Id                  *string `json:"id,omitempty"`
```
victorges, line 32: re what happens when the field is not set, you're right that it will just be `nil`!

victorges (approved):
LGTM

PR #1356, opened 2024-08-12
analytics: Don't log on missing X-Region-Name header

leszko:
Geo header X-Region-Name is not available in all regions, therefore we should treat it as optional.

### DIFF
--- a/handlers/analytics.go
+++ b/handlers/analytics.go
@@ -116,7 +116,7 @@ func (c *AnalyticsHandlersCollection) Log() httprouter.Handle {
 		}
 		geo, err := parseAnalyticsGeo(r)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


pwilczynskiclearcode (approved):

pwilczynskiclearcode:
I would also change [this line](https://github.com/livepeer/catalyst-api/blob/main/handlers/analytics.go#L119)
to produce a warning without `err` as that misleads that's some kind of error...
```
cannot parse geo info from analytics log request header, missing geo headers ...
```
instead of:
```
cannot parse geo info from analytics log request header, err=missing geo headers:
```

thomshutt (approved):

leszko:
> this line

Updated in f84ec5cf509a2d4cae9820d15db24bab3d4f4d33

PR #1357, opened 2024-08-13
Fix version in the Docker CI Build

leszko:
The `version` was not present in the Docker image

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -67,6 +67,7 @@ jobs:
           push: true
           build-args: |
             GIT_VERSION=${{ github.ref_type == 'tag' && github.ref_name || github.event.pull_request.head.sha || github.sha }}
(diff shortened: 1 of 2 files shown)


.github/workflows/docker.yaml
```diff
@@ -67,6 +67,7 @@ jobs:
           push: true
           build-args: |
             GIT_VERSION=${{ github.ref_type == 'tag' && github.ref_name || github.event.pull_request.head.sha || github.sha }}
+            version=${{ github.ref_type == 'tag' && github.ref_name || github.sha }}
```
hjpotter92, line 70: the `Dockerfile` also needs to be updated so that it accepts version.


i'd recommend using `GIT_VERSION` already defined in workflow (and the dockerfile as an ARG)

leszko: Ok, Updated. PTAL

hjpotter92 (approved):

PR #1360, opened 2024-08-15
Implement percentage node replacement for load balancer

mjh1:
This is a hack to allow us to re-route a portion of traffic from an overloaded region to alternative nodes. We have a region where we'd like to balance more traffic away from onto alternative regions instead.

For example to re-route 25% of `regionX` to a selection of other nodes:
```
lb-replace-host-match = regionX
lb-replace-host-percent = 25
lb-replace-host-list = otherNode1,otherNode2
```

### DIFF
--- a/balancer/balancer.go
+++ b/balancer/balancer.go
@@ -126,4 +126,8 @@ type Config struct {
 	MistHost                 string
 	OwnRegion                string
 	OwnRegionTagAdjust       int
+
+	ReplaceHostMatch   string
+	ReplaceHostList    []string
+	ReplaceHostPercent int
 }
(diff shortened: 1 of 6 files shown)


mjh1:
Think I should be able to add some unit tests... hold tight

mjh1:
@thomshutt please could you review? thanks

thomshutt (approved):

leszko (approved):
@mjh1 could you add to the description the info "why are we doing it? 🙃

mjh1:
Sure, I had meant the reason was that we need to `re-route a portion of traffic from an overloaded region to alternative nodes`

PR #1362, opened 2024-08-21
Proxy Mist triggers from catalyst to catalyst-api and remove dynamic Mist Trigger Setup

leszko:
I always thought that catalyst-api dynamically sets Mist triggers. This is true, but this config gets overwritten with the default config present [here](https://github.com/livepeer/livepeer-infra/blob/a82a30fcd185fc6b9fc3fa52412c8608725b47c4/values/catalyst.values.yaml#L191) every time MistController restarts of the config file is changed. The only reason it worked in prod is that our dynamic Mist Trigger configuration is exactly the same as the config map configuration.

I think it's super misleading and it led to the issues when I tried to use the standalone catalyst-api, because my config got overwritten and all stopped working at some point.

-----

First of all, I suggest removing this whole dynamic mist triggers setup (to avoid further confusion). Secondly, we need to think how to solve it if catalyst-api is deployed separately.

1. Option 1: What I implemented in this PR => Proxy each Mist trigger request from catalyst to catalyst-api
2. Option 2: Monitor Mist triggers and dynamically override them each time the config changes
3. Option 3: Investigate it further, maybe if we remove the Mist triggers configuration from [here](https://github.com/livepeer/livepeer-infra/blob/a82a30fcd185fc6b9fc3fa52412c8608725b47c4/values/catalyst.values.yaml#L191) it will not override the dynamically set configuration; the reason I didn't start from this option is that I'm afraid we'll come to some corner cases like Mist crashing or something and then we'll end up with Mist with no triggers set up at all
4. Option 4: Have a separate config map for each catalyst pod (then we can differentiate between triggers in the infra config)

### DIFF
--- a/api/http_internal.go
+++ b/api/http_internal.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"context"
 	"database/sql"
+	"fmt"
 	"net/http"
 	"net/http/httptest"
 	"time"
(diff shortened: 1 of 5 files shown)


mjh1 (approved):

mjh1:
Could you link to the block(s) of code for the dynamic triggers setup which in theory could be remove? Or are you saying you've removed those bits already in this PR?
This idea of proxying doesn't seem to bad since we're already having to do that for some events right? (implemented in https://github.com/livepeer/catalyst-api/pull/1347)

PR #1367, opened 2024-08-26
Add 'catalyst=true' label to Mist-enriched metrics

leszko:
When the catalyst-api was run inside catalyst, we used VM Agent to collect the metrics and then added the `catalyst=true` label in the [VM Agent config](https://github.com/livepeer/livepeer-infra/blob/89b7549035488359fee8761375ffba9e3b3c2856/values/catalyst.values.yaml#L176).

Now, for the standalone Catalyst API we use Kubernetes prometheus scrape which does not add this label. At the same time, we use this label in a really high number of charts, so the simple fix is to just add this label here.

An alternative would be to update all Grafana charts that use `catalyst=true` metric with a condition `app_kubernetes_io_name="catalyst-api"`. However I see there are really a lot of these charts, so I'm not sure there is a simple way to update  them all.

### DIFF
--- a/mapic/mistmetrics.go
+++ b/mapic/mistmetrics.go
@@ -75,7 +75,7 @@ func (mc *mac) parsePlaybackID(line string) (string, bool) {
 
 func (mc *mac) enrichLabels(playbackID string) string {
 	res := mc.streamLabel(playbackID)
-	res += ",catalyst=true"
+	res += `,catalyst="true"`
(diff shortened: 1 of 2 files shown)


hjpotter92 (approved):

pwilczynskiclearcode (approved):

pwilczynskiclearcode:
Good catch

PR #1368, opened 2024-08-26
Add timeout for fetching Mist metrics from catalyst-api

leszko:
During the catalyst deployment, we sometimes lost the catalyst-api metrics, because we had no timeout and the `/metrics` endpoint was waiting forever when the Mist was down. 

This PR adds a simple timeout

### DIFF
--- a/mapic/mistmetrics.go
+++ b/mapic/mistmetrics.go
@@ -1,18 +1,25 @@
 package mistapiconnector
 
 import (
+	"context"
 	"fmt"
 	"github.com/golang/glog"
 	"io"
 	"net/http"
 	"strings"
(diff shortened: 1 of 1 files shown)


pwilczynskiclearcode (approved):

pwilczynskiclearcode:
Shouldn't we have a timeout on all the queries to mist?

leszko:
> Shouldn't we have a timeout on all the queries to mist?

I think we should, though the usual (POST) request to Mist didn't cause any issues. Wrt metrics, for some reason, if Mist is down, then catalyst-api becomes blocked and stops returning metrics.

PR #1369, opened 2024-08-28
Increase analytics buffer from 25k to 200k

leszko:


handlers/analytics.go
```diff
@@ -26,7 +26,7 @@ import (
 
 const (
 	GeoHashPrecision     = 3
-	LogChannelBufferSize = 25000
+	LogChannelBufferSize = 200000
```
pwilczynskiclearcode, line 29: I hope that some player bug won't cause some enormous kafka events overproduction. Just thinking corner case - all viewers sending 200K events every couple of seconds.

pwilczynskiclearcode (approved):

thomshutt (approved):

handlers/analytics.go
leszko: yeah, fair point.

PR #1370, opened 2024-08-28
Limit Serf User Events

leszko:
### Problem

This PR fixes the issue with the catalyst deployment:
1. Catalyst cluster always stores all the user events (since the beginning of the serf cluster formation (currently ~10k user events)
2. After I split catalyst-api to run as a separate app, serf events are propagated with http request from catalyst => catalyst-api
3. When we deploy catalyst, it joins the serf cluster, it reads these ~10k user events, and it sends ~10k http requests to catalyst-api
4. This causes `cannot assign requested error`

### Fix
The main fix is to configure serf cluster [to not process old user events](https://github.com/livepeer/catalyst-api/pull/1370/files#diff-86cbfc045d9b3441a4f4c2235e276915275698f1e70a42e38e74a846cb9bdcc4R160). This should be enough. But to be on the safe side, we introduced a few additional changes:
- [Reconcile cluster members on startup too](https://github.com/livepeer/catalyst-api/pull/1371) - in case the (old) cluster events are not propagated (thanks @mjh1 )
- Limit the numbers of user events processed at the same time (in the case of more than `20` concurrent user events, next events will be dropped)
- Make this `50` configurable with catalyst-api param and make the serf cluster internal event buffer const to `100000` (the same as before, but not it's not configurable with a param, we don't need to configure it)
- Process events sequentially (thanks @pwilczynskiclearcode )


pwilczynskiclearcode:
alternatively we could remove `go` coroutine from [processClusterEvent()](https://github.com/livepeer/catalyst-api/blob/main/main.go#L432-L446) to make sure that events are processes sequentially and not in parallel

pwilczynskiclearcode (approved):

mjh1:
Does limiting this buffer mean the excess events just get dropped?

pwilczynskiclearcode:
> Does limiting this buffer mean the excess events just get dropped?

It would block [on the producer side](https://go.dev/doc/effective_go#channels):
> If the channel has a buffer, the sender blocks only until the value has been copied to the buffer; if the buffer is full, this means waiting until some receiver has retrieved a value.

But @leszko might need to confirm what would be the consequence.

pwilczynskiclearcode (approved):

leszko:
> alternatively we could remove `go` coroutine from [processClusterEvent()](https://github.com/livepeer/catalyst-api/blob/main/main.go#L432-L446) to make sure that events are processes sequentially and not in parallel.
> 
> @leszko @mjh1

It's a good idea. Actually we need it. Thanks @pwilczynskiclearcode Updated in 17d9f548a11d38bca56e0188d75539af066b1d5a

leszko:
> > Does limiting this buffer mean the excess events just get dropped?
> 
> It would block [on the producer side](https://go.dev/doc/effective_go#channels):
> 
> > If the channel has a buffer, the sender blocks only until the value has been copied to the buffer; if the buffer is full, this means waiting until some receiver has retrieved a value.
> 
> But @leszko might need to confirm what would be the consequence.

After this change to process events sequentially, it will effectively mean that if we have more than `<limit>` events to process, events will get dropped. I think it's not a problem, but to be on the safe side, I've changed to `50`

mjh1:
@leszko just realised my change doesn't quite work because `c.serf` is nil to begin with on startup which is why the cuke tests are failing. I can push another commit to fix it, bit messy though annoyingly

mjh1 (approved):

PR #1389, opened 2025-01-07
Write server stats to postgres for catabalancer

mjh1:
Switch to postgres to store node stats for catabalancer. Previously when we used serf to propagate this data throughout the cluster we ran into bottlenecks with messages queueing up, it didn't seem it was designed for that sort of use case.

- The catalyst nodes (the ones running mist) will periodically update the db with their stats.
- The catalyst-api pods will read the data from the db for all nodes and make the load balancing decisions.

Future PRs:
- strip out the unused serf message handling code
- cache the db results from `RefreshNodes()` at the moment it can be called multiple times concurrently, possibly just change it to a periodic refresh.

### DIFF
--- a/balancer/catabalancer/catalyst_balancer.go
+++ b/balancer/catabalancer/catalyst_balancer.go
@@ -2,6 +2,8 @@ package catabalancer
 
 import (
 	"context"
+	"database/sql"
+	"encoding/json"
 	"fmt"
 	"math/rand"
 	"sort"
@@ -23,6 +25,7 @@ type CataBalancer struct {
 	NodeMetrics         map[string]NodeMetrics // Node name -> NodeMetrics
 	metricTimeout       time.Duration
 	ingestStreamTimeout time.Duration
+	MetricsDB           *sql.DB
 }
 
 type Streams map[string]Stream // Stream ID -> Stream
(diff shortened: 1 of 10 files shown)


balancer/catabalancer/catalyst_balancer.go
```diff
@@ -79,8 +83,36 @@ func (s ScoredNode) String() string {
 	)
 }
 
+// JSON representation is deliberately truncated to keep the message size small
+type NodeUpdateEvent struct {
```
mjh1, line 87: just moved this code from the events package to make the dependencies easier

balancer/catabalancer/catalyst_balancer.go
```diff
@@ -88,6 +119,7 @@ func NewBalancer(nodeName string, metricTimeout time.Duration, ingestStreamTimeo
 		NodeMetrics:         make(map[string]NodeMetrics),
 		metricTimeout:       metricTimeout,
 		ingestStreamTimeout: ingestStreamTimeout,
+		MetricsDB:           nodeStatsDB,
```
leszko, line 122: IIUC This will be a different database than our metrics DB. Then, could you rename it to something like `NodeStatsDB`?

balancer/catabalancer/catalyst_balancer.go
```diff
@@ -291,7 +327,58 @@ func truncateReturned(scoredNodes []ScoredNode, numNodes int) []ScoredNode {
...
+		if err != nil {
+			return fmt.Errorf("failed to unmarshal node update event: %w", err)
+		}
+
+		if isStale(event.NodeMetrics.Timestamp, c.metricTimeout) {
+			log.LogNoRequestID("catabalancer skipping stale data while refreshing", "nodeID", event.NodeID, "timestamp", event.NodeMetrics.Timestamp)
+			continue
+		}
+
+		c.UpdateNodes(event.NodeID, event.NodeMetrics)
```
leszko, line 361: Do we actually need to keep any internal state for nodes? If we make a query to DB with every playback redirect request, then couldn't we reason everything from there?

leszko (approved):
Other 2 comments. Other than that LGTM

thomshutt (approved):

balancer/catabalancer/catalyst_balancer.go
mjh1: Yep sorry I totally forgot to mention refactoring that as a plan for another PR, in two minds about whether I should do that in this PR but I think since it's been approved I think I'll just do it now but in a fresh PR.

PR #1390, opened 2025-01-13
Refactor to remove local catabalancer state

mjh1:
Now that we're using a DB for the node stats we can remove the internal state for this data. The only state we still maintain in catabalancer is the list of nodes updated by serf.

### DIFF
--- a/balancer/balancer.go
+++ b/balancer/balancer.go
@@ -5,7 +5,6 @@ package balancer
 import (
 	"context"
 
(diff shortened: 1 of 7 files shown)


leszko (approved):

mjh1:
Had to make a small change not to use the serf member updates, catalyst-api doesn't receive those, we can rely on the DB data instead and exclude nodes where the data is stale.

leszko (approved):

PR #1395, opened 2025-01-16
Cache node stats db results

mjh1:
Background load balancer calls were seen taking multiple hours in prod (only on the regions with higher request rate) due to the relatively small number of connections available: https://github.com/livepeer/catalyst-api/blob/f0d8c83827abe68af240afa1a8dcfc894ec7c259/main.go#L241 so cache the results for a short time (500ms currently)


mjh1:
I think I'll probably make the cache expiry time configurable

leszko (approved):

PR #1399, opened 2025-01-27
Add timeout for node metric sending

mjh1:
In prod I saw stale metrics being ignored so we need to check the durations of the send operation and add an appropriate timeout.

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -34,19 +34,21 @@ type AnalyticsMetrics struct {
 }
 
 type CatalystAPIMetrics struct {
-	Version                         *prometheus.CounterVec
-	UploadVODRequestCount           prometheus.Counter
-	UploadVODRequestDurationSec     *prometheus.SummaryVec
-	TranscodeSegmentDurationSec     prometheus.Histogram
-	PlaybackRequestDurationSec      *prometheus.SummaryVec
-	CDNRedirectCount                *prometheus.CounterVec
-	CDNRedirectWebRTC406            *prometheus.CounterVec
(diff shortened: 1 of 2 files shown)


metrics/metrics.go
```diff
@@ -132,6 +134,16 @@ func NewMetrics() *CatalystAPIMetrics {
...
 			Buckets: []float64{.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10},
 		}, []string{"success", "request_type", "mist_match", "background"}),
+		CatabalancerSendMetricDurationSec: promauto.NewHistogram(prometheus.HistogramOpts{
+			Name:    "catabalancer_send_metric_duration",
+			Help:    "Time taken to send catabalancer node metric updates",
+			Buckets: []float64{.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10},
+		}),
+		CatabalancerSendDBDurationSec: promauto.NewHistogramVec(prometheus.HistogramOpts{
+			Name:    "catabalancer_send_db_duration",
+			Help:    "Time taken to send catabalancer node metric updates to DB",
```
pwilczynskiclearcode, line 144: This one is for sending metric update to DB.
What is `catabalancer_send_metric_duration` about? It includes the time for pulling the metrics from `mist`, right?
I'm not sure if I understand the difference reading the `Help:` message

pwilczynskiclearcode (approved):

metrics/metrics.go
mjh1: That's right so we have a better idea of where any slowness lies, I'll update the help text

PR #1400, opened 2025-02-05
Implement timeout for whole metrics send function

mjh1:
I saw occasional slowness from the mist GetState function. The mist client has a 1 minute timeout and we need a much shorter timeout here as the loop runs every 5 seconds

### DIFF
--- a/balancer/catabalancer/catalyst_balancer.go
+++ b/balancer/catabalancer/catalyst_balancer.go
@@ -412,6 +412,7 @@ func (c *CataBalancer) MistUtilLoadSource(ctx context.Context, streamID, lat, lo
 }
 
 var StatsUpdateInterval = 5 * time.Second
+var StatsUpdateTimeout = StatsUpdateInterval - 500*time.Millisecond // have the timeout sit within the update interval so we don't miss sending updates
 
 func isStale(timestamp time.Time, stale time.Duration) bool {
 	return time.Since(timestamp) >= stale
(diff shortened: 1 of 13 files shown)


mjh1:
The diff is best viewed with whitespace hidden

balancer/catabalancer/catalyst_balancer.go
```diff
@@ -421,58 +421,76 @@ func StartMetricSending(nodeName string, latitude float64, longitude float64, mi
 	ticker := time.NewTicker(updateNodeStatsEvery)
 	go func() {
 		for range ticker.C {
-			start := time.Now()
-			sysusage, err := GetSystemUsage()
-			if err != nil {
-				log.LogNoRequestID("catabalancer failed to get sys usage", "err", err)
+			done := make(chan bool)
```
leszko, line 424: I think the more "golang way" of doing the same would be passing context with timeout into `sendMetrics()`. So you could have something like this:

```golang
ctx, cancel := context.WithTimeout(context.Background(), timeout)
sendMetrics(ctx, nodeName, latitude, longitude, mist, nodeStatsDB)
```

And then inside `sendMetrics()` to have

```golang
GetSystemUsage(ctx)
```

And so on. The good thing about this approach would be that in case of the timeout you cancel all system and network calls.

leszko (approved):
Added a suggestion. Either way, LGTM

balancer/catabalancer/catalyst_balancer.go
mjh1: @leszko I've pretty much re-done this PR, please could you take a look? thanks

main.go
```diff
@@ -267,6 +267,7 @@ func main() {
 
 		if catabalancerEnabled && nodeStatsDB != nil {
 			if cli.Tags["node"] == "media" { // don't announce load balancing availability for testing nodes
+				mist := clients.NewMistAPIClient(cli.MistUser, cli.MistPassword, cli.MistHost, cli.MistPort, catabalancer.StatsUpdateInterval-time.Second)
```
leszko, line 270: Could you add a comment with the reasoning why it's `catabalancer.StatsUpdateInterval-time.Second`? It's not clear to me.

leszko (approved):
Added one inline comment. Other than that, LGTM

